PR #12, opened 2021-06-29
fix: remove print statements from folding

akinsho:
Whilst cycling through folds using `<S-tab>` the cycle level is printed to the message area. This PR removes this (assuming it's unintended)


kristijanhusak:
Actually it's intended. Emacs prints these when toggling the folding, and I wanted to mimic that.

akinsho:
Ah, alright, was debating an issue but thought it was accidental, and I'd just remove them. Will close this out then 👍🏾

PR #15, opened 2021-06-30
Add file line number hyperlink

patgro1:
In addition to file link, emacs org mode uses the line number in the template. I decided to add it to this org-mode and add the functionality needed to jump to the line number when editing the file.


kristijanhusak:
Nice! Thanks!

PR #21, opened 2021-07-02
Add capture expansions for journaling

jamesob:
Adds various capture expansions necessary to generate journal headings, e.g.
```
*** 2021-07-02 Friday
**** [2021-07-02 Fri 08:58]

wowowow it's a journal entry
```
Also documents the `target` option of captures.

### DIFF
--- a/doc/orgmode.txt
+++ b/doc/orgmode.txt
@@ -185,17 +185,21 @@ ORG_AGENDA_TEMPLATES                                *orgmode-org_agenda_template
 
 type: `table<string, table>`
 default value: `{ t = { description: 'Task', template: '* TODO %?\n  %u' } }`
-Templates for capture/refile prompt.
+Templates for capture/refile prompt. Use `target =` to specify which file to append to.
 Variables:
   * `%t`: Prints current date (Example: `<2021-06-10 Thu>`)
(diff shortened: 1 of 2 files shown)


kristijanhusak:
From the [Orgmode's documentation](https://orgmode.org/manual/Template-expansion.html#Template-expansion), I see that `%x` and `%A` has some other functionality.
I don't want to override any of these.

We could introduce `%(EXP)` or `%<FORMAT>` which would allow to dynamically insert whatever you want, or figure out some neovim specific specification for custom expansions that are available only here (maybe `$A` or something like that).

jamesob:
Yeah, good call. Figured those might be stepping on some toes. Might be nice, as you suggest, to provide an escape hatch that just gets forwraded to `os.date(...)`.

kristijanhusak:
@jamesob I just merged a `%<FORMAT>` option that you could use to insert any date format in template. See [DOCS](https://github.com/kristijanhusak/orgmode.nvim/blob/master/DOCS.md#org_agenda_templates). Evaluating any expression with `%(EXP)` is tricky because I cannot use simple regex parser to parse everything between brackets, since code can easily contain more brackets (example: `%(os.date('%Y-%m-%d'))`).
This should do the trick for journaling.

Closing this in favor of that.

PR #23, opened 2021-07-04
Allow Agenda.tags to list TODOs only

chuwy:
Hello!

Just a small disclaimer:

* This is my first Lua code ever
* This is my first Vim code in 6 years
* I have very small experience with org-mode
* I generally have no idea what I'm doing

This PR allows users to filter only TODO items with specific keys by:

```vim
:lua require('orgmode').action("agenda.tags", { clear_sarch = true, tags = "personal", todo_only = true })
```

Above, the second argument becomes a table allowing `Agenda.tags` function to be non-interactive and filter only TODOs. I'm not sure if this is how it supposed to be - I see that you tend to be very conservative in representing Emacs behavior, but I needed that functionality and decided why not.

I couldn't find if `Agenda.tags` used anywhere with passing boolean explicitly.

P.S. That's a great plugin, thanks!


kristijanhusak:
Hi!

Thanks for the contribution. Code looks ok, no much complains there.

I investigated a bit how orgmode does it, and noticed there is an Agenda option `M` (we have only `m` here), which does exactly this. It searches tags only on unfinished tasks.

If you want, I can guide you through code review how to add the new option and provide this functionality, or I can introduce it myself and close this PR in favor of that.

chuwy:
Hey @kristijanhusak!

> we have only `m` here

That was actually the place where we were passign a boolean explicitly! Fixed it.

> If you want, I can guide you through code review how to add the new option and provide this functionality, or I can introduce it myself and close this PR in favor of that.

That's obviously completely up to you - I'm happy to get my hands dirty with NeoVim. There's one thing though that I think is really worth to preserve - a non-interactive way to call the function. I personally really dislike org-mode menu and keen to use [which-key](https://github.com/folke/which-key.nvim/) for these purposes.

lua/orgmode/agenda/init.lua
```diff
@@ -306,17 +306,30 @@ function Agenda:search(clear_search)
 end
 
 -- TODO: Add PROP/TODO Query
-function Agenda:tags(clear_search)
-  if clear_search then
+function Agenda:tags(opts)
+  local tags
```
kristijanhusak, line 310: Move opts default up and define tags as default from opt.tags:

```
opts = opts or {}
local tags = opts.tags

if not tags then
 tags = vim.fn.input('Match: ', self.last_search, 'customlist,v:lua.orgmode.autocomplete_agenda_filter_tags')
endif
```

kristijanhusak (changes requested):

lua/orgmode/agenda/init.lua
chuwy: Done!

kristijanhusak (approved):

kristijanhusak:
@chuwy I'll merge this, since it implements the missing feature. Note that I'll update it a bit to not access todo only tags via `agenda.tags`, but I'll instead expose a different public method called `agenda.tags_todo`. I'll leave tags property on both so it can be overriden on demand.

Thanks for the PR.

chuwy:
Thanks, @kristijanhusak! I'm planning to start working on `org-journal` functionality somewhere later this week. Please don't hesitate to tell me if I need to change something in future PRs or that the functionality isn't what it supposed to be.

kristijanhusak:
I pushed the change. You can use it like this for regular tag view:

```vim
:lua require('orgmode').action("agenda.tags", { tags = "personal" })
```

and for todo tags like this:
```vim
:lua require('orgmode').action("agenda.tags_todo", { tags = "personal" })
```

kristijanhusak:
You mean [this](https://github.com/bastibe/org-journal) ?

I'm not sure how much new code it will introduce, but if it's a bit too much, you should create it as a separate plugin. Idea is to have built in only what orgmode has in core, and anything additional should go as a plugin. 

I'll think about the plugins infrastructure and how we could expose some api for plugins. Do not hesitate to add some suggestions on https://github.com/kristijanhusak/orgmode.nvim/issues/26 regarding this topic. 

Once you have something working let me know to check it. I would love to expose a public API that will be stable, since this is still in beta, and some changes regarding the parsing will probably happen.

chuwy:
> You mean this ?

Yup.

I agree that public API should be the way to go. I'll start doing this separately and let you know what issues I stumble upon.

PR #32, opened 2021-07-06
Lazy Load with packer.nvim on Org filetype

shaunsingh:


kristijanhusak:
@shaunsingh are you sure it works? I got an issue [here](https://github.com/kristijanhusak/orgmode.nvim/issues/18#issue-933551086) where they say it's not working that way.

shaunsingh:
I've been using it with my configuration for a while now with no folding issues for other filetypes. However as the issue says, orgmode.nvim is already quite lightweight so it doesn't make much of a difference (around 2-3ms)

kristijanhusak:
OK, I just needed a confirmation that it works. Thanks.

akinsho:
*Unsolicited opinion incoming*: I think given that people trying out a plugin tend to just copy-paste the installation instructions, I think starting with a lazy loaded example is going to cause more problems than it solves since someone will copy-paste the ft loaded example and find that they can't use `<leader>oa` or `<leader>oc` till they open an org file which, if they are new to org mode which I think is maybe the case for the majority of vim users then they won't *have any*.

I think the lazy loaded option should be in addition to a normal installation if at all, which is much more common for most plugin installation instructions, rather than a lazy loaded example by default (which people might think is necessary)

kristijanhusak:
@akinsho thanks for the heads up. I thought plugin is fully functional when lazy loaded. I'll update readme.

akinsho:
@kristijanhusak for packer if you add a `ft` key then plugin will be in the `opt` directory for packages and not `packadd`ed till the right filetype command is hit available, so the global mappings won't work at all 😅 .

kristijanhusak:
@akinsho for some reason I thought packer is doing lazy loading some other way, but there's no other way :D I'll return old readme and put note about lazy loading.

shaunsingh:
> starting with a lazy loaded example is going to cause more problems than it solves since someone will copy-paste the ft loaded example and find that they can't use <leader>oa or <leader>oc till they open an org file which, if they are new to org mode which I think is maybe the case for the majority of vim users then they won't have any.

I didn't think of this scenario. As you said, it makes more sense to keep the README example as before then. Thank you!

Shrimpram:
> find that they can't use \<leader\>oa or \<leader\>oc till they open an org file

Packer lets you lazy load plugins on specific commands or keymaps. So for example, this plugin could be loaded on `keys = <leader>oa`. Check the [Specifying Plugins](https://github.com/wbthomason/packer.nvim#specifying-plugins) section in packer, it has explanations for all the options you can load a plugin with. This is just off the top of my head, I can open a PR for an updated lazy load in the readme once I've tested a solution that works.

akinsho:
@Shrimpram I'm aware of the `keys` and other lazy loading mechanisms. I guess the point I'm making is that the *vast* majority of plugins do not start by advertising how to lazy load them, tbh I think lazy loading should only really be done if a user has identified a real bottle neck after *profiling* their startup. In this case profiling shows that this plugin loads in an amount of time that a human could not possibly notice and so doing it is definitely not a required optimisation.

I think given that the README is a user's first introduction to a plugin, the installation process should be for the common use case and as simple and bare bones as possible just to get things started. If a power user wants to lazy load having already done some profiling, they can but I don't think it needs to be advertised on a plugin's landing page.

Regarding the behaviour of `<leader>oc` and `oa` using the keys key that is not expected packer behaviour (I contribute to the project) pressing the key should load the plugin and also then trigger the mapping again so the loaded plugin should immediately respond, rather than requiring a repeat key press.

I've also noticed an issue with the global mappings not triggering immediately but haven't had time to look into if it's a packer issue or my config specifically, so haven't raised an issue since I'm not sure where the issue is yet.

Shrimpram:
@akinsho Yeah that's fair, people love to copy and paste from Readmes, even when it's stated that what they're copying and pasting isn't a proper solution, or is only a hackish way of getting things done. In this case, given that the plugin itself loads in such little time trying to optimize it would be premature and best left up to users to decide if/how they want to do.

Re: what you said about packer. I've tried lazy loading a couple of my other plugins that take a significant amount of time to startup with both `keys` and `cmd` and found that only `cmd` works properly (loads & executes the plugin). What's odd is that the `keys` that I've selected are just remaps that call a command (e.g. <kbd>\<leader\>ut</kbd> just does `<CMD>UndotreeToggle`) yet the plugin is only loaded and executed when I directly specify the `cmd` and not the `keys`. I'll try to troubleshoot a bit and maybe raise an issue there.

PR #41, opened 2021-07-12
feat: Open capture in float window

tami5:
I'm so used to floats and I appreciate them, maybe others would find
this useful.

Issues:
- `C-c` won't close the window, just saves the captured content.
  (:close<CR> saves the captured content and closes the window without
  issues)

TODOs:
- [ ] make `C-c` or capture kill close the window.
- [ ] make capture menu float.
- [ ] expose window configuration to the user???, e.g. window fill size,
  border .. etc,
- [ ] refactor

### DIFF
--- a/lua/orgmode/capture/init.lua
+++ b/lua/orgmode/capture/init.lua
@@ -3,6 +3,8 @@ local utils = require('orgmode.utils')
 local config = require('orgmode.config')
 local Files = require('orgmode.parser.files')
 local Templates = require('orgmode.capture.templates')
+local winnr = nil -- HACK for Capture:kill
+
 vim.cmd[[augroup OrgCapture]]
 vim.cmd[[autocmd!]]
 vim.cmd[[augroup END]]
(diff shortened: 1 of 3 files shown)


kristijanhusak:
This is now available via [win_split_mode](https://github.com/nvim-orgmode/orgmode/commit/12809f9c76bb80bc2fed889f3d83cb3382c755ce) option. Here's a short description how to do it:

```lua
require('orgmode').setup({
  win_split_mode = function(name)
    local bufnr = vim.api.nvim_create_buf(false, true)
    --- Setting buffer name is required
    vim.api.nvim_buf_set_name(bufnr, name)

    local fill = 0.8
    local width = math.floor((vim.o.columns * fill))
    local height = math.floor((vim.o.lines * fill))
    local row = math.floor((((vim.o.lines - height) / 2) - 1))
    local col = math.floor(((vim.o.columns - width) / 2))

    vim.api.nvim_open_win(bufnr, true, {
      relative = "editor",
      width = width,
      height = height,
      row = row,
      col = col,
      style = "minimal",
      border = "rounded"
    })
  end

})
```

kristijanhusak:
I also added built in way to use float:

Default 70% width
```lua
win_split_mode = 'float'
```

Custom width:
```lua
win_split_mode = {'float', 0.9}
```

PR #59, opened 2021-07-28
Set calendar buffer as nonmodifiable

Shrimpram:
Currently users are able to modify the calendar buffer, this can lead to situations where they might accidentally delete parts of the calendar and leave it unusable. Emacs' org mode calendar works a bit differently but it is not editable either. Setting the calendar buffer as not modifiable after rendering it solves the issue.


kristijanhusak:
I don't know how I missed that. Thanks!

PR #60, opened 2021-08-03
Folded text respects org_hide_leading_stars

rahulsalvi:
This commit adds a better ```foldtext``` function that replaces leading stars with spaces when ```org_hide_leading_stars``` is true.

Example without this commit (Lines 2 and 5 are folded):
2021-08-02-214216_355x140_scrot

Same example with this commit:
2021-08-02-214126_358x140_scrot

Thanks


kristijanhusak:
Nice, thanks!

PR #79, opened 2021-09-05
Handle return for numbered lists

lukas-reineke:
I started working on this before you pushed your version

But my version also updates the numbers of all following items, when you add one in the middle.

To get this to work nicer, I also had to make a change to the treesitter grammar.
That PR is open here: https://github.com/milisims/tree-sitter-org/pull/5

As a bonus, this also handles multi line list items.


lua/orgmode/org/mappings.lua
```diff
@@ -246,8 +246,40 @@ function OrgMappings:handle_return(suffix)
...
     end
     if number_in_list then
-      vim.fn.append(vim.fn.line('.'), string.format('%s%d%s ', indent, tonumber(number_in_list) + 1, closer))
-      vim.fn.cursor(vim.fn.line('.') + 1, 0)
+      local next_sibling = item.node:parent()
+      local text_edits = {}
+      local counter = 1
+      local new_start
+      while next_sibling do
+        if next_sibling:type() == 'listitem' then
```
kristijanhusak, line 254: Can this ever not be `listitem`? Does it make sense to break out if this is false?

kristijanhusak (commented):
I'll wait for the upstream PR to be merged first before testing this out.

lua/orgmode/org/mappings.lua
lukas-reineke: `next_sibling` returns all nodes, named and unamed. So it can be the newline at the end of each list item.
I'll change it to `ts_utils.get_next_node`, that only returns named nodes, that is easier to understand.

lukas-reineke:
I added a bit more stuff.

All lists and headlines now support multiline. Doesn't matter what is nested in them. On enter they will create a new item on the same level they are, after the end of their section.

lua/orgmode/org/mappings.lua
```diff
@@ -218,37 +218,78 @@ function OrgMappings:handle_return(suffix)
   suffix = suffix or ''
   local item = Files.get_current_file():get_current_node()
   if item.type == 'headline' or item.node:parent():type() == 'headline' then
-    if item.type ~= 'headline' then
+    if item.type ~= 'stars' then
```
kristijanhusak, line 221: Why is this changed? `stars`, `item` and `tags` nodes are all on the same level, so this should handle return from all of those.
Also, below `section` variable will not return section node if you are on `item` or `tags` node, since this will go one parent up, and that will go 2 more parents up, which can end up on the top level document. Maybe I'm missing something, didn't tested it yet.

lua/orgmode/org/mappings.lua
```diff
@@ -218,37 +218,78 @@ function OrgMappings:handle_return(suffix)
...
+    vim.fn.cursor(end_row + 1, 0)
     return vim.cmd([[startinsert!]])
   end
-  if item.type == 'list' or item.type == 'listitem' then
-    if item.type == 'list' then
-      vim.fn.cursor(vim.fn.line('.'), vim.fn.col('$'))
-      item = Files.get_current_file():get_current_node()
+
+  if item.type == 'list' then
+    vim.cmd([[normal ^]])
```
kristijanhusak, line 233: I think this should be `normal! ^`, since `normal` uses mapped key without `bang`. Also, does `startofline` option affect this functionality?

lua/orgmode/utils.lua
```diff
@@ -261,7 +261,7 @@ function utils.get_ts_matches(query, node, file_content)
 end
 
 function utils.get_node_text(node, content)
-  if not node then
+  if not node or node == '' then
```
kristijanhusak, line 264: When is node empty string? This function should either get a ts node or nil.

kristijanhusak (dismissed):

lua/orgmode/org/mappings.lua
lukas-reineke: The logic before was a bit broken, but still worked.
If it was not a `headline` type, it does `ts_utils.goto_node(item.node:parent())`, which goes to the start of the headline. But then `Files.get_current_file():get_current_node()` returns the smallest node at that point, which is `stars`.
But that's fine, as stars also has the level.
But now for the change I made, I need to know which level I am on, so I can go up to the section. So if the cursor is between the stars and the line, it is type `headline`. In that case I do the same as before to change `item` to `stars`.

I agree, it is not really obvious. Maybe the code should just walk up the parent until it reaches `section` to be easier to understand.

lua/orgmode/org/mappings.lua
lukas-reineke: good point about `!`
I think `startofline` should not affect this.

lua/orgmode/utils.lua
lukas-reineke: I don't really know why this happened. This is unrelated to the rest of my changes.
It had a state where this throws errors, even on the current `tree-sitter` branch.

I'll see if I can reproduce this, and see where it comes from

lua/orgmode/org/mappings.lua
kristijanhusak: > I agree, it is not really obvious. Maybe the code should just walk up the parent until it reaches section to be easier to understand.


I like this. You can introduce a helper in utils that would do this traversing to closest parent, something like good ol' Jquery's [parentsUntil](https://api.jquery.com/parentsuntil/)

lua/orgmode/utils.lua
kristijanhusak: Yeah, please try to reproduce it. If it's not related to this PR, i'll fix it separately.

lukas-reineke: It happens when there is an empty headline.

just opening a file like this throws.

```org
*
```

kristijanhusak: Thanks, I'll take a look.

kristijanhusak: Fixed it. I was using default as empty string instead of nil for todo keyword node, which is missing in this case.

kristijanhusak (commented):
This looks ok for now. Once grammar PR is merged, I'll give this a test and merge it. Thanks!

kristijanhusak:
Since upstream PR is now merged, I'll start testing this manually tomorrow. 
Today I added some UI tests that does testing of the mappings. I would appreciate if you would rebase and add a test for this, so we have it covered. Here's the current test that tests my simple version of numbered list increase: https://github.com/kristijanhusak/orgmode.nvim/blob/tree-sitter/tests/plenary/ui/mappings_spec.lua#L330. Feel free to add any other tests if you think is needed.

lukas-reineke:
oh `next_named_sibling` is Neovim new in 0.6.
I'll revert this back to use the `ts_util` helper

lukas-reineke:
I added some tests, let me know if there is anything else you'd like tested.

Also about the tests, I think it would be better to have dedicated fixtures for tests that jump around the whole file.
It is inconvenient to have to update unrelated tests when you want to add a new test case.

kristijanhusak:
> I added some tests, let me know if there is anything else you'd like tested.
> 
> Also about the tests, I think it would be better to have dedicated fixtures for tests that jump around the whole file.
> It is inconvenient to have to update unrelated tests when you want to add a new test case.

Thanks! Looks great!

Yeah, I noticed that issue with tests. I'll switch to creating a fixture for each separate test. It will be slower, but much less fragile and unstable.

Once I do some manual tests around this I'll merge it (hopefully today).

kristijanhusak:
I gave this a test, works great!
Thank you for your contribution!

PR #80, opened 2021-09-07
headline with four * trigger bold syntax

lukas-reineke:
This causes a lot of weird behaviour, especially when
`org_hide_emphasis_markers` is true.

To fix it, change the `org_bold` syntax to not allow bold statements that start with two `*`.


kristijanhusak:
Thanks! Not using `org_hide_emphasis_markers` myself so I missed it.

PR #85, opened 2021-09-10
bug: rename `itemtag` to `description`

lukas-reineke:
I forgot to rename this here after I renamed it in tree-sitter-org

Bonus fix, enter in listitems if the cursor is between the bullet and the item

I wrote test cases for both of them.


kristijanhusak:
Thanks!

PR #89, opened 2021-09-14
Decapitalize keywords in completion, add syntax highlighting for decapitalized blocks

shaunsingh:
lowercase keywords are the norm in org-mode, uppercase is just used in the manuals to make them stand out more. This changes the completion to use lowercase instead of uppercase 

TODO: Add back syntax highlighting for uppercase keywords

### DIFF
--- a/lua/orgmode/org/syntax.lua
+++ b/lua/orgmode/org/syntax.lua
@@ -12,7 +12,7 @@ local function load_code_blocks()
   end
 
   for _, ft in ipairs(orgfile.source_code_filetypes) do
-    vim.cmd(string.format([[syntax include @orgmodeBlockSrc%s syntax/%s.vim]], ft, ft))
+    vim.cmd(string.format([[silent! syntax include @orgmodeBlockSrc%s syntax/%s.vim]], ft, ft))
     vim.cmd([[unlet! b:current_syntax]])
   end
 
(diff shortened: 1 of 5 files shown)


lua/orgmode/org/syntax.lua
```diff
@@ -19,7 +19,7 @@ local function load_code_blocks()
   for _, ft in ipairs(orgfile.source_code_filetypes) do
     vim.cmd(
       string.format(
-        [[syntax region orgmodeBlockSrc%s matchgroup=comment start="^\s*#+BEGIN_SRC\ %s\s*$" end="^\s*#+END_SRC\s*$" keepend contains=@orgmodeBlockSrc%s,org_block_delimiter]],
+        [[syntax region orgmodeBlockSrc%s matchgroup=comment start="^\s*#+begin_src\ %s\s*$" end="^\s*#+END_SRC\s*$" keepend contains=@orgmodeBlockSrc%s,org_block_delimiter]],
```
kristijanhusak, line 22: You missed end src here

kristijanhusak (changes requested):

kristijanhusak:
Is there any specific documentation about this? I know lowercase also works, but I didn't knew it was norm. I don't see why all of the manual would use uppercase as an example if it's not norm?

Also, general issue with this is that we should preserve both uppercase and lowercase for certain functionality (highlighting), which current change doesn't cover.

kristijanhusak:
You can update all syntax matches to start with `\c`, that will make it case insensitive.

TravonteD:
@kristijanhusak this pr can be closed

PR #100, opened 2021-10-10
Add clocking time feature.

kristijanhusak:


kristijanhusak:
@ourigen you have to use `tree-sitter` branch, and I would advise you to switch. `tree-sitter` branch will become master at some point, but I'm not sure when. It's still in a testing phase.
All new features are added to this branch.

PR #112, opened 2021-10-24
Enhance file: link support

levouh:
As a summary:

- Complete org files on 'file:'

https://user-images.githubusercontent.com/31262046/138616359-e4f256f7-af76-4983-b5a1-945f5e56542b.mp4

- Complete headers based on current 'file:' context

https://user-images.githubusercontent.com/31262046/138616389-346f8e01-4d5f-4826-b19c-db7a850834c9.mp4

- Support jumping to specific section in file from hyperlink

https://user-images.githubusercontent.com/31262046/138616420-61c4f986-d2ab-4d07-9a0b-002f05058fa3.mp4

---

For original _Emacs_ functionality, see [here](https://orgmode.org/manual/External-Links.html). I still have some things to review myself and clean up before I would consider this "non-WIP", but the main functionality is there.

### DIFF
--- a/lua/orgmode/org/autocompletion/omni.lua
+++ b/lua/orgmode/org/autocompletion/omni.lua
@@ -9,23 +9,38 @@ local data = {
   metadata = { 'DEADLINE:', 'SCHEDULED:', 'CLOSED:' },
 }
 
-local directives = { rgx = vim.regex([[^\#+\?\w*$]]), line_rgx = vim.regex([[^\#\?+\?\w*$]]), list = data.directives }
+local directives = {
+  line_rgx = vim.regex([[^\#\?+\?\w*$]]),
+  rgx = vim.regex([[^\#+\?\w*$]]),
+  list = data.directives,
+}
+
 local begin_blocks = {
-  rgx = vim.regex([[\(^\s*\)\@<=\#+\?\w*$]]),
   line_rgx = vim.regex([[^\s*\#\?+\?\w*$]]),
+  rgx = vim.regex([[\(^\s*\)\@<=\#+\?\w*$]]),
   list = data.begin_blocks,
 }
+
 local properties = {
   line_rgx = vim.regex([[\(^\s\+\|^\s*:\?$\)]]),
   rgx = vim.regex([[\(^\|^\s\+\)\@<=:\w*$]]),
+  extra_cond = function(line, _)
+    return not string.find(line, 'file:.*$')
+  end,
   list = data.properties,
 }
+
 local links = {
-  line_rgx = vim.regex([[\(\(^\|\s\+\)\[\[\)\@<=\(\*\|\#\)\?\(\w\+\)\?]]),
-  rgx = vim.regex([[\(\*\|\#\)\?\(\w\+\)\?$]]),
+  line_rgx = vim.regex([[\(\(^\|\s\+\)\[\[\)\@<=\(\*\|\#\|file:\)\?\(\(\w\|\/\|\.\|\\\|-\|_\|\d\)\+\)\?]]),
+  rgx = vim.regex([[\(\*\|\#\|file:\)\?\(\(\w\|\/\|\.\|\\\|-\|_\|\d\)\+\)\?$]]),
   fetcher = Hyperlinks.find_matching_links,
 }
-local metadata = { rgx = vim.regex([[\(\s*\)\@<=\w\+$]]), list = data.metadata }
+
+local metadata = {
+  rgx = vim.regex([[\(\s*\)\@<=\w\+$]]),
+  list = data.metadata,
+}
+
 local tags = {
   rgx = vim.regex([[:\([0-9A-Za-z_%@\#]*\)$]]),
   fetcher = function()
@@ -38,6 +53,9 @@ local tags = {
 local filetags = {
   line_rgx = vim.regex([[\c^\#+FILETAGS:\s\+]]),
   rgx = vim.regex([[:\([0-9A-Za-z_%@\#]*\)$]]),
+  extra_cond = function(line, _)
+    return not string.find(line, 'file:.*$')
+  end,
   fetcher = function()
     return vim.tbl_map(function(tag)
       return ':' .. tag .. ':'
@@ -46,8 +64,8 @@ local filetags = {
 }
 
 local todo_keywords = {
-  rgx = vim.regex([[\(^\(\*\+\s\+\)\?\)\@<=\w*$]]),
   line_rgx = vim.regex([[^\*\+\s\+\w*$]]),
+  rgx = vim.regex([[\(^\(\*\+\s\+\)\?\)\@<=\w*$]]),
   fetcher = function()
     return config:get_todo_keywords().ALL
   end,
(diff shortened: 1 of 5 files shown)


lua/orgmode/org/autocompletion/omni.lua
```diff
@@ -9,23 +9,38 @@ local data = {
...
+  extra_cond = function(line, _)
+    return not string.find(line, 'file:$')
+  end,
   list = data.properties,
 }
+
 local links = {
-  line_rgx = vim.regex([[\(\(^\|\s\+\)\[\[\)\@<=\(\*\|\#\)\?\(\w\+\)\?]]),
-  rgx = vim.regex([[\(\*\|\#\)\?\(\w\+\)\?$]]),
+  line_rgx = vim.regex([[\(\(^\|\s\+\)\[\[\)\@<=\(\*\|\#\|file:\)\?\(\w\+\)\?]]),
```
kristijanhusak, line 34: You should update regexes to this:
```lua
  line_rgx = vim.regex([[\(\(^\|\s\+\)\[\[\)\@<=\(\*\|\#\|file:\)\?\(\(\w\|\/\|\.\|\\\)\+\)\?]]),
  rgx = vim.regex([[\(\*\|\#\|file:\)\?\(\(\w\|\/\|\.\|\\\)\+\)\?$]]),
```
It's a bit crazy, but it will allow autocompleting part of the filename.
For example, if you type this:

```
[[file:/home/user/orgfiles/t]]
```

It will autocomplete only fiels that start with `t`. Needs more changes below though, I leave the comment.

lua/orgmode/org/autocompletion/omni.lua
```diff
@@ -38,6 +53,9 @@ local tags = {
 local filetags = {
   line_rgx = vim.regex([[\c^\#+FILETAGS:\s\+]]),
   rgx = vim.regex([[:\([0-9A-Za-z_%@\#]*\)$]]),
+  extra_cond = function(line, _)
+    return not string.find(line, 'file:$')
```
kristijanhusak, line 57: Update `file:$` to `file:.*$`, since `filetags` overtakes the context after `file:` is typed.

lua/orgmode/org/autocompletion/omni.lua
```diff
@@ -9,23 +9,38 @@ local data = {
...
   line_rgx = vim.regex([[^\s*\#\?+\?\w*$]]),
+  rgx = vim.regex([[\(^\s*\)\@<=\#+\?\w*$]]),
   list = data.begin_blocks,
 }
+
 local properties = {
   line_rgx = vim.regex([[\(^\s\+\|^\s*:\?$\)]]),
   rgx = vim.regex([[\(^\|^\s\+\)\@<=:\w*$]]),
+  extra_cond = function(line, _)
+    return not string.find(line, 'file:$')
```
kristijanhusak, line 28: Update `file:$` to `file:.*$`, since `filetags` overtakes the context after `file:` is typed.

lua/orgmode/org/hyperlinks.lua
```diff
@@ -45,32 +91,39 @@ function Hyperlinks.find_by_dedicated_target(base, skip_mapping)
...
-  local prefix = base:sub(1, 1)
-  if prefix == '#' then
-    return Hyperlinks.find_by_custom_id_property(base, skip_mapping)
+function Hyperlinks.find_matching_links(ctx)
+  ctx = ctx or {}
+  ctx.base = ctx.base and vim.trim(ctx.base) or nil
+
+  update_hyperlink_ctx(ctx)
+
+  if ctx.base == 'file:' and not ctx.skip_add_prefix then
```
kristijanhusak, line 113: Update `ctx.base == 'file:'` to `ctx.base:find('^file:')`, since base will now contain items even after `file:`. `find_by_filepath` will properly filter that base.

lua/orgmode/org/hyperlinks.lua
```diff
@@ -2,33 +2,79 @@ local Files = require('orgmode.parser.files')
...
+    end
+
+    ctx.base = hyperlinks_ctx.headline or hyperlinks_ctx.custom_id or ctx.base
+  end
+
+  ctx.hyperlinks = hyperlinks_ctx
+end
+
+function Hyperlinks.find_by_filepath(_)
+  local filenames = Files.filenames()
```
kristijanhusak, line 40: Add additional filtering to this:
```lua
function Hyperlinks.find_by_filepath(ctx)
  local filenames = Files.filenames()
  local file_base = ctx.base:gsub('^file:', '')
  if vim.trim(file_base) ~= '' then
  filenames = vim.tbl_filter(function(f)
    return f:find('^'..file_base)
  end, filenames)
  end

  -- Outer checks already filter cases where `ctx.skip_add_prefix` is truthy,
  -- so no need to check it here
  return vim.tbl_map(function(path)
    return 'file:' .. path
  end, filenames)
end
```
Idea is to grab everything after the base `file:` and filter only filenames that are matching the input, like I explained in the comment for regex change.

kristijanhusak (changes requested):
I gave this a test. It's working mostly fine, but I would like to also have more filtering after user starts typing the file. 
For example, if you have `~/orgmodes/todos.org`, `~/orgmodes/test.org` and `~/orgmodes/work.org`, and if you type `[[file:/home/user/orgmodes/t]]` I would expect to get only files that start with `t`. I explained more in the review comments.

lua/orgmode/org/autocompletion/omni.lua
levouh: I like it, this is a lot more intuitive.

levouh:
Rebased it into a single commit @kristijanhusak, in hindsight that makes it a bit harder to review "net new" changes so apologies for that. Versus what existed before (that you'd reviewed), I've added the changes from your comments above and one other small change to ensure we modify the filename to be an absolute path before handing it off to `Files.get(...)` such that both relative and include things like `~` as part of the path.

kristijanhusak:
Everything looks good. Thanks for contributing!

cands:
Thanks for this important contribution, file links are most useful. I wonder if you could extend so that some more variants for file links are supported, as seen in the examples [here](https://orgmode.org/manual/External-Links.html). In particular, for my use case the possibility to use custom-id for the links is fundamental, since those links can be made unique and persistent (whereas the title of the header may be changed often, rendering the link broken). There doesn't seem to be a way to link using ‘file:projects.org::#custom-id’ or complete the custom-id strings.

By the way, thanks very much @kristijanhusak for making this totally awesome plugin, I have been thinking to use Orgmode for quite some time but did not want to switch editor, now I got the chance and it is great. Your plugin works very well, it is very smooth and quick to work with, splendid work!

levouh:
This would be relatively easy to add, can you make a new issue with the specific request @cands (and mark it as an enhancement)? I can take a look from there.

kristijanhusak:
@cands @levouh there was actually a small bug with finding the headlines by custom id. Should be fixed now.

cands:
Thanks for the bug fix @kristijanhusak It fixed the possibility to use custom-id, which is great! I don't think the possibility to use id strings (by Orgmode default is an UUID string that becomes globally unique and can be found even when the headline is moved around in the different org files). Also, we would need the ability to store links, see 'M-x org-store-link' [Handling Links (The Org Manual)](https://orgmode.org/manual/Handling-Links.html) to easily retrieve links by custom-id and id (from org files and agendas). Do you want me to open a new issue for this?

kristijanhusak:
@cands yeah, open up a separate issue, and please write an explanation how it exactly works in emacs orgmode. I never used it, and it's a bit hard to understand it from this documentation. I managed to store the links and insert it via prompt, but I didn't see any UUIDs or anything.

PR #123, opened 2021-10-30
Remove excess `*` in drawer fold text

ourigen:
This minor change removes the extra `*` in a folded `:PROPERTIES:` drawer

Before

After


kristijanhusak:
Thanks!

Issue #124, opened 2021-10-31
orgmode_hide_emphasis_markers with tree-sitter


levouh:
### Are you using "tree-sitter" branch?

Yes

### Describe the bug

More of a question, but the idea that the feature isn't working would seemingly mean that this is a bug? Not familiar with _Treesitter_ at all, but when using syntax markers in the a file the `/.../` markers should be concealed, right? I don't see any issues 

### Steps to reproduce

1. Setup `orgmode.nvim` with `org_hide_emphasis_markers` as `true`
2. Open an `org` file
3. Type text surrounded by `/.../` markers

### Expected behavior

Markers for italic, bold, etc. are hidden.

### Screenshots and recordings

![image](https://user-images.githubusercontent.com/31262046/139603021-f3195b85-2743-4c7a-a089-c4c8b9343bae.png)

### OS / Distro

Linux

### Neovim version/commit

NVIM v0.6.0-dev+518-g1dbbaf89b

### Additional context

Again, not sure what the expected behavior here is.

kristijanhusak:
Are you sure it works for other things (bold, strikethrough, etc). Works fine for me:

![screenshot](https://i.imgur.com/ZEgzojO.png)

What is your `&conceallevel` ?

kristijanhusak:
@levouh can we close this now?

PR #127, opened 2021-11-02
Docs update and notification fix

levouh:
Two changes:

1. Add note in doc. about `conceallevel` being required to hide emphasis markers, "fixes" #124
2. Fix default checks for 0/1 instead of Lua-style boolean for `vim.fn.executable` (used with notifications in this case, but checked elsewhere and other places look good from what I can tell too)

### DIFF
--- a/DOCS.md
+++ b/DOCS.md
@@ -133,6 +133,7 @@ vim.cmd[[autocmd ColorScheme * hi link OrgHideLeadingStars MyCustomHlGroup]]
 *type*: `boolean`<br />
 *default value*: `false`<br />
 Conceal bold/italic/underline/code/verbatim markers.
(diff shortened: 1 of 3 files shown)


levouh:
Don't merge this, something is up with my `md2vim` it looks like.

levouh:
Not 100% what is up with `md2vim`, do you use a special build @kristijanhusak? I just downloaded the one from [here](https://github.com/FooSoft/md2vim).

kristijanhusak:
Good catch. I'm using Linux so first if always worked for me.

I'm downloading latest with go: `go get github.com/FooSoft/md2vim@latest`. I think prepackaged builds are outdated.

levouh:
Should be good to go now.

kristijanhusak:
Thanks!

Issue #134, opened 2021-11-03
Is there any way to denote folded nodes with `...` instead of the grey highlight


lanza:
I find the the current highlighting setup pretty confusing as a folded level1 has the same color as a folded level2 node yet a different color as an unfolded level1 node. Is there any way to disable this and potentially add emac's typical `...` setup?

kristijanhusak:
Vim/Neovim doesn't allow having different folding colors per any condition. All folds are colored the same.

chuwy:
A relevant issue to follow: https://github.com/neovim/neovim/issues/12649

ourigen:
Folds are denoted with `...` now in the `tree-sitter` branch, or you can configure it with any string set to `org_ellipsis` in the newest commit. Highlighting however is still limited by Neovim

kristijanhusak:
You can use https://github.com/kevinhwang91/nvim-ufo until there's built-in support. Just make sure to disable any provider for org filetype:

```lua
  require('ufo').setup({
    provider_selector = function(_, filetype)
      if filetype == 'org' then
        return ''
      end
      return { 'treesitter', 'indent' }
    end,
  })
```

kristijanhusak:
This now works as expected using `master` branch and Neovim nightly.
![screenshot_2024_02_25_18_04_04](https://github.com/nvim-orgmode/orgmode/assets/1782860/7a09e471-1cca-4e30-b0d3-90963361aa44)

PR #137, opened 2021-11-04
feat: add org-ellipsis

ourigen:
I should have made this configurable from the start but now users can configure what to replace fold markers with. This replicates the `org-ellipsis` setting [available on Emacs](https://emacs.stackexchange.com/a/10986).

Also partially addresses #134


kristijanhusak:
Oh, wasn't aware of it. Thanks!

PR #148, opened 2021-11-11
Enhance template expansion

TravonteD:
Adds support for the following template expansions:
- username
- clipboard
- calling file
- prompts
- expressions

### DIFF
--- a/lua/orgmode/capture/templates.lua
+++ b/lua/orgmode/capture/templates.lua
@@ -47,6 +47,7 @@ function Templates:compile(template)
   end
   content = self:_compile_dates(content)
   content = self:_compile_prompts(content)
+  content = self:_compile_expressions(content)
   for expansion, compiler in pairs(expansions) do
     if content:match(vim.pesc(expansion)) then
       content = content:gsub(vim.pesc(expansion), compiler())
@@ -105,6 +106,18 @@ function Templates:_compile_prompts(content)
       end
     end
     content = content:gsub(vim.pesc(exp), response)
+  end
+  return content
+end
+
+function Templates:_compile_expressions(content)
+  for exp in content:gmatch('%%%b()') do
+    local snippet = exp:match('%((.*)%)')
+    local func = load(snippet)
+    local ok, response = pcall(func)
+    if ok then
+      content = content:gsub(vim.pesc(exp), response)
+    end
   end
   return content
 end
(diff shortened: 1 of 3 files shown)


lua/orgmode/capture/templates.lua
```diff
@@ -72,4 +88,47 @@ function Templates:_compile_dates(content)
...
+    local title, default = parts[1], parts[2]
+    local response
+    if #parts > 2 then
+      local completion_items = vim.list_slice(parts, 3, #parts)
+      local prompt = { string.format('%s [%s]:', title, default) }
+      for i, item in ipairs(completion_items) do
+        table.insert(prompt, i .. '. ' .. item)
+      end
+      local response_number = vim.fn.inputlist(prompt)
+      response = response_number == 0 and default or completion_items[response_number + 1]
```
kristijanhusak, line 106: You don't need `+ 1` here since prompt title is added to the `prompt` variable, not `completion_items` variable. This currently causes `1` to select option `2`.

tests/plenary/capture/templates_spec.lua
```diff
@@ -3,8 +3,11 @@ local Templates = require('orgmode.capture.templates')
 describe('Capture template', function()
   local templates = Templates:new()
   it('should compile expression', function()
+    ---Backup and restore the clipboard
+    local clip_backup = vim.fn.getreg('+')
+    vim.fn.setreg('+', 'test')
     local result = templates:compile({
-      template = '* TODO\n%<%Y-%m-%d>\n%t\n%T--%T\n%<%H:%M>\n%<%A>',
+      template = '* TODO\n%<%Y-%m-%d>\n%t\n%T--%T\n%<%H:%M>\n%<%A>\n%x',
```
kristijanhusak, line 10: Would be awesome to write a test for `%(EXP)`. I know writing tests for prompt is not really possible, but this one should work.

kristijanhusak (changes requested):
Awesome work, thanks!

There are just 2 minor things to be addressed and this goes straight to master!

kristijanhusak:
Thanks!

kristijanhusak:
@TravonteD Note that I slightly updated your implementation for prompt. I checked how Emacs is working, and it gives an input prompt with autocompletion, instead of inputlist. I tweaked it to work the same way now. Example:

With this prompt settings: `%^{Choose|the_default|first|second|third}`

You get this:

screenshot

PR #151, opened 2021-11-13
feat: Edit source blocks in a buffer with the associated filetype

### LINKED ISSUE #146: Opening code blocks in a temporal buffer

Yes

### Orgmode link

By the end of the Literal-Exampels section, there's a keybind for what seems to be opening code blocks in a separate buffer.
https://orgmode.org/org.html#Literal-Examples

### Feature value

Editing in a separate buffer with its own filetype is much more comfortable than doing it within the block. Furthermore, that makes it easier to interoperate with other plugins - just the top off my head, autocompletion and diagnostics via LSP would be trivially included. 

### Additional context

_No response_

levouh:
This resolves #146.

- [x] Add/fix tests
- [x] Check parity with _Emacs_ functionality
- [x] Investigate using `extmarks`

Right now there's a lot going on in a function that is really only supposed to be handling the mapping, so if you'd rather have this put somewhere else let me know.

The first iteration used an `acwrite` buffer which ends up being a bit cleaner as we aren't creating unecessary files, but most language servers (or at least things in `nvim-lspconfig`) expect a file on disk (e.g. such that a root directory can be found). From this then, I've just gone with `:h tempname()` which solves those problems and is mostly no different.

For now the default is using a simple split as a floating window has some issues with the completion menu (it ends up covering the current cursor position) and I imagine the main use case here is for language servers or similar.

Quick demo:

https://user-images.githubusercontent.com/31262046/141659086-6a954561-cca1-4a4f-bbc8-2a6e9a3f0c6e.mp4

---

Side note, I did some additional edits to have the `utils.echo_*` functions use `:h nvim_echo` only if `:h vim.notify` doesn't exist to follow "standards" but didn't include it in this _PR_ as it is a bit out of scope. I can add that back here if that's something you are alright with changing.

### DIFF
--- a/DOCS.md
+++ b/DOCS.md
@@ -9,8 +9,9 @@
    2. [Agenda mappings](#agenda-mappings)
    3. [Capture mappings](#capture-mappings)
    4. [Org mappings](#org-mappings)
-   5. [Text objects](#text-objects)
-   6. [Dot repeat](#dot-repeat)
+   5. [Edit Src mappings](#edit-src)
+   6. [Text objects](#text-objects)
+   7. [Dot repeat](#dot-repeat)
 3. [Document Diagnostics](#document-diagnostics)
 4. [Autocompletion](#autocompletion)
 5. [Abbreviations](#abbreviations)
@@ -146,7 +147,6 @@ Marker used to indicate a folded headline.
 When set to `time`(default), adds `CLOSED` date when marking headline as done.<br />
 When set to `false`, it is disabled.
 
-
 #### **org_highlight_latex_and_related**
 *type*: `string|nil`<br />
 *default value*: `nil`<br />
@@ -165,6 +165,19 @@ Possible values:
 Possible values:
 * `indent` - Use default indentation that follows headlines/checkboxes/previous line indent
 * `noindent` - Disable indentation. All lines start from 1st column
+
+#### **org_src_window_setup**
+*type*: `string|function`<br />
+*default value*: "top 16new"<br />
+If the value is a string, it will be run directly as input to `:h vim.cmd`, otherwise if the value is a function it will be called. Both
+values have the responsibility of opening a buffer (within a window) to show the special edit buffer. The content of the buffer will be
+set automatically, so this option only needs to handle opening an empty buffer.
+
+#### **org_edit_src_content_indentation**
+*type*: `number`<br />
+*default value*: 0<br />
+The indent value for content within `SRC` block types beyond the existing indent of the block itself. Only applied when exiting from
+an `org_edit_special` action on a `SRC` block.
 
 #### **org_custom_exports**
 *type*: `table`<br />
@@ -623,6 +636,11 @@ using the '+' character. Example: `file:/home/user/.config/nvim/init.lua +10`
 * Headline with `CUSTOM_ID` property (starts with `#`)
 * Fallback: If file path, opens the file, otherwise, tries to find the Headline title.
 When date is under the cursor, open the agenda for that day.<br />
+#### **org_edit_special**
+*mapped to*: `<Leader>o'`<br />
+Open a source block for editing in a temporary buffer of the associated `filetype`.<br />
+This is useful for editing text with language servers attached, etc. When the buffer is closed, the text of the underlying source block in the original Org file is updated.
+*Note that if the Org file that the source block comes from is edited before the special edit buffer is closed, the edits will not be applied. The special edit buffer contents can be recovered from :messages output*
 #### **org_cycle**
 *mapped to*: `<TAB>`<br />
 Cycle folding for current headline
@@ -738,6 +756,22 @@ require('orgmode').setup({
   }
 })
 ```
+
+### Edit Src
+
+Mappings applied when editing a `SRC` block content via `org_edit_special`.
+
+#### **org_edit_src_abort**
+*mapped to*: `<Leader>ok`<br />
+Abort changes made to temporary buffer created from the content of a `SRC` block, see above.<br />
+
+#### **org_edit_src_save**
+*mapped to*: `<Leader>ow`<br />
+Apply changes from the special buffer to the source Org buffer<br />
+
+#### **org_edit_src_show_help**
+*mapped to*: `g?`<br />
+Show help within the temporary buffer used to edit the content of a `SRC` block.<br />
 
 ### Text objects
 
@@ -919,6 +953,7 @@ For adding/changing TODO keyword colors see [org-todo-keyword-faces](#org_todo_k
 
 * The following use vanilla _Vim_ syntax matching, and will work without _Treesitter_ highlighting enabled.
 
+`OrgEditSrcHighlight`: The highlight for the source content in an _Org_ buffer while it is being edited in an edit special buffer
 `OrgAgendaDeadline`: A item deadline in the agenda view
 `OrgAgendaScheduled`: A scheduled item in the agenda view
 `OrgAgendaScheduledPast`: A item past its scheduled date in the agenda view
--- a/lua/orgmode/config/defaults.lua
+++ b/lua/orgmode/config/defaults.lua
@@ -35,6 +35,8 @@ return {
     heading = true,
     plain_list_item = false,
   },
+  org_src_window_setup = 'top 16new',
+  org_edit_src_content_indentation = 0,
   diagnostics = true,
   notifications = {
     enabled = false,
@@ -101,6 +103,7 @@ return {
       org_todo_prev = 'ciT',
       org_toggle_checkbox = '<C-Space>',
       org_open_at_point = '<Leader>oo',
+      org_edit_special = [[<Leader>o']],
       org_cycle = '<TAB>',
       org_global_cycle = '<S-TAB>',
       org_archive_subtree = '<Leader>o$',
@@ -132,6 +135,11 @@ return {
       org_clock_goto = '<leader>oxj',
       org_set_effort = '<leader>oxe',
       org_show_help = 'g?',
+    },
+    edit_src = {
+      org_edit_src_abort = '<Leader>ok',
+      org_edit_src_save = '<Leader>ow',
+      org_edit_src_show_help = 'g?',
     },
     text_objects = {
       inner_heading = 'ih',
--- a/lua/orgmode/org/mappings.lua
+++ b/lua/orgmode/org/mappings.lua
@@ -1,13 +1,14 @@
-local ts_utils = require('nvim-treesitter.ts_utils')
 local Calendar = require('orgmode.objects.calendar')
 local Date = require('orgmode.objects.date')
-local TodoState = require('orgmode.objects.todo_state')
-local PriorityState = require('orgmode.objects.priority_state')
-local Hyperlinks = require('orgmode.org.hyperlinks')
-local utils = require('orgmode.utils')
+local EditSpecial = require('orgmode.objects.edit_special')
 local Files = require('orgmode.parser.files')
-local config = require('orgmode.config')
 local Help = require('orgmode.objects.help')
+local Hyperlinks = require('orgmode.org.hyperlinks')
+local PriorityState = require('orgmode.objects.priority_state')
+local TodoState = require('orgmode.objects.todo_state')
+local config = require('orgmode.config')
+local ts_utils = require('nvim-treesitter.ts_utils')
+local utils = require('orgmode.utils')
 
 ---@class OrgMappings
 ---@field capture Capture
@@ -488,6 +489,16 @@ end
 
 function OrgMappings:show_help()
   return Help.show()
+end
+
+function OrgMappings:edit_special()
+  local edit_special = EditSpecial:new()
+  edit_special:init_in_org_buffer()
+  edit_special:init()
+end
+
+function OrgMappings:_edit_special_callback()
+  EditSpecial:new():done()
 end
 
 function OrgMappings:open_at_point()
--- a/doc/orgmode.txt
+++ b/doc/orgmode.txt
@@ -20,9 +20,11 @@ CONTENTS                                                        *orgmode-content
             1.1.1.10. org_log_done..........................|orgmode-org_log_done|
             1.1.1.11. org_highlight_latex_and_related.|orgmode-org_highlight_latex_and_related|
             1.1.1.12. org_indent_mode....................|orgmode-org_indent_mode|
-            1.1.1.13. org_custom_exports..............|orgmode-org_custom_exports|
-            1.1.1.14. org_time_stamp_rounding_minutes.|orgmode-org_time_stamp_rounding_minutes|
-            1.1.1.15. org_blank_before_new_entry.|orgmode-org_blank_before_new_entry|
+            1.1.1.13. org_src_window_setup..........|orgmode-org_src_window_setup|
+            1.1.1.14. org_edit_src_content_indentation.|orgmode-org_edit_src_content_indentation|
+            1.1.1.15. org_custom_exports..............|orgmode-org_custom_exports|
+            1.1.1.16. org_time_stamp_rounding_minutes.|orgmode-org_time_stamp_rounding_minutes|
+            1.1.1.17. org_blank_before_new_entry.|orgmode-org_blank_before_new_entry|
         1.1.2. Agenda settings...........................|orgmode-agenda_settings|
             1.1.2.1. org_deadline_warning_days.|orgmode-org_deadline_warning_days|
             1.1.2.2. org_agenda_span.....................|orgmode-org_agenda_span|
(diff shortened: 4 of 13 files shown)


lua/orgmode/org/mappings.lua
```diff
@@ -462,6 +462,147 @@ function OrgMappings:show_help()
   return Help.show()
 end
 
+function OrgMappings:edit_src_block()
+  local start_bufnr = vim.api.nvim_get_current_buf()
+  local file = Files.get_current_file()
+  local pos = vim.api.nvim_win_get_cursor(0)
+
+  local get_blocks = function()
```
kristijanhusak, line 470: I don't think there's need to query all blocks in a document to find the right one. You can just get the current node via `file:get_node_at_cursor()` and traverse up the parent until you find the src block. Also consider that someone might do this on the exact SRC node so there's no need to traverse in that case.

lua/orgmode/org/mappings.lua
```diff
@@ -462,6 +462,147 @@ function OrgMappings:show_help()
...
+    },
+  })
+
+  if not bufnr or not vim.api.nvim_buf_is_valid(bufnr) then
+    utils.echo_info('Cancel editing source block')
+    return
+  end
+
+  local funcname = string.format('_org_edit_src_block_%s', bufnr)
+  _G[funcname] = function(wipe)
```
kristijanhusak, line 543: I don't think it's needed to create a global function for each buffer. It should work the same for all edits. Open in tmp buffer, save, return to previous buffer and update the block. 
For the state that you need, just put it in the buffer variable and use it when needed. You can save the file+node range in the buffer variable so you know what to update after.
Functionality should be really similar to the "capture" functionality. See how capture functionality is done https://github.com/kristijanhusak/orgmode.nvim/blob/master/lua/orgmode/capture/init.lua#L39

lua/orgmode/config/defaults.lua
```diff
@@ -87,6 +94,7 @@ return {
       org_todo_prev = 'ciT',
       org_toggle_checkbox = '<C-Space>',
       org_open_at_point = '<Leader>oo',
+      org_edit_src_block = '<Leader>ose',
```
kristijanhusak, line 97: Lets name this `org_edit_special` since it's called like that in emacs orgmode.

lua/orgmode/config/defaults.lua
```diff
@@ -31,6 +31,13 @@ return {
   org_custom_exports = {},
   org_indent_mode = 'indent',
   org_time_stamp_rounding_minutes = 5,
+  org_edit_src_block = {
```
kristijanhusak, line 34: Lets call this `org-src-window-setup` and set it up to directly be a string (or a function), instead of nesting it as an `open` property.
Default value should be `top 16new`, but it can accept the function as well. Do the `:e tempname()` yourself once this is called, because that needs to happen anyway.

This also removes the need for a function to return anything.

kristijanhusak (changes requested):

lua/orgmode/org/mappings.lua
levouh: Makes sense, I've implemented it in a similar fashion and changed the callback to happen on `BufWipeout` instead of `BufWritePost`. The latter feels a bit more "responsive" but granted it can be a bit overkill (and useless if the user decides to specify something like `tabnew` for editing the temporary buffer).

lua/orgmode/org/mappings.lua
levouh: Updated, not super familiar with _Treesitter_ (and the functions that are available for each node) but I have changed to to function as you've mentioned here.

lua/orgmode/config/defaults.lua
levouh: There might be scenarios where the user might want to handle creating the temporary file (say to have the file in a project such that the language server can read contextual type definitions, etc.), but for now this will likely work just fine. Another option would be to have the `string` option be the "simple" case, and the `function` option rely on the user to do _all_ setup of the buffer (except perhaps the wipeout handling and setting of the file content). Removing the file on disk would also have to be up to the user in this case, but handing around a callback isn't super hard to do.

Either way, I've changed it to be handled as mentioned here for now. It'd be trivial to change should the feature be needed in the future.

lua/orgmode/config/defaults.lua
levouh: Sounds good, I'll be more careful to check the _Emacs_ options in the future.

lua/orgmode/config/defaults.lua
levouh: Do you think `org_edit_special_window_setup` might be better?

lua/orgmode/org/mappings.lua
```diff
@@ -462,6 +462,111 @@ function OrgMappings:show_help()
...
+      [[autocmd BufWipeout <buffer=%s> ++once lua require('orgmode').action('org_mappings._edit_special_callback')]],
+      bufnr
+    )
+  )
+end
+
+function OrgMappings:_edit_special_callback()
+  local success, ctx = pcall(vim.api.nvim_buf_get_var, 0, '__org_edit_special_ctx')
+  if not success or ctx.bufnr ~= vim.api.nvim_get_current_buf() then
+    return { message = '' }
```
levouh, line 529: Need to double check if the return value here is supposed to actually specify this, as I believe it is just a part of the `pcall` result when there is an error.

lua/orgmode/org/mappings.lua
```diff
@@ -462,6 +462,111 @@ function OrgMappings:show_help()
...
+    vim.api.nvim_buf_set_lines(
+      ctx.start_bufnr,
+      content_start,
+      content_end,
+      false,
+      vim.api.nvim_buf_get_lines(0, 0, -1, false)
+    )
+  end
+
+  vim.api.nvim_buf_set_option(ctx.bufnr, 'modified', false)
```
levouh, line 566: Not needed, remove.

lua/orgmode/org/mappings.lua
```diff
@@ -462,6 +462,111 @@ function OrgMappings:show_help()
...
+
+  vim.api.nvim_buf_set_option(bufnr, 'bufhidden', 'wipe')
+  vim.api.nvim_buf_set_option(bufnr, 'filetype', ft)
+  vim.api.nvim_buf_set_lines(bufnr, 0, -1, false, content)
+  vim.api.nvim_buf_set_option(ctx.bufnr, 'modified', false)
+  vim.api.nvim_buf_set_var(bufnr, '__org_edit_special_ctx', ctx)
+
+  vim.cmd(
+    string.format(
+      [[autocmd BufWipeout <buffer=%s> ++once lua require('orgmode').action('org_mappings._edit_special_callback')]],
```
levouh, line 520: Can probably just use `<buffer>`, don't need to be more specific than that.

lua/orgmode/utils.lua
```diff
@@ -370,4 +371,55 @@ function utils.prompt_autocomplete(arg_lead, list, split_chars)
   end, matches)
 end
 
+function utils.get_nearest_src_block(file, cursor)
```
kristijanhusak, line 374: This function is very specific to the src blocks. Would a general function that finds the nearest node do the trick here?
For example, `block` node has `name`, `parameters` and `contents` children. If cursor is on the `block` node, just return it. If it's children, go up the parent until you find the required node. For example `utils.get_parents_until(node, 'block')` would return block node.

lua/orgmode/config/defaults.lua
kristijanhusak: I guess we could have a function do everything, and string be a simple case. Function should be well documented if we take that route.

Regarding the name, I got it from here https://orgmode.org/manual/Editing-Source-Code.html#:~:text=org-src-window-setup .

lua/orgmode/utils.lua
levouh: There's already a function in `utils` that does something similar to this, however it automatically assumes that it should traverse from the parent of the passed node, rather than the passed node itself. I've just modified it to allow traversing from the passed node.

On top of that there's a simple wrapper for getting named child nodes (allows for filtering by name or grabbing all of them), and then a wrapper for getting a block node of a given name. That's about as simplified/abstract as I was able to get it, let me know if that looks good.

lua/orgmode/utils.lua
```diff
@@ -308,8 +309,13 @@ end
 ---@param node table
 ---@param type string
 ---@return table
-function utils.get_closest_parent_of_type(node, type)
-  local parent = node:parent()
+function utils.get_closest_parent_of_type(node, type, accept_at_cursor)
+  local parent = node
+
+  if not accept_at_cursor then
+    node:parent()
```
kristijanhusak, line 316: this needs to be:
```lua
parent = node:parent()
```

lua/orgmode/utils.lua
kristijanhusak: Oh, you're right. I completely missed that. Looks ok.

kristijanhusak:
Everything looks good @levouh , awesome work!
Do you think some tests could be written for this? If it's too tricky, don't bother too much. I assume it's not simple because it opens a separate buffer and all of that.

Besides that, this is good to go!

levouh:
@kristijanhusak I'll look at adding tests, especially since this is modifying buffer content and could end up deleting lines that it shouldn't, etc. if done incorrectly. If it gets too ridiculous I'll let you know, but my plan was to do something similar to how the mappings tests are done and just check the node content, nothing crazy.

levouh:
Added one new check to ensure that the underlying "source" _Org_ buffer has not changed between when the edit special operation started and finished (this leads to some oddities that could end up messing up the _Org_ file content for the user, which should be avoided). If this case is hit (where the underlying source buffer changed so the changes from the special edit buffer are _not_ applied), they are echo-ed to the `:messages` buffer so that the user does not lose their edits, and can restore them manually if needed.

Also added tests, so pending a review of this last set of changes, should be good to go @kristijanhusak.

mjlbach:
> most language servers (or at least things in nvim-lspconfig) expect a file on disk (e.g. such that a root directory can be found)

FYI with single-file-mode in lspconfig, this shouldn't matter as much anymore

kristijanhusak:
I gave this a test. It works with an empty org file that I created for test, but it's not working with my personal org where I have some blocks, getting "Unable to find source block used to initialize edit special buffer" error. I'll try to figure out how to reproduce it tomorrow.

lua/orgmode/org/mappings.lua
```diff
@@ -489,6 +489,129 @@ function OrgMappings:show_help()
...
+  vim.cmd([[autocmd BufWipeout <buffer> ++once lua require('orgmode').action('org_mappings._edit_special_callback')]])
+end
+
+function OrgMappings:_edit_special_callback()
+  local success, ctx = pcall(vim.api.nvim_buf_get_var, 0, '__org_edit_special_ctx')
+  if not success or ctx.bufnr ~= vim.api.nvim_get_current_buf() then
+    error({ message = 'Edit special callback in buffer with no context variable' })
+  end
+
+  if not vim.deep_equal(vim.api.nvim_buf_get_lines(ctx.start_bufnr, 0, -1, false), ctx.lines) then
```
kristijanhusak, line 560: This can be really expensive, especially on big files. You can just save `b:changedtick` before opening the tmp buffer and then compare it here. I'm doing similar thing when parsing the files.

kristijanhusak:
Ok I reproduced it. Seems that indentation on the BEGIN/END Blocks are causing the issue. This content works:

```
* TODO Test
#+BEGIN_SRC lua
  local parser_config = require "nvim-treesitter.parsers".get_parser_configs()
  parser_config.org = {
    install_info = {
      url = '~/github/tree-sitter-org',
      files = {'src/parser.c', 'src/scanner.cc'},
    },
    filetype = 'org',
  }
#+END_SRC
```

But this one doesn't (indented block):
```
* TODO Test
  #+BEGIN_SRC lua
    local parser_config = require "nvim-treesitter.parsers".get_parser_configs()
    parser_config.org = {
      install_info = {
	url = '~/github/tree-sitter-org',
	files = {'src/parser.c', 'src/scanner.cc'},
      },
      filetype = 'org',
    }
  #+END_SRC
```

Seems like it's caused by wrong node being returned when you are on the indented part. I think simplest way to go around this is to go to the last column when trying to find the matching block:

```diff
diff --git a/lua/orgmode/org/mappings.lua b/lua/orgmode/org/mappings.lua
index e48739a..3208c45 100644
--- a/lua/orgmode/org/mappings.lua
+++ b/lua/orgmode/org/mappings.lua
@@ -578,7 +578,7 @@ function OrgMappings:_edit_special_callback()
     file,
     'SRC',
     { 'name', 'contents', 'parameters' },
-    { ctx.src_start_line, 0 },
+    { ctx.src_start_line, 99999 },
     true
   )
   if not src_block then
```

kristijanhusak:
@levouh I would just get the exact indent value of the `BEGIN_SRC` line, and prepend it to all lines in the block, including the `END_SRC`. I believe that would give desired results.
Since you set up emacs for Priority PR, you can also give this a try in emacs to see how they are handling it.

levouh:
Good call, will poke around with both and see where it ends up.

levouh:
Tests are failing due to mentioned leading white space stuff above, will check parity with _Emacs_ and then fix those accordingly. Other issues with column offset, etc. should be fixed now and tests added for them.

lua/orgmode/org/mappings.lua
```diff
@@ -490,6 +490,130 @@ function OrgMappings:show_help()
   return Help.show()
 end
 
+function OrgMappings:edit_special()
+  local start_bufnr = vim.api.nvim_get_current_buf()
+  local changedtick = vim.api.nvim_buf_get_var(start_bufnr, 'changedtick')
+  local start_pos = vim.api.nvim_win_get_cursor(0)
+  local file = Files.get_current_file()
+
+  local make_temp_buf = function()
```
TravonteD, line 499: Could we break this out and pass in the config? It maybe something that could be reused for [capture](https://github.com/nvim-orgmode/orgmode/pull/41)

muniter:
Just an idea. Instead of searching again for the src block once editing in the temp buffer is finished. Why not set an extmark around the block contents; so after editing besearch for the extmark (one has the id) and replace the contents.

This would allow to edit the main file and the temp buffer at the same time. And you can be sure the extmarks are always pointing to the src block you want. So it doesn't mater If the tree changes while editing. 

Can also have some guard like check if the text inside the extmark is the same compared to when the edit was started, to confirm overwritin, etc.

Just an idea :)

lua/orgmode/org/mappings.lua
levouh: Not 100% sure what you mean, can you give an example?

Are you just talking about a generic way to open a buffer, that can be overridden and/or customized by the user?

levouh:
>set an extmark around the block contents

I had zero clue what extmarks were or how they work, but this sounds like an infinitely better idea. I will give it a look and take a crack at implementing them. Any reading or examples you recommend beyond `:h api-extended-marks` @muniter?

levouh:
Indent handling in _Emacs_ (completely vanilla) is a bit all over the place as well:

https://user-images.githubusercontent.com/31262046/144358677-8e96a1ae-85fc-421f-8723-31ee1002cb39.mp4

https://user-images.githubusercontent.com/31262046/144358691-3148a8bf-d907-498d-a980-99ca635dcb8b.mp4

muniter:
Yeah extmarks move with text edits meaning if the initial SRC block locations was line 15 and the user deletes 5 lines at the top, when you go back to insert the edited text, and ask neovim for the extmark position it will give you line 10, dice the mark sticks to the text. 

What you would do is also remember the contents of the SRC block when you started editing it externally so you can compare it with the contents after the external edit is finish. This way ask the user to confirm an overwrite or if the block was not just edited but deleted completely warn the user. 

Extmarks are simple but powerful play around with them and you'll see :)

kristijanhusak:
Seems like they are reindenting the whole block. Don't kill yourself with this, just make it respect the BEGIN_SRC indentation and that should be enough.

levouh:
One interesting note is that _Treesitter_ (or at least this parser?) seems to not report offsets for leading tab characters "correctly", for example:

image

which causes some interesting behavior with regards to indent matching and whatnot (pretty easy to circumvent by just using `vim.api.nvim_buf_get_lines(...)`, just noting it here).

---

`:h extmark` stuff is working (no more `changedtick` checks mostly), just a few more things to fix with regards to leading space fun and it should be good to go.

levouh:
Changes from the last round:

- Add [org-edit-src-content-indentation](http://doc.endlessparentheses.com/Var/org-edit-src-content-indentation.html) to allow indenting of content within the source block itself.
- Add [org-edit-src-abort](https://github.com/bzg/org-mode/blob/5e35de34a6f54c2f76b90aa23b69876aca6587ee/lisp/org-src.el#L701) to allow aborting content and not applying changes from the edit special buffer.
- Fixed issue with edit special blocks being indented.
- Added somewhat naive handling for stripping/re-applying leading indents for the content of the block. The basic strategy here it to match the leading indent of the first line of the block itself, and strip that from the start of each line. See tests for more examples on how this works.
- Refactor to support different "edit special" actions (don't see many being added in the future, but the refactor makes things easier to process as well).
- Support `g?` within the edit special buffer.
- Refactor `help.lua` to facilitate generating requested help content on the fly, backwards compatible.

---

Need to sleep, but will give it a once-over in the morning @kristijanhusak.

lua/orgmode/config/defaults.lua
```diff
@@ -133,6 +136,10 @@ return {
       org_set_effort = '<leader>oxe',
       org_show_help = 'g?',
     },
+    edit_src = {
```
kristijanhusak, line 139: I think there is no need to nest this into `edit_src` object, we could have it on top level.

DOCS.md
```diff
@@ -739,6 +756,18 @@ require('orgmode').setup({
 })
 ```
 
+### Edit Src
```
kristijanhusak, line 759: Since we have a separate "section" for this, lets just add it up to the table of contents so it can be navigated.

kristijanhusak:
Wow, awesome work @levouh! I tested it, it works great! 

One thing that could be additionally improved is to return to the org buffer after the edit src buffer is closed. Currently, it automatically goes to some of the buffers. For me it focused nvim-tree buffer for example.
Other than that, and the minor docs fix, no objections. Works flawlessly.

Thanks for your hard work on this, it is much more complex feature than I anticipated.

lua/orgmode/config/defaults.lua
levouh: It gets a bit confusing as _Vim_ gets confused when using `%` to hop between matching `{`/`}` pairs, but this is just under the `mappings` section. Is that where it should be?

levouh:
More changes this round, beyond the fixes you mentioned above:

- Add [org-edit-src-save](https://github.com/bzg/org-mode/blob/5e35de34a6f54c2f76b90aa23b69876aca6587ee/lisp/org-src.el#L702) so that the content in the source _Org_ buffer can be updated without quitting out of the edit special buffer
- Use `extmark`s to determine the range to update rather than parsing the entire _Treesitter_ tree again, etc. This makes the below feature easier to deal with, as parsing node content would have to happen twice otherwise (once to find the range to update, and again after we apply the changes from the edit special buffer and refresh the file). With `extmark`s being sticky, we can avoid this parsing entirely and just use those points instead. Awesome feature.
- Add highlighting of the source block content in the _Org_ buffer just like _Emacs_ does. Link this highlight group to `Visual` by default, but allow users to override it.
- Re-focus the proper _Org_ buffer when the edit special buffer has exited (I interpreted this to apply to buffers visible in a window in the current tab, as jumping to another tab or opening a new window, etc. would seem a bit jarring, let me know if that aligns with your thoughts @kristijanhusak).

lua/orgmode/config/defaults.lua
kristijanhusak: No, I actually thought to have it on the same level as other org mappings, but then I figured out this is a separate buffer so it needs to be split. Everything is ok.

kristijanhusak:
> Add org-edit-src-save so that the content in the source Org buffer can be updated without quitting out of the edit special buffer

I don't think we need to worry about this. It's a tmp buffer, and regular Vim's save works fine. Unless this is something like saving to a place where it's not discarded on close.

> Use extmarks to determine the range to update rather than parsing the entire Treesitter tree again

GIve it a try, but if the current solution works, don't spend too much time on this

> Add highlighting of the source block content in the Org buffer just like Emacs does

👍🏻 

> Re-focus the proper Org buffer when the edit special buffer has exited

Yeah, only if buffer is opened in any of the windows in the same tab. Don't do anything if you can't find it according to those conditions.

levouh:
All of these changes are already made just _FYI_.

What `org-edit-src-save` does is apply the changes from the temporary buffer back to the source _Org_ buffer explicitly. Right now if you make edits to the temporary buffer and save it, nothing happens. If you then press `,ow` (with changes to the temporary buffer), it will update the source _Org_ buffer explicitly. We could of course wrap this into `BufWritePost` and have it be a bit more "Vim-native" as well rather than providing a separate mapping for it.

Its also completely applicable to remove this entirely and just have the source _Org_ buffer content updated only when the temporary buffer is closed. Up to you.

lua/orgmode/org/mappings.lua
TravonteD: Yes, that's what was intended. Having a generic way to open a buffer could be useful elsewhere.

kristijanhusak:
> All of these changes are already made just FYI.

Oh wow, I haven't even looked at the code, I thought it's just something you plan to do :)

 > We could of course wrap this into BufWritePost and have it be a bit more "Vim-native" as well rather than providing a separate mapping for it.

Lets leave how it is. Emacs also requires doing a mapping (C-x C-s) to apply the changes to the org buffer, so having the mapping for this is perfect.

Regarding this:

> Re-focus the proper Org buffer when the edit special buffer has exited

I think I maybe didn't explain what's exactly the issue. Here's a short gif:

edit-src-focus

Once I finished editing and closed the edit src buffer, my cursor went back to the nvim-tree buffer. I would like it to go back to the org buffer from where I opened the src buffer.

Also, while testing, I noticed that if you edit a src block that's at the end of the file, content is duplicated. This is most likely due to tree-sitter parser not handling end of file lines properly. I already had similar issues somewhere, I think with text objects.
edit-src

I don't think you need to worry about this too much at this moment, since parser is being rewritten and hopefully these issues will be gone.

levouh:
Gotcha, the change I made to re-focus the right _Org_ source buffer (see [here](https://github.com/nvim-orgmode/orgmode/blob/a5acb8adb93c60f6cec1631afabb14d9bc044e49/lua/orgmode/objects/edit_special/init.lua#L152)) seems to not be working. Ideally it would work just like you noted. It is in a `BufWipeout` callback, so me changing the focused window is actually just a bad idea and likely this needs to be wrapped in a `vim.schedule(...)` call to delay it until after the `autocmd` scope is complete. This _should_ also be easy-ish to add tests for, so I will look at doing that and ensure it works.

I'll take a look to see if there is an easy workaround for the source blocks being at the end of the buffer, otherwise leave it be if not.

levouh:
OK yea, it was just a simple `autocmd` scope thing. Delaying the setting of the window by wrapping it in a `vim.schedule(...)` call fixes it @kristijanhusak. This is pushed up (sorry I have a bad habit of always rebasing and force pushing :sweat_smile:). Examples of before:

https://user-images.githubusercontent.com/31262046/145095713-c734bd49-0a20-49ad-9ce6-712a8b74c4f2.mp4

and after:

https://user-images.githubusercontent.com/31262046/145095742-fbf7c4b0-d172-44b4-a5b4-dd8c0ee431d9.mp4

---

I will take a look at the other issue you mentioned and see if there isn't an easy way around it.

kristijanhusak:
Yeah, that fixed it, thanks.

As far as I'm concerned, this is ready to go. Once you confirm the last thing can be fixed or not, let me know.

levouh:
I pushed up one more fix for removing the highlighting on the source content in the _Org_ buffer when the edit special buffer is aborted.

No luck with getting the "end of buffer" _Treesitter_ stuff to work correctly, so should I note that is a known bug in the docs or just roll with it?

Given feedback on the above, should be good to close this one (finally) @kristijanhusak.

kristijanhusak:
@levouh Thanks!
I don't think we need to worry about that issue at this point, I believe it's an edge case. Especially since parser is being rewritten, so I hope it will solve some of these strange issues.

joshpetit:
Hey @levouh! I've been trying to modify edit special src to support narrowing and widening subtrees (#200), but I don't really understand the implementation at all 😅. I'm new to lua plugin development so would appreciate any help if possible haha. You can reach out to me through my email on my profile if that's cool with you.

levouh:
Discord is easier if that also works for you @joshpetit, `Levouh#0003`.

PR #152, opened 2021-11-15
Ability to set & remove the priority on headline items

levouh:
Based off functionality in _Orgmode_ found [here](https://orgmode.org/manual/Priorities.html). A few things to do still:

- [x] Better integration with priority with a `Section` object
- [x] Switch todo state change actions to use `utils.choose`
- [x] Ability to cycle priorities
- [x] Better mappings, as anything with `cip` conflicts with builtin _Vim_ action (would be used similar to next/previous state for current todo mappings, but `p/P` instead of `t/T`)
- [x] Add tests


lua/orgmode/org/mappings.lua
```diff
@@ -258,6 +258,58 @@ function OrgMappings:change_date()
...
+
+  local start = high_priority
+  local fin = low_priority
+  if is_alphabetical then
+    start = string.byte(high_priority)
+    fin = string.byte(low_priority)
+  end
+
+  local enumerated = {}
+  for i = start, fin do
```
levouh, line 276: Priority count could be high, probably good to bound this somewhere.

lua/orgmode/utils.lua
```diff
@@ -370,4 +370,32 @@ function utils.prompt_autocomplete(arg_lead, list, split_chars)
...
 end
 
+function utils.choose(items)
+  items = items or {}
+
+  local output = {}
+  for _, item in ipairs(items) do
+    table.insert(output, { string.format('[%s] ', item.choice_text) })
+
+    if item.desc_text and item.desc_hl then
```
kristijanhusak, line 380: I don't think check for `desc_hl` is needed because it defauts to Normal inside the block.

lua/orgmode/org/mappings.lua
```diff
@@ -258,6 +258,58 @@ function OrgMappings:change_date()
...
+  end
+
+  table.insert(enumerated, { choice_value = ' ', choice_text = '<Space>', desc_text = 'Remove', desc_hl = 'Comment' })
+
+  local choice = utils.choose(enumerated)
+  local linenr = item.range.start_line
+  local stars = string.rep('%*', item.level)
+
+  local static_state = todo.value
+  if static_state ~= '' then
```
kristijanhusak, line 292: I think this if is not necessary, value points to itself.

lua/orgmode/org/mappings.lua
```diff
@@ -258,6 +258,58 @@ function OrgMappings:change_date()
...
+      key = string.char(i)
+    end
+
+    table.insert(enumerated, { choice_value = key, choice_text = key, desc_text = '', desc_hl = 'Comment' })
+  end
+
+  table.insert(enumerated, { choice_value = ' ', choice_text = '<Space>', desc_text = 'Remove', desc_hl = 'Comment' })
+
+  local choice = utils.choose(enumerated)
+  local linenr = item.range.start_line
```
kristijanhusak, line 288: I would like to move all this logic of setting priority to the `Section` file as `set_priority` method, that just accepts the priority valueas an argument.

lua/orgmode/org/mappings.lua
kristijanhusak: Actually emacs allows 3 level of priorities, i don't think it can go beyond that. This looks fine to me.

kristijanhusak (changes requested):

lua/orgmode/org/mappings.lua
levouh: Pulling it from the quote:

>When using numeric priorities, you need to set org-priority-highest, org-priority-lowest and org-priority-default to integers, which must all be strictly inferior to 65

mentioned [here](https://orgmode.org/manual/Priorities.html).

lua/orgmode/utils.lua
levouh: Good call.

levouh:
Will need to fix/add tests, but want to get your thoughts before spending time testing @kristijanhusak. Let me know what you think when you have time, appreciate the review.

kristijanhusak:
Everything looks ok, you can proceed with tests.

levouh:
Should be good to go @kristijanhusak (not sure if I have to mention you every time, so apologies if it gets annoying :wink:).

I added a handful of tests and moved the priority calculation portion into a separate file/object to make it easier to test similar to what was done with the todo state stuff. Using `nvim_call_function` is no fun you gotta do what you gotta do :smile:.

lua/orgmode/objects/priority_state.lua
```diff
@@ -0,0 +1,98 @@
...
+end
+
+---@return string
+function PriorityState:increase()
+  if self.current_priority == self.high_priority then
+    print('increase return early')
+    return self.current_priority
+  end
+
+  return self:_apply(self:is_alphabetical() and -1 or 1)
```
kristijanhusak, line 42: I just tested Emacs orgmode and it actually treats number priority the same way as char priority. Which means that `1` is highest priority, and `3` is the lowest. This means that you can always do `string.byte` and `string.char`, and it should give you the desired result.

Also, I figured out that my current implementation of getting the priority number for sorting is not correct. I always assumed that it's possible only to set one of those 3 config options, but it turns out you can also go in between, especially with numbers.

Could you update that method on section so it properly works with this change?

kristijanhusak:
Awesome work!

I wanted to suggest to move it to a separate "priority state" but didn't want to bother you too much.

I left you a comment regarding the number priorities, change should actually simplify some things.

levouh:
>I wanted to suggest to move it to a separate "priority state" but didn't want to bother you too much.

I'm always in for learning how to do things better as most of what I write is for my own config. and isn't targeted to be consumable or tested, etc. Always let me know if you think there are ways I could do things better :tada:

lua/orgmode/objects/priority_state.lua
levouh: I assume you are talking about [this](https://github.com/kristijanhusak/orgmode.nvim/blob/012a7eb5e292311d26301b882193add0bf4c058b/lua/orgmode/parser/section.lua#L193) section?

lua/orgmode/objects/priority_state.lua
```diff
@@ -0,0 +1,98 @@
...
+end
+
+function PriorityState:is_alphabetical()
+  return string.find(self.high_priority, '%a')
+end
+
+---@return string
+function PriorityState:increase()
+  if self.current_priority == self.high_priority then
+    print('increase return early')
```
levouh, line 38: Remove.

lua/orgmode/objects/priority_state.lua
levouh: Should be fixed, fixing tests then will push up.

kristijanhusak: Yeah that section.

levouh: Should be fixed, let me know how that looks.

lua/orgmode/objects/priority_state.lua
```diff
@@ -0,0 +1,116 @@
...
+end
+
+function PriorityState:is_alphabetical()
+  return string.find(self.high_priority, '%a')
+end
+
+---@return number
+function PriorityState:get_sort_value()
+  local priority = self.priority
+  if self:is_alphabetical() then
```
kristijanhusak, line 42: You can actually remove the check for `alphabetical` and always convert to byte and back. `string.byte(1)` is for example 49, 2 is 50, and so on. Theoretically someone can mix both numbers and letters if they want, if it's according to this byte value.

kristijanhusak: There is mention of ascii value here https://emacs.stackexchange.com/a/2741, which is taken from the orgmode source.

levouh: Not sure this will work if users use multi-byte values:

```lua
print(string.char('a'))  -- 97
print(string.byte('aa'))  -- 97
print(string.byte('a7'))  -- 97
```

kristijanhusak: I can't really figure out how emacs is handling these. From the same link I attached above, you can see that someone uses this:
```elisp
(setq org-highest-priority ?U)
(setq org-default-priority ?i)
(setq org-lowest-priority ?o)
```
Since I wasn't sure what `?` means, I googled it, and it turns out that `?` [converts the value to ascii](https://stackoverflow.com/a/19867243/1474465), so they are doing something with ascii here.

Generally it seems that emacs orgmode itself is not handling these too well, according to few of these SO questions.

I think your solution of converting both things to string and negating it will do the trick. This will reduce complexity by removing the check if it's alphabetical, and should work in most of cases.

levouh: Did a bit more poking at this, and as mentioned [here](https://emacs.stackexchange.com/questions/29935/customize-org-priorities-with-strings):

>Emacs converts the priority letters or numbers into character numbers for purposes of sorting, etc. -- e.g., (string-to-char "A") is 65.

which makes sense why it is noted [here](https://orgmode.org/manual/Priorities.html) that all numeric priorities:

>must all be strictly inferior to 65

---

I took some time to get _Emacs_ setup to compare feature parity, and the way we are choosing priorities here is actually a bit off as well. There are two things that need to be changed:

1. The prompt should be more just like a vanilla `vim.ui.input` (or just `input`) that allows the user to type in the priority and hit `<CR>`.
2. When cycling priorities, the priority should be removed if the user goes above/below the highest priority

Going to change these two things, make sure the sorting function coincides with this and then go from there.

kristijanhusak:
@levouh I just gave this a test, and it's working really great!
I don't have anything else to add here. Do you want to do something more or it's good to go?

kristijanhusak (approved):

levouh:
Working on a quick fix right now where the priority can't be set on an item unless it has a todo state because the replace pattern matches `%s+%s+` (because the `TODO` keyword or similar that is in between is empty). After that, should be good to go.

levouh:
Good to go @kristijanhusak.

kristijanhusak:
Awesome work, thanks!

PR #158, opened 2021-11-19
fix: allow multiple key mappings to global

grepinsight:
Hello! thank you for this amazing plugin. 
I've encountered one issue while using thils.

The manual says it supports multiple key mappings, so I tried this
```lua
        global = {
            org_agenda = {'gA', '<C-c>a'},
            org_capture = 'gC'
        },
```
However, this gives following error message:

Screen Shot 2021-11-19 at 12 46 31 PM

This PR is an attempt to fix this. Thanks!

### DIFF
--- a/lua/orgmode/config/init.lua
+++ b/lua/orgmode/config/init.lua
@@ -175,12 +175,14 @@ function Config:setup_mappings(category)
     return
   end
   if not category then
-    if type(self.opts.mappings.global.org_agenda) == "table" then
-        for _, k in ipairs(self.opts.mappings.global.org_agenda) do
-        utils.keymap('n', k, '<cmd>lua require("orgmode").action("agenda.prompt")<CR>')
-        end
+    if type(self.opts.mappings.global.org_agenda) == "string" then
+        keys = { self.opts.mappings.global.org_agenda }
     else
-        utils.keymap('n', self.opts.mappings.global.org_agenda, '<cmd>lua require("orgmode").action("agenda.prompt")<CR>')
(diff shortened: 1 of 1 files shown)


lua/orgmode/config/init.lua
```diff
@@ -175,8 +175,28 @@ function Config:setup_mappings(category)
     return
   end
   if not category then
-    utils.keymap('n', self.opts.mappings.global.org_agenda, '<cmd>lua require("orgmode").action("agenda.prompt")<CR>')
-    utils.keymap('n', self.opts.mappings.global.org_capture, '<cmd>lua require("orgmode").action("capture.prompt")<CR>')
+    agenda_key = self.opts.mappings.global.org_agenda
```
kristijanhusak, line 178: This can be shortened in both agenda and capture sections a bit like this:
```lua
local agenda_keys = self.opts.mappings.global.org_agenda
if type(agenda_key) == 'string' then
  agenda_keys = { agenda_keys }
end

for _, k in ipairs(agenda_keys) do
  utils.keymap('n', k, '<cmd>lua require("orgmode").action("agenda.prompt")<CR>')
end
```

kristijanhusak (changes requested):
Good catch!

It is supported for all except global mappings, I missed that part :)

Just address this one comment I left and it's good to go. Thanks!

lua/orgmode/config/init.lua
grepinsight: Thank you for your comment! Addressed in 84d1da6

kristijanhusak:
Thanks!

PR #165, opened 2021-11-25
fix org_hide_leading_stars on folded lines

rahulsalvi:
I'm not sure what changed, but hiding leading stars on folded lines stopped working for me. This commit escapes the regex string more explicitly and seems to fix the issue.


kristijanhusak:
Thanks, this fixes it. Just run it through stylua (`make format`) so it gets formatted properly.

rahulsalvi:
Formatted as you requested. Thanks

kristijanhusak:
Thanks!

PR #166, opened 2021-11-30
feature: add support for cookies

### LINKED ISSUE #81: Auto-updating checkbox cookies

Yes

### Orgmode link

https://orgmode.org/manual/Checkboxes.html

### Feature value

_No response_

### Additional context

Would it be possible to have auto-updating checkbox cookies such that checking off an item in a list updates the progress? More specifically, say I have this task tree:

```
* My project [/]
- [ ] Task 1
- [ ] Task 2
- [ ] Task 3
```

Checking off `Task 1` by keybind will update the cookie

```
* My project [1/3]   <- cookie updated to number of items in tree
- [x] Task 1
- [ ] Task 2
- [ ] Task 3
```

And checking off `Task 2` will automatically increment the cookie

```
* My project [2/3]   <- number of item completed updated
- [x] Task 1
- [x] Task 2
- [ ] Task 3
```

lukas-reineke:
Adds support for cookies in headlines and lists

https://orgmode.org/manual/Checkboxes.html

Add `[/]` or `[%]` to a headline or list with sub-items, and it should display the cookie.
It needs this to work https://github.com/milisims/tree-sitter-org/pull/11

fix #81

---

orgmode also supports this for TODO states if the cookie is in the headline. This PR does not cover that.

### DIFF
--- a/lua/orgmode/org/syntax.lua
+++ b/lua/orgmode/org/syntax.lua
@@ -1,5 +1,10 @@
 local Files = require('orgmode.parser.files')
 local config = require('orgmode.config')
+local utils = require('orgmode.utils')
+local ts_utils = require('nvim-treesitter.ts_utils')
+
+local checkbox_query = vim.treesitter.parse_query('org', '(list (listitem (checkbox) @checkbox))')
+local headline_cookie_query = vim.treesitter.parse_query('org', '(headline (cookie) @cookie)')
 
 local function load_code_blocks()
   local file = vim.api.nvim_buf_get_name(0)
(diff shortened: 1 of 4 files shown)


lua/orgmode/org/syntax.lua
```diff
@@ -36,7 +42,73 @@ local function add_todo_keywords_to_spellgood()
...
+  })
+end
+
+local function _get_cookie_checked_and_total(parent)
+  local parent_type = parent:type()
+  local start_row, _, end_row, _ = parent:range()
+  local checked, total = 0, 0
+  for _, checkbox in checkbox_query:iter_captures(parent, 0, start_row, end_row + 1) do
+    local closest_parent = utils.get_closest_parent_of_type(checkbox, parent_type)
+    if closest_parent and closest_parent == parent then -- only count direct children
```
lukas-reineke, line 73: it would be nicer if the query could do this directly, but I don't know if that is possible.

kristijanhusak:
> orgmode also supports this for TODO states if the cookie is in the headline. This PR does not cover that.

Can you give an example of this? I don't understand it.

ourigen:
@kristijanhusak I believe he means that for example if we have

```
* TODO Some project [2/3]
- [x] Task 1
- [x] Task 2
- [ ] Task 3
```
and we check the last box, then `TODO` will be converted to `DONE` automatically
```
* DONE Some project [3/3]
CLOSED: [2021-11-30 17:00]
- [x] Task 1
- [x] Task 2
- [x] Task 3
```

lukas-reineke:
> and we check the last box, then TODO will be converted to DONE automatically

No, this is not what I mean.
`TODO` and `DONE` headlines can also be the source for cookies.
Like this:

```org
* TODO bake a cake [3/6]
** DONE buy ingredients
** DONE start cooking
** DONE realize you forgot eggs, dammit
** TODO drive back to the store and buy eggs
** TODO wait, I needed THREE sticks of butter?
** TODO drive back to the store and just buy a damn cake
```

This is from [the documentation](https://orgmode.org/manual/Checkboxes.html)

> In a headline, a cookie can count either checkboxes below the heading or TODO states of children, and it displays whatever was changed last. Set the property ‘COOKIE_DATA’ to either ‘checkbox’ or ‘todo’ to resolve this issue.

lua/orgmode/org/syntax.lua
```diff
@@ -36,7 +42,77 @@ local function add_todo_keywords_to_spellgood()
...
+  if cookie_text == '[/]' then
+    text = string.format('[%d/%d]', checked, total)
+  else
+    if total > 0 then
+      text = string.format('[%d%%]', (100 * checked) / total)
+    else
+      text = '[0%]'
+    end
+  end
+  vim.api.nvim_buf_set_extmark(0, cookie_namespace, row, column + (leading_spaces or 0), {
```
kristijanhusak, line 63: Emacs orgmode is actually modifying the content for this instead of using something like virtual text. That's why they require `[/]` to be added to the end. I'd love to propose 2 solutions to this:

1. Do the same thing that emacs does (Really replace the text)
2. Implement this as a virtual text without a need for a user to type `[/]` 

Current solution is kind of a hybrid, and I think having empty `[/]` as a content will cause confusion.
I'd love to go with 1st solution, since it copies the emacs orgmode functionality which I'm trying to achieve most of the time.
I know that 2nd solution is simpler, and we could have it enabled by default, and to allow it to be disabled through config.
If first one is too complicated, we can go with 2nd one.

lukas-reineke: :+1: I didn't know emacs modifies the file.
shouldn't be that hard to do.

lukas-reineke: Updating the text in the buffer is very expensive.. it gets very  slow even with only a couple cookies.
Maybe we use virtual text and only update the actual text on save?

kristijanhusak: @lukas-reineke how are you updating it? It should update only when you check/uncheck something, maybe you are triggering it more often. Maybe if you have file with several thousand lines, but even than shouldn't cause that many issues.

lukas-reineke: I was updating all of them before, but yeah without virt text it makes much more sense to only trigger this on checkbox toggle.
I rewrote the logic to do this.
Added bonus, checkboxes update their parents now as well.

lua/orgmode/org/syntax.lua
```diff
@@ -36,7 +41,115 @@ local function add_todo_keywords_to_spellgood()
   end
 end
 
+local function _get_cookie_checked_and_total(parent)
```
kristijanhusak, line 44: Lets move this thing to a separate file, it's fairly big at this point.

lua/orgmode/org/syntax.lua
lukas-reineke: it doesn't need recursion. It should only check the first level of checkboxes. The treesitter query finds all nested ones as well, so I had to do this hacky check.

The implementation with just `iter_children` isn't super pretty either, but you are probably right, it's easier to understand and faster

```lua
local function _get_cookie_checked_and_total(parent)
  local checked, total = 0, 0
  for child in parent:iter_children() do
    if child:type() == 'listitem' then
      for listitem_child in child:iter_children() do
        if listitem_child:type() == 'checkbox' then
          local checkbox_text = vim.treesitter.get_node_text(listitem_child, 0)
          if checkbox_text:match('%[[x|X]%]') then
            checked = checked + 1
          end
          total = total + 1
        end
      end
    end
  end
  return checked, total
end
```

kristijanhusak:
This looks ok to me for now. I'll wait for the upstream tree-sitter-org PR to get merged to test this. I'm just not sure why tests started to fail.

lukas-reineke:
The treesitter queries will need to be updated anyway after the change in tree-sitter-org.

I'll take a look at the tests then.

Do you want to keep support for 0.5 now that 0.6 is released? I think I used some 0.6 APIs.
Other plugins like `nvim-lspconfig` have already dropped support for 0.5, and it would make it easier. But I can try to make it work in 0.5 as well.

kristijanhusak:
Do you know what exactly did you use from 0.6? I would love to keep 0.5 support, but if it complicates things drastically, we could have a release before merging it.

lukas-reineke:
Oh maybe I didn't. I thought `vim.treesitter.get_node_text` and `tsnode:named_child` were 0.6, but I just checked and they are 0.5.
Maybe it's fine then. I'll write tests for this anyway, then we'll know for sure.

YuCao16:
Is there any progress on this feature? Other than that, I think it might be useful if the concept of weights could be introduced. 

Thanks for you work, enjoy this Plug so much!

lukas-reineke:
I am waiting on this https://github.com/milisims/tree-sitter-org/issues/13

Once that is done I can update this PR.

Is weights a native org feature?

YuCao16:
> I am waiting on this https://github.com/milisims/tree-sitter-org/issues/13
> 
> Once that is done I can update this PR.
> 
> Is weights a native org feature?

No, as far as I know it isn't.  But weights can be very useful if you can prioritize checkboxes.  But this is just an idea :D

lukas-reineke:
I agree it could be useful, but if it is not in org, it will most likely also not make it in this repo.

YuCao16:
> I agree it could be useful, but if it is not in org, it will most likely also not make it in this repo.

I totally understand and look forward to the cookie going merge soon, thank you for your work.

PhosCity:
> I am waiting on this [milisims/tree-sitter-org#13](https://github.com/milisims/tree-sitter-org/issues/13)
> 
> Once that is done I can update this PR.


[milisims/tree-sitter-org#13](https://github.com/milisims/tree-sitter-org/issues/13) has been closed. Will there be any update to this? It would be wonderful to have this feature in the plugin.

lukas-reineke:
I will finish this. Just have to see when I can find some time.

gerazov:
This is awesome :sunglasses: 

I saw it's mentioned but atm will it also work with TODO/DONE cookies as well?

kristijanhusak:
Closing in favor of https://github.com/nvim-orgmode/orgmode/pull/287

PR #191, opened 2022-01-04
Conceal links in fold text

levouh:


lua/orgmode/org/indent.lua
```diff
@@ -92,10 +92,17 @@ end
...
   local line = vim.fn.getline(vim.v.foldstart)
+
   if config.org_hide_leading_stars then
-    return vim.fn.substitute(line, '\\(^\\*\\+\\)', '\\=repeat(" ", len(submatch(0))-1) . "*"', '')
-      .. config.org_ellipsis
+    line = vim.fn.substitute(line, '\\(^\\*\\+\\)', '\\=repeat(" ", len(submatch(0))-1) . "*"', '')
+  end
+
+  if vim.opt.conceallevel:get() > 0 then
+    line = string.gsub(line, "%[%[.-%]%[(.-)%]%]", function(captures)
```
jgollenz, line 101: This does not seem to work with folded `[[links]]` that have no name. I hacked it with 
```
line = string.gsub(line, "%[%[(.-)%]%[?(.-)%]?%]", function(link, name)
    if (name == "") then
        return link
    else
        return name
    end  
end)
```

levouh:
Work as expected @jgollenz?

jgollenz:
> Work as expected @jgollenz?

Almost. Having issues for a headline like this: 
`* [[link]]]`

The third closing bracket is concealed when the headline is folded, but being shown when unfolded. Not that important of a bug though

kristijanhusak:
@levouh looks good to me. Do you think the bug that @jgollenz found needs to be addressed?

levouh:
18 days! I completely forgot this one existed, thanks for the patience :)

Is this the ideal behavior then @jgollenz:

https://user-images.githubusercontent.com/31262046/151002337-92e506fe-76d6-491f-bf06-aa843356410c.mp4

?

kristijanhusak:
That seems perfect, since it behaves as regular conceal.

levouh:
Updated.

jgollenz:
> is this the ideal behavior then

It looks good to me. Can't really remember what the issue was about the second `?` in the pattern. Just give me a couple more hours and I will check it locally

kristijanhusak:
@levouh this works ok for links that have "alias", but it doesn't hide brackets for links without it. Also, regex matches only links at the end of the line. Here are few gifs:

Issue with links without alias:
no-alias-link

Issue with end of line:
end-of-line-issue

I'm not sure how much this complicates things, but I would like to keep this as fast as possible, because this can slow things down when people are using folds.
One minor optimization could be an additional check for `[[` via `line:find('%[%[')` to avoid doing the replacement for each line. Not sure if it will help much, but we can give it a try.

levouh:
Definitely. The `$` is dumb in my part, it should really just check for "the end of the token", a subset of which is "the end of a line". I will fix that and look at the concealing of links without an alias.

Appreciate the thorough testing on your guys' part.

jgollenz:
@levouh both issues mentioned seem to go away when I make the closing bracket of the alias optional in addition to removing the `$`:

``` lua
line = string.gsub(line, '%[%[(.-)%]%[?(.-)%]?%]', function(link, text)
                                             ^
                                            here
```
alias

levouh:
Added that edit @jgollenz as well as the "candidacy check" @kristijanhusak. Will mess around with the use-cases that have been brought up here to make sure it handles all of them.

kristijanhusak:
@levouh I gave it a test, all of the above issues are fixed.
Only thing that @jgollenz initially reported which I can't figure out why it happens is that it conceals an extra `]` when there is no alias. Regex probably catches it and conceals it.

Unfolded:
unfolded

Folded:
folded

I don't think this is a blocker, we can go without it.

kristijanhusak:
Thanks!

jgollenz:
@levouh whoah, how did you make me the actual author of ca963a0dba18f0520ce43b81697e542db224e2c8? :open_mouth: 
First of all, thank you! But more importantly, please teach me :pray:

levouh:
Less magic than you think:

`git commit --amend --no-edit --author "..."`.

You have to put the author credentials in the right format though. I have an `fzf` completion for `git` authors so I just search that way, but you can also (usually) see it in the `git log` given the right set of flags.

kristijanhusak:
TIL

PR #194, opened 2022-01-04
Check if vim.filetype is defined before calling functions

### LINKED ISSUE #193: Error in filetype when starting nvim
### Describe the bug

When pulling the latest master, I get this error when starting. Everything worked before this. I believe this is caused by 022b519a20d802c445ed80cae64813085425cfba.

### Steps to reproduce

Just start nvim, I get:
```
Error detected while processing /Users/wys/.local/share/nvim/site/pack/packer/start/orgmode/ftdetect/filetype.lua:                                                                                                                         
E5113: Error while calling lua chunk: ...vim/site/pack/packer/start/orgmode/ftdetect/filetype.lua:1: attempt to index field 'filetype' (a nil value)                                                                                       
stack traceback:                                                                                                                                                                                                                           
        ...vim/site/pack/packer/start/orgmode/ftdetect/filetype.lua:1: in main chunk  
```

### Expected behavior

It should work without errors.

### Emacs functionality

NA

### Minimal init.lua

The minimal causes this issue, though you need to fix the bug on line 29 for it to work.

_No response_

### OS / Distro

Macos

### Neovim version/commit

0.6.0

### Additional context

_No response_

levouh:
Fixes #193

### DIFF
--- a/ftdetect/filetype.lua
+++ b/ftdetect/filetype.lua
@@ -1,6 +1,8 @@
-vim.filetype.add({
-  extension = {
-    org = "org",
-    org_archive = "org",
-  },
-})
+if vim.filetype then
+  vim.filetype.add({
+    extension = {
+      org = 'org',
+      org_archive = 'org',
+    },
+  })
+end


kristijanhusak:
I thought the check is not needed. Shouldn't filetype.lua be sourced only in latest master?

levouh:
Sourcing of these files (as far as I understand it, by a specific name like `init.lua`) comes from `$USER/filetype.lua` or `$VIMRUNTIME/*.lua` (iffy on the ladder as I'm not sure about special names, the post below notes, just like normal `ftdetect` files, it could be something like `org.lua` and it would work the same).

Based on the comment [here](https://www.reddit.com/r/neovim/comments/rvwsl3/introducing_filetypelua_and_a_call_for_help/) stating:

>`filetype.lua` is currently opt-in, so it will not be enabled unless you explicitly enable it yourself. You can do so by adding the following to your `init.vim`/`init.lua` (or you can create a new `filetype.lua` file and add it there: the result is the same):

>`let g:do_filetype_lua = 1`

with an emphasis on:

>or you can create a new `filetype.lua` file and add it there: the result is the same

would lead me to assume that merely creating files like this enables it.

kristijanhusak:
Thanks for the detailed explanation. Anyway, it doesn't hurt to have the check.

levouh:
I updated it (several times) cause my understanding changed, the basic problem is just that this is on our `&rtp`.

Either way, checking doesn't hurt like you said.

PR #198, opened 2022-01-06
Toggle headlines

### LINKED ISSUE #164: Implement `org-toggle-heading` functionality

Yes

### Orgmode link

https://orgmode.org/manual/Structure-Editing.html

### Feature value

Ability to quickly turn arbitrary text into org headings will improve UX

### Additional context

`org-toggle-item` which is  `C-c -` in emacs would be also nice to have. This converts text into a list.

jgollenz:
fixes #164

todo:

- [x] toggle top level headlines (bug)
- [x] toggle plain list items
- [x] toggle first line of paragraph (bug)
- [x] add tests
- [x] toggle checkboxes to TODO headlines/DOCS.md#org_indent_mode)
- [x] add docs

### DIFF
--- a/lua/orgmode/config/defaults.lua
+++ b/lua/orgmode/config/defaults.lua
@@ -102,6 +102,7 @@ return {
       org_todo = 'cit',
       org_todo_prev = 'ciT',
       org_toggle_checkbox = '<C-Space>',
+      org_toggle_heading = '<Leader>o*',
       org_open_at_point = '<Leader>oo',
       org_edit_special = [[<Leader>o']],
       org_cycle = '<TAB>',
--- a/lua/orgmode/org/mappings.lua
+++ b/lua/orgmode/org/mappings.lua
@@ -296,6 +296,36 @@ function OrgMappings:todo_prev_state()
   self:_todo_change_state('prev')
 end
 
+function OrgMappings:toggle_heading()
+  local line = vim.fn.getline('.')
+  local parent = Files.get_closest_headline()
+  if not parent then
+    line = '* ' .. line
+    vim.fn.setline('.', line)
+    return
+  end
+
+  if parent.line_number == vim.api.nvim_win_get_cursor(0)[1] then
+    line = line:gsub('^%*+%s', '')
+  else
+    line = line:gsub('^(%s*)', '')
+    if line:match('^[%*-]%s') then -- handle lists
+      line = line:gsub('^[%*-]%s', '') -- strip bullet
+      line = line:gsub('^%[([X%s])%]%s', function(checkbox_state)
+        if checkbox_state == 'X' then
+          return config:get_todo_keywords().DONE[1] .. ' '
+        else
+          return config:get_todo_keywords().TODO[1] .. ' '
+        end
+      end)
+    end
+
+    line = string.rep('*', parent.level + 1) .. ' ' .. line
+  end
+
+  vim.fn.setline('.', line)
+end
+
 function OrgMappings:_todo_change_state(direction)
   local item = Files.get_closest_headline()
   local was_done = item:is_done()
--- a/DOCS.md
+++ b/DOCS.md
@@ -637,6 +637,9 @@ Cycle todo keyword backward on current headline.
 #### **org_toggle_checkbox**
 *mapped to*: `<C-Space>`<br />
 Toggle current line checkbox state
+#### **org_toggle_heading**
+*mapped to*: `<Leader>o*`<br />
+Toggle current line to headline and vice versa. Checkboxes will turn into TODO headlines.
 #### **org_open_at_point**
 *mapped to*: `<Leader>oo`<br />
 Open hyperlink or date under cursor.<br />
--- a/README.md
+++ b/README.md
@@ -295,6 +295,7 @@ set concealcursor=nc
   * Change headline TODO state: forward<kbd>cit</kbd> or backward<kbd>ciT</kbd>
   * Open hyperlink or date under cursor: <kbd>\<Leader\>oo</kbd>
   * Toggle checkbox: <kbd>\<C-space\></kbd>
+  * Toggle current line to headline and vice versa: <kbd>\<Leader\>o*</kbd>
   * Toggle folding of current headline: <kbd>\<TAB\></kbd>
   * Toggle folding in whole file: <kbd>\<S-TAB\></kbd>
   * Archive headline: <kbd>\<Leader\>o$</kbd>
--- a/lua/orgmode/config/mappings.lua
+++ b/lua/orgmode/config/mappings.lua
@@ -48,6 +48,7 @@ return {
     org_priority_down = { 'org_mappings.priority_down' },
     org_todo_prev = { 'org_mappings.todo_prev_state' },
     org_toggle_checkbox = { 'org_mappings.toggle_checkbox' },
+    org_toggle_heading = { 'org_mappings.toggle_heading' },
     org_open_at_point = { 'org_mappings.open_at_point' },
     org_edit_special = { 'org_mappings.edit_special' },
     org_cycle = { 'org_mappings.cycle' },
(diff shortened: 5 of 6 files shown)


lua/orgmode/org/mappings.lua
```diff
@@ -295,6 +295,20 @@ function OrgMappings:todo_prev_state()
...
+function OrgMappings:toggle_heading()
+  local node = Files.get_node_at_cursor()
+  local line = vim.fn.getline('.')
+  local parent_level = Files.get_closest_headline().level
+
+  if (node:type() == 'headline' or node:parent():type() == 'headline') then
+    line = line:gsub('^(%*+)', string.rep(' ', parent_level - 1))
+  elseif (node:type() == 'paragraph') then
+    line = line:gsub('^(%s+)', string.rep('*', parent_level + 1) .. ' ')
+  end
```
levouh, line 307: Would it suffice to just check if the `node:type() == 'headline'`?

Some testing in _Emacs_ makes it seem like the only time the headline is toggled _off_ is when you are directly on a headline, otherwise it just adds the leading stars as `closest_headline_star_count + 1`.

lua/orgmode/org/mappings.lua
```diff
@@ -295,6 +295,20 @@ function OrgMappings:todo_prev_state()
   self:_todo_change_state('prev')
 end
 
+function OrgMappings:toggle_heading()
+  local node = Files.get_node_at_cursor()
```
levouh, line 299: _nit_: Since we have the nearest section, we could compare line numbers rather than parsing nodes. Not entirely sure it makes any difference, but is perhaps more explicit.

```lua
if parent.line_number == vim.api.nvim_win_get_cursor(0)[1] then
  ...
```

lua/orgmode/config/defaults.lua
```diff
@@ -102,6 +102,7 @@ return {
       org_todo = 'cit',
       org_todo_prev = 'ciT',
       org_toggle_checkbox = '<C-Space>',
+	  org_toggle_heading = '<Leader>oh',
```
levouh, line 105: _nit_: Perhaps `<Leader>o*` to match what _Emacs_ has for this binding. However we've strayed pretty far at this point so not sure it matters.

levouh (commented):
Don't forget to run `make format` as well.

lua/orgmode/org/mappings.lua
jgollenz: > what is being done here

My goal is to enable toggling a headline off from anywhere on the actual line. I looked at [this example](https://github.com/milisims/tree-sitter-org#example) and the `node:type()` might not be `headline` but can also be `stars`, `item` or `tags`. Anything that has `headline` as parent. Not evident from the code, but I thought documentation is not really intended in the code in this repo. Correct me if I'm wrong ;)

> when you are directly on a headline

I assume you mean "anywhere on the line of the headline" by that. At least I can do that in my Emacs.

> t just adds the leading stars as `closest_headline_star_count + 1`

I don't know what that means. Could you please elaborate?

lua/orgmode/config/defaults.lua
kristijanhusak: +1 on this. `<Leader>o*` seems fine to me.

lua/orgmode/org/mappings.lua
```diff
@@ -295,6 +295,20 @@ function OrgMappings:todo_prev_state()
...
+  local line = vim.fn.getline('.')
+  local parent_level = Files.get_closest_headline().level
+
+  if (node:type() == 'headline' or node:parent():type() == 'headline') then
+    line = line:gsub('^(%*+)', string.rep(' ', parent_level - 1))
```
kristijanhusak, line 304: Make sure to check for `config.org_indent_mode` here. We do not need indent spaces if that option is not set to `indent`. You can find the similar check on several places in code.

lua/orgmode/org/mappings.lua
kristijanhusak: I think using a simple regex match would be much simpler in this case. There's no need to get treesitter nodes. 
From what I tested in emacs, if current line is a headline, it removes the asterisks, if it's not a headline, it just makes it a child of the parent headline (if any). 
Also, I think there is no need to preserve any indentation. From what I see in emacs, once the asterisks are removed, there is no indentation left, no matter of `org-indent-mode` setting.

lua/orgmode/org/mappings.lua
```diff
@@ -295,6 +295,31 @@ function OrgMappings:todo_prev_state()
...
+
+  if parent.line_number == vim.api.nvim_win_get_cursor(0)[1] then
+    line = line:gsub('^(%*+)', string.rep(' ', parent.level - 1))
+  else
+    line = line:gsub('^(%s*)', '')
+    if line:match('^[%*-]%s') then -- handle lists
+      line = line:gsub('^[%*-]%s', '') -- strip bullet
+      line = line:gsub('^%[([X%s])%]%s', function(checkbox_state)
+        if checkbox_state == 'X' then
+          return 'DONE '
```
jgollenz, line 310: This trailing whitespace is necessary here, since I capture (and thus replace) it with gsub or else we can't be sure whether the line doesn't actually look like this: `- [X]missing whitespace` and is not a valid checkbox.

Maybe there is a more elegant way to do this. If we stick with this logic, I will make it more explicit and `return 'DONE' .. ' '`

lua/orgmode/org/mappings.lua
```diff
@@ -295,6 +295,31 @@ function OrgMappings:todo_prev_state()
...
+        else
+          return 'TODO '
```
jgollenz, line 312: I'm now realizing that this does not take [org_todo_keywords](https://github.com/nvim-orgmode/orgmode/blob/master/DOCS.md#org_todo_keywords) into account. I will add this as well

lua/orgmode/org/mappings.lua
```diff
@@ -295,6 +295,31 @@ function OrgMappings:todo_prev_state()
   self:_todo_change_state('prev')
 end
 
+function OrgMappings:toggle_heading()
+  local line = vim.fn.getline('.')
+  local parent = Files.get_closest_headline()
```
jgollenz, line 300: This does not work on lines without a parent headline. However, Emacs Orgmode will turn such a line into a top-level headline.

lua/orgmode/org/mappings.lua
jgollenz: I went with the way @levouh proposed, checking if the line number of the closest headline is the current line number. Works fine, except for lines that have no closest headline (because they are on the top level). Any idea how to detect those?

@kristijanhusak yeah, unfortunately Emacs doesn't do this. I'll remove the logic then

jgollenz:
@kristijanhusak the orgmode [docs](https://orgmode.org/manual/Structure-Editing.html) talk about "regions" that can be toggled as a whole. I don't use Emacs, do you know what is meant by this term? Is this something that's already used by nvim-orgmode?

jgollenz:
@kristijanhusak I'd need your opinion regarding the visual-mode mappings I mentioned in #164 (bottom) to move on with this PR ;)

jgollenz:
@kristijanhusak one last thing: I cannot detect a line that has no parent-headline with `get_closest_headline()`, because it will throw an error. I hacked this together: 

```lua
function Files.has_headline(id)
  local current_file = Files.get_current_file()
  local msg = 'Make sure there are no errors in the document'
  if not vim.diagnostic then
    msg = string.format('%s by running :OrgDiagnostics command', msg)
  end
  if not current_file then
    error({ message = string.format('Failed to parse current file. %s.', msg) })
  end
  if not current_file:get_closest_headline(id) then
    return false
  else
    return true
  end
end
```

But it introduces some code duplication and I hate that. Do you have a better idea for how to check if a line even has a headline?

kristijanhusak:
@jgollenz I pushed small change where `get_closest_headline()` should return `nil` instead of throwing an error. Rebase and give it a try.

jgollenz:
@kristijanhusak works like a charm, thanks! You prefer squashing commits, right?

kristijanhusak:
@jgollenz Yeah, squashed commit is perfect.
I gave this a test, works great! Thanks!

Issue #87, opened 2021-09-11
LaTeX export doesn't use Emacs Config


shaunsingh:
### Are you using "tree-sitter" branch?

No

### Describe the bug

Exporting to latex (and latex-pdf) using <leader>oe doesn't respect my emacs config. My emacs config tells it to use tectonic instead of pandoc (along with a few other visual tweaks) so it would be nice to not have to leave neovim

### Steps to reproduce

1. Configure latex within emacs
2. Export to pdf from neovim
3. See export either fail or produce export without emacs config

### Expected behavior

It should utilize my emacs config

### Emacs functionality

Emacs uses my config to produce the pdf, which tells it to respect certain fonts and use tectonic, etc

### OS / Distro

Debian

### Neovim version/commit

0.6

kristijanhusak:
Sorry for a late response on this. Can you give me an example config that you have in your emacs so I can check? I'm not using exports so I'm not very familiar with it. Export is called via emacs exacutable , I thought it should pick up emacs configuration. Are you maybe using non default configuration file?

shaunsingh:
Sorry for the delay
My emacs config can be found here: https://github.com/shaunsingh/nix-darwin-dotfiles/tree/main/configs/doom
Now I have a different issue, it uses the emacs config but errors out
```lisp
[orgmode] Export error:
Debugger entered--Lisp error: (user-error "Unknown LaTeX class ‘chameleon’")
  user-error("Unknown LaTeX class `%s'" "chameleon")
  org-latex-make-preamble((:export-options nil :back-end #s(org-export-backend :name latex :parent nil :transcoders (... ... ... ... ... ... ... ... ... .
.. ... ... ... ... ... ... ... ... ... ... ... ... ... ... ... ... ...) :options (... ... ... ... ... ... ... ... ... ... ... ... ... ... ... ... ... ...
... ... ... ... ... ... ... ... ...) :filters (... ... ... ...) :blocks nil :menu (108 "Export to LaTeX" ...)) :translate-alist ((bold . org-latex-bold) (
center-block . org-latex-center-block) (clock . org-latex-clock) (code . org-latex-code) (drawer . org-latex-drawer) (dynamic-block . org-latex-dynamic-bl
ock) (entity . org-latex-entity) (example-block . org-latex-example-block) (export-block . org-latex-export-block) (export-snippet . org-latex-export-snip
[...]

PR #199, opened 2022-01-07
Bug Fix: emacs html export

aareman:
emacs html export requires the htmlize package which is no longer included in the latest orgmode for emacs (I'm using emacs 26.3). If you export to html -> emacs with a code block it breaks since it uses htmlize to syntax highlight. (btw, get the same error in emacs when trying to export)

That being said, when I install the htmlize package in emacs, exporting from emacs works, but still fails from orgmode.nvim. The reason is because the `--batch` flag doesn't source your emacs configuration and any packages you have. This PR adds a `--load` switch to the export command that points at your emacs config. This tells emacs running in batch mode to load a specific startup file, in this case the main emacs config.

> **NOTE** Changes beyond this. We may want to have a config option for specifying the location of an emacs config to use, since its not standard on all systems.

I know its failing tests/formatting. Just want to make sure this is an acceptable fix before polishing this PR. (See NOTE above for one outstanding issue)

### DIFF
--- a/lua/orgmode/export/init.lua
+++ b/lua/orgmode/export/init.lua
@@ -108,14 +108,18 @@ end
 function Export.emacs(format, extension)
   local file = vim.api.nvim_buf_get_name(0)
   local target = vim.fn.fnamemodify(file, ':p:r') .. '.' .. extension
-  if vim.fn.executable('emacs') ~= 1 then
+  local emacs = config.emacs_config.executable_path
+  local emacs_config_path = config.emacs_config.config_path
+  if vim.fn.executable(emacs) ~= 1 then
     return utils.echo_error('emacs executable not found. Make sure emacs is in $PATH.')
   end
 
   local cmd = {
-    'emacs',
+    emacs,
     '-nw',
     '--batch',
+    '--load',
+    emacs_config_path,
     string.format('--visit=%s', file),
     string.format('--funcall=%s', format),
   }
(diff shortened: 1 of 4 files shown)


kristijanhusak:
Didn't know about that, thanks. I guess this should also fix https://github.com/nvim-orgmode/orgmode/issues/87. 
Only issue that I see with this is that it's hardcoded. People may be using different path like `$HOME/.emacs` for example.
Lets do this:

* Introduce a property `emacs_config` in the `config/defaults.lua` file that should look something like this:
```lua
emacs_config = {
  executable_path = 'emacs',
  config_path = '$HOME/.emacs.d/init.el`
}
```
* Use the above config in the exporter 

That should solve everything, and make it configurable.

aareman:
@kristijanhusak I've made the updates and fixed the formatting issues. I also added basic documentation for it in the DOC.md file, Is there anywhere else I should put it?

lua/orgmode/export/init.lua
```diff
@@ -111,11 +111,15 @@ function Export.emacs(format, extension)
   if vim.fn.executable('emacs') ~= 1 then
     return utils.echo_error('emacs executable not found. Make sure emacs is in $PATH.')
   end
+  local emacs = config.emacs_config.executable_path
```
kristijanhusak, line 114: Move this above to line 110, and include it in the check on line 111 instead of `'emacs'` string.

kristijanhusak:
@aareman looks ok, just address that one comment, and please squash commits into a single one. Thanks.

lua/orgmode/export/init.lua
aareman: Done

aareman:
I made the changes, and squashed the commits.

kristijanhusak (approved):

kristijanhusak:
Awesome, thanks!

PR #201, opened 2022-01-13
fix(mappings): Default offset encoding when handling <CR> press

levouh:
Per [here](https://github.com/neovim/neovim/issues/14090) there are a number of functions that now require specifying an `offset_encoding`. Some examples of other updates to things like [nvim-cmp](https://github.com/hrsh7th/nvim-cmp/commit/614e00de2424d1c19a3f653120b63d5b2d10b21a) and [null-ls](https://github.com/jose-elias-alvarez/null-ls.nvim/commit/e5f71b759618f0c9613d0f7cda86d2124ce1361e).

Looking at all of those functions, this is the only one that I found that gets used in this plugin. As we do not have any attached clients, I do not see this being a huge issue (although I am also pretty naive in terms of the usage and impact) to follow the "norm" and use `utf-16` as a default.

### DIFF
--- a/lua/orgmode/org/mappings.lua
+++ b/lua/orgmode/org/mappings.lua
@@ -7,6 +7,7 @@ local Hyperlinks = require('orgmode.org.hyperlinks')
 local PriorityState = require('orgmode.objects.priority_state')
 local TodoState = require('orgmode.objects.todo_state')
 local config = require('orgmode.config')
+local constants = require('orgmode.utils.constants')
 local ts_utils = require('nvim-treesitter.ts_utils')
 local utils = require('orgmode.utils')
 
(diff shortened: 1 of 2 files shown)


levouh:
Moving to a draft to research & experiment a bit more.

kristijanhusak:
LGTM. I'm a bit confused why utf-16 instead of utf-8, but since [Michael suggested it](https://github.com/hrsh7th/nvim-cmp/commit/614e00de2424d1c19a3f653120b63d5b2d10b21a#r63818690), I guess that's the right path.

kristijanhusak (approved):

kristijanhusak:
Thanks!

mjlbach:
I think the correct fix for you all is using `utf-8`, `utf-16` is only correct if you are applying text edits from a language server. `utf-8` assumes the position is encoded with raw bytes, `utf-16` assumes code points are encoded with one or two 16-bit code units.

Issue #196, opened 2022-01-06
Feat: Refile to heading from capture


joshpetit:
### Does this feature exist in Emacs orgmode core?

Yes

### Orgmode link

https://orgmode.org/manual/Template-elements.html
```
(setq org-capture-templates
      '(("w" "Work time (file+headline "~/org/refile.org" "Work Timesheet")
         "** Working on %?\n%T")))
```

### Discussed in https://github.com/nvim-orgmode/orgmode/discussions/184

<div type='discussions-op-text'>

<sup>Originally posted by **joshpetit** December 13, 2021</sup>
I'm not sure if this is implemented yet, but I'd like to refile a template to a heading. For example having something like this:

```lua
    org_agenda_templates = {
        w = {
            description = 'Work time',
            template = '** Working on  %<%Y-%m-%d>\n%?',
            target = 'refile.org/Work Timesheet'
        },
    },
```

I'm not sure if I just do't have to syntax correct, if it's not implemented yet I can file a feature request.</div>

joshpetit:
@kristijanhusak  I could get started working on this, the API I had in mind would be:

```lua
    org_agenda_templates = {
        w = {
            description = 'Work time',
            template = '** Working on  %<%Y-%m-%d>\n%?',
            target = 'refile.org'
            headline= 'Work Timesheet'
        },
    },
```
I think it would make sense to make the target include the header (i.e `refile.org/Work Timesheet`) just as you do for refiling within captures, but I think it matches the original API more by having it as a property. Thoughts?

jgollenz:
Neat! With this path-like syntax you could also extend it for non-unique headlines (e.g. `refile.org/Work Timesheet/2022/projectA`), as long as there are no headlines with the same name on the same level. That would work regardless of it being a property or not. The property variant makes it a bit more clear imo.

Not really your focus with this issue, but how would this scale if we were to include the other target possibilities that orgmode offers (e.g. regex, id, etc.)?

kristijanhusak:
@joshpetit that looks ok. Should be fairly simple to implement since refile prompt already supports something similar.

joshpetit:
finished

PR #202, opened 2022-01-13
Add refile_to_headline

joshpetit:
#196 

Haven't written a test yet but would appreciate feed back on whether it looks like a good implementation, thanks


kristijanhusak:
This looks ok. I don't think you need to throw an error, just let it refile to the end of the file if there's no headline found.

joshpetit:
just fixed the lint, I see that there are no tests yet, I'm pretty confused on how I would write one for this. Could you give me an example of how one might look here?  I've taken a look at the agenda and template spec.

kristijanhusak:
It's a bit tricky to write tests for this because it manages multiple buffers. It's ok to not have test for this for now.

kristijanhusak:
Looks good, thanks!

PR #208, opened 2022-01-26
Add '%b' as variable for file-basename when parsing archive-location

### LINKED ISSUE #206: Expose filename as variable for archive location

N/A

### Orgmode link

_No response_

### Feature value

Currently there is one variable `%s` that expands to the path of the current file. This makes it impossible to create a hierarchy like the following:

```
~/org/
    todo.org
    ... 
    archive/
        todo.org_archive
        ...
```
Because setting `org_archive_location` to `~/org/archive/%s_archive` will expand to `/home/user/org/archive//home/user/org/todo.org_archive`, since the `%s` variable holds the full path. What would be needed is e.g. `%b` that holds only the basename of the current file, `todo.org`.

As for **feature value**: no need to clutter your org directory with archive files while at the same time not needing to give up granularity

### Additional context

_No response_

jgollenz:
closes #206

@kristijanhusak let me know what you think of this. I found [this TODO](https://github.com/nvim-orgmode/orgmode/blob/master/lua/orgmode/config/init.lua#L242) about archiving to headlines. Any specific plans on how to implement that, while I'm at it?

### DIFF
--- a/lua/orgmode/config/init.lua
+++ b/lua/orgmode/config/init.lua
@@ -242,6 +242,7 @@ function Config:parse_archive_location(file, archive_loc)
   -- TODO: Support archive to headline
   local parts = vim.split(archive_loc, '::')
   local archive_location = vim.trim(parts[1])
+  archive_location = archive_location:gsub('%%b', vim.fn.fnamemodify(file, ':t'))
   if archive_location:find('%%s') then
     return string.format(archive_location, file)
   end


TravonteD:
I think that we could avoid adding `%b` and instead have `%s` always be the `basename`, with the default location being the value `org_agenda_files`. That seems to be how the behavior is supposed to work in the first place. https://orgmode.org/manual/Moving-subtrees.html#index-archive-locations

jgollenz:
Currently the default value of `org_archive_location` is `%s_archive`, which expands to `full/path/to/file_archive`. Let's say `%s` only expands to `file` and not `full/path/to/file`. Now for the default to be saved to the same directory as the file from which we archive from (which is what we expect), it would be necessary to programmatically prepend the expanded default value with the path to that directory, `full/path/to`, right? That effectively locks any archive file to the directory of the file from which you archive. Except if you make the default value a special-case and only prepend the full path when `org_archive_location` is set to the default of `%s_archive`, which is confusing imo. 

So if I'm not overlooking something, expanding `%s` to basename makes it impossible to have archive files in other directories than the ones from which you archive. I'm not sure how emacs handles this, but it is definitely possible in nvim-orgmode to have your archive file in another directory

kristijanhusak:
@jgollenz @TravonteD is actually right. That's how Emacs does it. `%s` is a basename, and it knows how to read it.
This is what Emacs says about this variable:
```
When the filename is omitted, archiving happens in the same file.
%s in the filename will be replaced by the current file
name (without the directory part
```

Here's few examples from emacs orgmode docs:

"%s_archive::"
	If the current file is Projects.org, archive in file
	Projects.org_archive, as top-level trees.  This is the default.

"~/org/archive.org::"
	Archive in file ~/org/archive.org (absolute path), as top-level trees.

So I guess we just need to tweak this to not read the full path, but to know how to replace only the file name part. Once this is fixed, you could have something like `archive/%s_archive::` as configuration, and it would know how to handle your use case from the issue.

I'll look into this.

kristijanhusak:
@jgollenz I pushed a fix to address this. It should be working now as expected.
This means that you can do `archive/%s_archive::` and everything will go under the archive/filename.org_archive.
You can keep me posted on the issue if something is not working as expected. Thanks for catching this.

jgollenz:
It works when the archive directory I want is a sub-directory of the one my file is in. But trying to set `org_archive_location` to something like `~/where/my/logs/are/%s_archive::` will create the log file (+directory structure) at `~/org/~/where/my/logs/are/`

So I cannot archive to an arbitrary location. Not sure if emacs can either, but I think it's a valid option to give to users. Then again, I'm biased regarding this issue

kristijanhusak:
@jgollenz You cannot have an absolute location that has a placeholder. To achieve your example from issue:
```
~/org/
    todo.org
    ... 
    archive/
        todo.org_archive
        ...
```

You would set `org_archive_location = 'archive/%_archive`, which would for `todo.org` expand to `~/org/archive/todo.org_archive`. To put it somewhere else you could theoretically do something like `../../archive/%s_archive`, but I didn't try that, and I woudn't suggest doing that.

jgollenz:
@kristijanhusak sorry, I took another look at my original issue and your fix indeed addresses that. 

> You cannot have an absolute location that has a placeholder

I'm not sure if you mean it is impossible or it's not possible with your fix? If it's the former, could you please explain why? The issue I have is that the apparent emacs behavior locks users into the directory of the file they're archiving from, but only when they use the `%s` placeholder. I can happily archive to `~/path/to/wherever/archive.org` but not to `~/path/to/wherever/%s_archive.org`. When `%s` really is just the basename, why can't I do that? This surely is a different issue to the one I reported, but its related

jgollenz:
I guess it's just this discrepancy between those two cases:

1. I want to use the placeholder, but I can only do so in the directory the file is in
2. I want to archive to another directory, but I can only do so without using a placeholder

And thinking about it more: it is because there is no way to tell where the default location should be. Now I understand. Well, bummer.

PR #209, opened 2022-01-30
Restrict cursor movement in calendar

nathom:
This PR makes the default cursor movements "magnetic", so that they stick onto the days. I think this behavior is much more intuitive than leaving `hjkl` to be the same as it is for plain text.

Demo:

[asciicast](https://asciinema.org/a/ErRBNen2uEIBb37I2l2oEgG0T)


kristijanhusak:
Works great, thanks!

PR #220, opened 2022-02-15
[doc] Replace org_agenda_file by org_agenda_files

maxmahlke:
Several config examples use the non-existing `org_agenda_file` variable, which I replace by the correct `org_agenda_files`


kristijanhusak:
Good catch, thanks!

Issue #234, opened 2022-03-24
Adding custom queries


ranjithshegde:
### Discussed in https://github.com/nvim-orgmode/orgmode/discussions/233

<div type='discussions-op-text'>

<sup>Originally posted by **ranjithshegde** March 22, 2022</sup>
Since setting [conceal](https://github.com/neovim/neovim/pull/17776) from treesitter has been merged in core, I have been trying to set my own queries for org bullets like this
```query
(
((stars) @stars (#eq? @stars "*")) @_conceal
(#set! conceal "◉")
)
(
((stars) @stars (#eq? @stars "**")) @_conceal
(#set! conceal "○")
)
(
((stars) @stars (#eq? @stars "***")) @_conceal
(#set! conceal "✸")
)
(
((stars) @stars (#eq? @stars "****")) @_conceal
(#set! conceal "✿")
)
```
When I append the above snippet to `packer/*/orgmode/queries/org/highlights.scm` it works perfectly. However when I do it in my `~/.config/nvim/after/queries/org/highlights.scm` it does not apply. 

according to @kristijanhusak's checking, disabling todo-keyfaces highlight solves this as a workaround for now  </div>

gerritvreeman:
I think the issue is [here](https://github.com/nvim-orgmode/orgmode/blob/e1f3054987ce054525258d9a3cc5837bf6e75212/lua/orgmode/colors/todo_highlighter.lua#L39). According to the treesitter docs ([here](https://github.com/nvim-treesitter/nvim-treesitter#adding-queries)), the `set_query()` function will override the defaults, so any user query files in `/after/queries/highlights.scm` are ignored after running `filetype detect`.

The solution might be to iterate over all query_files and gather the user-set queries then call `vim.treesitter.set_query()` and `filetype detect` at the end.

kristijanhusak:
@ranjithshegde can you check if it works now?

ranjithshegde:
I somehow missed the merge notification. Yes it works!!
Thanks @kristijanhusak and @gerritvreeman for your wonderful work

PR #240, opened 2022-03-31
fix custom highlight queries

gerritvreeman:
Fix for #234 by gathering all highlight queries then calling vim.treesitter.set_query().


kristijanhusak:
Thanks!

Issue #248, opened 2022-04-07
Support Multi-Key Capture shortcuts


TravonteD:
### Does this feature exist in Emacs orgmode core?

Yes

### Orgmode link

https://orgmode.org/manual/Template-elements.html

> When using several keys, keys using the same prefix key must be sequential in the list and preceded by a 2-element entry explaining the prefix key

### Additional context

This would be a great addition since it would allow for categories of captures. For example, a "p" prefix for project-level captures that write to a readme.org within a source repo.

jgollenz:
Ah nice! I didn't know that orgmode offers that. I just thought about adding this in my personal fork recently, as I already have some collisions for capture shortcut chars. Have not looked into it yet, but I strongly second this feature :+1:

TravonteD:
I'm working on a PR to add this functionality

kristijanhusak:
This was added by @TravonteD in https://github.com/nvim-orgmode/orgmode/pull/253. Thanks!

PR #253, opened 2022-04-12
#248 Add multi-key capture shortcut support

TravonteD:
This adds the orgmode functionality documented
here: https://orgmode.org/manual/Template-elements.html

### DIFF
--- a/lua/orgmode/capture/init.lua
+++ b/lua/orgmode/capture/init.lua
@@ -18,21 +18,49 @@ function Capture:new()
   return data
 end
 
-function Capture:prompt()
-  local templates = {}
-  for key, template in pairs(self.templates:get_list()) do
-    table.insert(templates, {
-      label = template.description,
-      key = key,
-      action = function()
-        return self:open_template(template)
-      end,
-    })
+function Capture:_get_subtemplates(base_key, templates)
+  local subtemplates = {}
+  for key, template in pairs(templates) do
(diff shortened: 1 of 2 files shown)


TravonteD:
@kristijanhusak I've run `make format`, but the format ci still fails. Anything else that I need to do?

TravonteD:
Got it, needed a stylua update

lua/orgmode/capture/init.lua
```diff
@@ -18,21 +18,49 @@ function Capture:new()
...
+  end
+  return subtemplates
+end
+
+function Capture:_create_menu_items(templates)
+  local menu_items = {}
+  for key, template in pairs(templates) do
+    if string.len(key) == 1 then
+      local item = {
+        label = template.description,
```
kristijanhusak, line 36: label is `nil` when this is a "parent" template (string).

kristijanhusak (changes requested):
Generally works ok, only that issue with the label.
I tested how Emacs does this, and basically it's same, but it has some indications that there are nested templates. Here's the gif:
nested-template
If you think those would be simple to add, that would be great. If it complicates it, I believe this implementation is also good enough.

TravonteD:
I've added the ellipses indicator for toplevel templates as well, it was straight forward to add.

kristijanhusak (approved):
Looks good, thanks!

PR #257, opened 2022-04-17
Reveal headline when jumping from agenda

joaomsa:
When jumping to a headline from the agenda with `<CR>` or `<Tab>` and if the file
is folded, just setting the cursor doesn't reveal the folds.

By calling `zv` after positioning the cursor we reveal up to the cursor


kristijanhusak:
Thanks!

cands:
Great, thank you. I think the same functionality would be useful when opening hyperlinks (mapped to <Leader>oo by default), if easy to implement for that purpose as well.

jgollenz:
@cands #262

PR #258, opened 2022-04-18
Fix lowercase src highlighting

### LINKED ISSUE #86: Lowercase `#+(begin|end)_src` blocks not syntax highlighted
### Are you using "tree-sitter" branch?

No

### Describe the bug

`#+BEGIN_SRC` and `#+END_SRC` block is properly syntax highlighted, but `#+begin_src` and `#+end_src` block is not.

### Steps to reproduce

This is my test block:
```
#+BEGIN_SRC r
library(tidyverse)
#+END_SRC

#+begin_src r
library(tidyverse)
#+end_src
```

I then saved and reloaded the file.

### Expected behavior

Both blocks should have syntax highlighting for the language specified

### Emacs functionality

Conventionally, uppercase `BEGIN_|END_` is used, but [lowercase is also accepted](https://orgmode.org/manual/Conventions.html)

### OS / Distro

Arch Linux

### Neovim version/commit

v0.6.0-dev+218-g3f526feeb

### Additional context

_No response_

TravonteD:
Closes #86 by adding the lowercase `begin_src/end_src` to the syntax match group for language highlighting

### DIFF
--- a/syntax/org_legacy.vim
+++ b/syntax/org_legacy.vim
@@ -70,10 +70,10 @@ syntax match org_list_item /.*$/ contained contains=org_bold,org_italic,org_unde
 
 " Block Delimiters: {{{1
 syntax case ignore
-syntax match  org_block_delimiter /^\s*#+BEGIN_.*/
-syntax match  org_block_delimiter /^\s*#+END_.*/
+syntax match  org_block_delimiter /^\s*#+\(BEGIN_\|begin_\).*/
+syntax match  org_block_delimiter /^\s*#+\(END_\|end_\).*/
 syntax match  org_key_identifier  /^#+[^ ]*:/
-syntax match  org_title           /^#+TITLE:.*/  contains=org_key_identifier
+syntax match  org_title           /^#+\(TITLE\|title\):.*/  contains=org_key_identifier
 hi def link org_block_delimiter Comment
 hi def link org_key_identifier  Comment
 hi def link org_title           Title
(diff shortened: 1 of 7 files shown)


TravonteD:
@kristijanhusak I'm looking into what's happening with the failing specs. Running the failing test files individually seems to pass, but when running `make test` they fail. This is with the latest `plenary.nvim`. Let me know if you have any insight.

kristijanhusak:
I think I'll have to drop support for versions under 0.7. Nvim-treesitter requires 0.7, and it doesn't have any tag versioning that can be used for older Neovim versions. I'll announce it on the breaking change issue.

kristijanhusak:
Regarding this change, there is an open PR https://github.com/nvim-orgmode/orgmode/pull/89 that converts everything to lowercase, but it's really outdated. Seems like lowercase keywords are the standard https://list.orgmode.org/87tuuw3n15.fsf@nicolasgoaziou.fr/ . I'd like to have everything compatible with the lowercase, not only begin/end src. Is there any specific reason why you set up only begin/end src?

TravonteD:
> Is there any specific reason why you set up only begin/end src?
 
Yes, the issue was that language highlighting didn't work, and that only applies to `src` blocks. `quote` and `example` blocks behave identically already.

kristijanhusak:
@TravonteD do you have any idea why tests return exit status 2? I can't reproduce it locally, it returns 0 correctly.

tests/plenary/org/autocompletion_spec.lua
```diff
@@ -150,8 +150,8 @@ describe('Autocompletion', function()
 
     result = OrgmodeOmniCompletion(0, '#+B')
```
kristijanhusak, line 151:
```suggestion
    result = OrgmodeOmniCompletion(0, '#+b')
```

syntax/org_legacy.vim
```diff
@@ -87,9 +87,9 @@ hi def link org_title           Title
 syntax match  org_verbatim /^\s*>.*/
 syntax match  org_code     /^\s*:.*/
 
-syntax region org_verbatim start="^\s*#+BEGIN_.*"      end="^\s*#+END_.*"      keepend contains=org_block_delimiter
-syntax region org_code     start="^\s*#+BEGIN_SRC"     end="^\s*#+END_SRC"     keepend contains=org_block_delimiter
-syntax region org_code     start="^\s*#+BEGIN_EXAMPLE" end="^\s*#+END_EXAMPLE" keepend contains=org_block_delimiter
+syntax region org_verbatim start="^\s*#+\(BEGIN_\|begin_\).*"      end="^\s*#+\(END_\|end\).*"      keepend contains=org_block_delimiter
```
kristijanhusak, line 90:
```suggestion
syntax region org_verbatim start="^\s*#+\(BEGIN_\|begin_\).*"      end="^\s*#+\(END_\|end_\).*"      keepend contains=org_block_delimiter
```

tests/plenary/org/autocompletion_spec.lua
kristijanhusak: This is causing tests to fail.

kristijanhusak:
[This line](https://github.com/nvim-orgmode/orgmode/blob/master/lua/orgmode/parser/file.lua#L367) should also be updated to this to properly detect filetypes:
```lua
  local blocks = self:get_ts_matches('(block name: (expr) @name parameter: (expr) @parameters (#any-of? @name "SRC" "src"))')
```

kristijanhusak (approved):

kristijanhusak:
Thanks!

PR #259, opened 2022-04-18
Make menu prompts more uniformly

jgollenz:
The separator char for the capture prompt in Emacs appears to be `=` ([here](https://youtu.be/qCdScs4YO8k?t=98)) but I think we can get by making them all `-` instead. As every menu should use a separator after the title, we can also get rid of some magic numbers.

### DIFF
--- a/lua/orgmode/agenda/init.lua
+++ b/lua/orgmode/agenda/init.lua
@@ -65,7 +65,6 @@ end
 function Agenda:prompt()
   self.filters:reset()
   return utils.menu('Press key for an agenda command', {
-    { label = '', separator = '-', length = 34 },
     {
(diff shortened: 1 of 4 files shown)


jgollenz:
@kristijanhusak hm, looks like I'm getting the same errors as #258 
`make test` works locally for me. nvim v0.6

TravonteD (approved):
Tested and the prompts look good, nice work 👍🏾

kristijanhusak:
> @kristijanhusak hm, looks like I'm getting the same errors as #258 `make test` works locally for me. nvim v0.6

This is due to nvim-treesitter requiring Neovim v0.7. I announced that there will be only support for v0.7 going forward because of it.

PR #263, opened 2022-04-20
feat(Api): Update to new autocmd API

ranjithshegde:
If nvim #version is `0.7` then call autocmds using the new API.

Not complete yet. There are 4 more autocmds that are not in an augroup.
@kristijanhusak Should I create an augroup for the files in help, calendar and notification modules?


kristijanhusak:
@ranjithshegde you can remove the check for `nvim-0.7` and just use new api directly without the fallback to the old cmds. I officially dropped support for lower versions because of nvim-treesitter.
You can add augroup where you think it's needed, I probably missed it on few spots.

ranjithshegde:
>  You can add augroup where you think it's needed, I probably missed it on few spots.

I have created 4 new `augroups`. Not every elegant. If you dont mind then its ready for review/merge. 
Unless, if the `setup_autocmd()` function in `lua/init.lua` is called before the the modules `edit_special`, `calendar`, `notifications` and `help`, then I can setup a single `augroup` for all of them, something like `org_modules` or something?

lua/orgmode/notifications/notification_popup.lua
```diff
@@ -37,7 +37,12 @@ function NotificationPopup:show()
   vim.defer_fn(function()
     pcall(vim.api.nvim_win_close, self.win, true)
   end, self.hide_after)
-  vim.cmd(string.format('autocmd BufEnter <buffer=%d> ++once :bw!', self.buf))
+  vim.api.nvim_create_augroup('org_notification', { clear = true })
+  vim.api.nvim_create_autocmd('BufEnter', {
```
TravonteD, line 41: Add `once` to the options here

lua/orgmode/init.lua
```diff
@@ -47,17 +47,35 @@ function Org:reload(file)
...
-  vim.cmd([[augroup END]])
+  vim.api.nvim_create_augroup('orgmode_nvim', { clear = true })
+  vim.api.nvim_create_autocmd('BufWritePost', {
+    pattern = { '*.org', '*.org_archive' },
+    group = 'orgmode_nvim',
+    callback = function()
+      require('orgmode').reload(vim.fn.expand('<afile>:p'))
+    end,
+  })
+  vim.api.nvim_create_autocmd({ 'BufWritePost', 'BufWritePost' }, {
```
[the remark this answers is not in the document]
ranjithshegde: Whoops! will fix it

TravonteD (approved):
Looks good!

lua/orgmode/capture/init.lua
```diff
@@ -292,8 +307,11 @@ function Capture.autocomplete_refile(arg_lead)
 end
 
 function Capture:kill()
-  vim.cmd([[autocmd! OrgCapture BufWipeout <buffer>]])
-  vim.api.nvim_win_close(0, true)
+  vim.api.nvim_create_autocmd('BufWipeout', {
+    buffer = 0,
+    group = 'OrgCapture',
+    callback = 'nvim_win_close(0, true)',
```
kristijanhusak, line 313: Are you sure this will work? I think `callback` needs to be lua function:
```lua
callback = function()
  return vim.api.nvim_win_close(0, true)
end
```

lua/orgmode/init.lua
```diff
@@ -47,17 +47,35 @@ function Org:reload(file)
...
-  vim.cmd(
-    [[autocmd BufWritePost *.org,*.org_archive call luaeval('require("orgmode").reload(_A)', expand('<afile>:p'))]]
-  )
-  vim.cmd(
-    [[autocmd BufReadPost,BufWritePost *.org,*.org_archive lua require('orgmode.org.diagnostics').print_error_state()]]
-  )
-  vim.cmd([[autocmd FileType org call luaeval('require("orgmode").reload(_A)', expand('<afile>:p'))]])
-  vim.cmd([[autocmd CursorHold,CursorHoldI *.org,*.org_archive lua require('orgmode.org.diagnostics').report()]])
-  vim.cmd([[augroup END]])
+  vim.api.nvim_create_augroup('orgmode_nvim', { clear = true })
```
kristijanhusak, line 50: Put this into variable and use the variable in the `nvim_create_autocmd` instead of string

lua/orgmode/capture/init.lua
ranjithshegde: I read in help that it does. but yeah, safer from lua. will fix

lua/orgmode/init.lua
kristijanhusak: @ranjithshegde I see you marked this as resolved, but it's not actually changed. This is what I suggest:
```suggestion
local org_augroup = vim.api.nvim_create_augroup('orgmode_nvim', { clear = true })
```

and below in autocmds set `group = org_augroup`. Example:

```lua
  vim.api.nvim_create_autocmd('BufWritePost', {
    pattern = { '*.org', '*.org_archive' },
    group = org_augroup,
    callback = function()
      require('orgmode').reload(vim.fn.expand('<afile>:p'))
    end,
  })
```

lua/orgmode/capture/init.lua
```diff
@@ -292,8 +307,13 @@ function Capture.autocomplete_refile(arg_lead)
 end
 
 function Capture:kill()
-  vim.cmd([[autocmd! OrgCapture BufWipeout <buffer>]])
-  vim.api.nvim_win_close(0, true)
+  vim.api.nvim_create_autocmd('BufWipeout', {
```
kristijanhusak, line 310: Here you actually need to remove the `BufWipeout` command to avoid it from triggering once the window is closed. I believe you need to save the initial `BufWipeout` into a variable and then delete it via `nvim_del_autocmd()`

ranjithshegde: It does have the `once` flag set to true. With that in mind, is removing it still necessary?

kristijanhusak: Yeah, because when you trigger this `Capture:kill()` function, BufWipeout wasn't triggered even once. `nvim_win_close` would trigger it, and it would cause it to save something that shouldn't be saved.

ranjithshegde: Should I delete it in the same call `Capture:kill()`?

kristijanhusak: Yes, this old command `vim.cmd([[autocmd! OrgCapture BufWipeout <buffer>]])` basically removes the autocmd defined previously.

kristijanhusak:
Thanks for PR! I'll resolve this small issue with clearing autocmd for capture buffer.

PR #264, opened 2022-04-22
Check for 'end of string' when searching headline

### LINKED ISSUE #254: Refiling doesnt work when another headline starts with the same string
### Describe the bug

If you refile to a specific headline, but there is a headline above that headline that starts with the same text it will refile to it rather than the one you specify

### Steps to reproduce

1. Create 2 org files
2. In the 2nd org file create two headlines
3. Headline 1: "Start ajsijasigfajsija"
4. Headline 2: "Start"
5. Try to refile something from file 1 to 2 to the headline "Start"

### Expected behavior

I expect it to go to headline 2

### Emacs functionality

_No response_

### Minimal init.lua

```lua
vim.cmd [[set runtimepath=$VIMRUNTIME]]
vim.cmd [[set packpath=/tmp/nvim/site]]

local function load_plugins()
  require('packer').startup {
    {
      'wbthomason/packer.nvim',
      {'nvim-treesitter/nvim-treesitter', run = ':TSUpdate'},
      {'kristijanhusak/orgmode.nvim', branch = 'master' },
    },
    config = {
      package_root = package_root,
    },
  }
end

_G.load_config = function()
  require('nvim-treesitter.configs').setup({})
  
  vim.cmd[[packadd nvim-treesitter]]
  vim.cmd[[TSUpdate]]

  require('orgmode').setup()
end

  vim.fn.system { 'git', 'clone', 'https://github.com/wbthomason/packer.nvim', install_path }
  load_plugins()
  require('packer').sync()
  vim.cmd [[autocmd User PackerComplete ++once lua load_config()]]
else
  load_plugins()
  require('packer').sync()
  _G.load_config()
end
```

nvim-org-bug

### OS / Distro

Arch

### Neovim version/commit

0.7

### Additional context

_No response_

jgollenz:
fixes #254

### DIFF
--- a/lua/orgmode/parser/file.lua
+++ b/lua/orgmode/parser/file.lua
@@ -202,9 +202,13 @@ end
 
 ---@param title string
 ---@return Section[]
-function File:find_headlines_by_title(title)
+function File:find_headlines_by_title(title, exact)
   return vim.tbl_filter(function(item)
-    return item.title:lower():match('^' .. vim.pesc(title:lower()))
+    local pattern = '^' .. vim.pesc(title:lower())
+    if exact then
+      pattern = pattern .. '$'
+    end
+    return item.title:lower():match(pattern)
   end, self.sections)
 end
 
@@ -238,8 +242,8 @@ end
 
 ---@param title string
 ---@return Section
-function File:find_headline_by_title(title)
-  local headlines = self:find_headlines_by_title(title)
+function File:find_headline_by_title(title, exact)
+  local headlines = self:find_headlines_by_title(title, exact)
   return headlines[1]
 end
 
--- a/tests/plenary/ui/mappings_spec.lua
+++ b/tests/plenary/ui/mappings_spec.lua
@@ -1,6 +1,9 @@
 local helpers = require('tests.plenary.ui.helpers')
+local org = require('orgmode')
 local config = require('orgmode.config')
+local Files = require('orgmode.parser.files')
 local Date = require('orgmode.objects.date')
+local Capture = require('orgmode.capture')
 
 describe('Mappings', function()
   after_each(function()
@@ -1567,5 +1570,30 @@ describe('Mappings', function()
     assert.are.same('* Test orgmode', vim.fn.getline(1))
     vim.cmd('norm ,o,a\r')
     assert.are.same('* [#A] Test orgmode', vim.fn.getline(1))
+  end)
+
+  it('should refile to headline that matches name exactly', function()
+    local destination_file = helpers.load_file_content({
+      '* foobar',
+      '* baz',
+      '** foo',
+    })
+
+    local source_file = helpers.load_file_content({
+      '* to be refiled',
+      '* not to be refiled',
+    })
+
+    source_file = Files.get_current_file()
+    local item = source_file:get_closest_headline()
+    org.instance().capture:refile_to_headline(destination_file, source_file:get_headline_lines(item), item, 'foo')
+    assert.are.same('* not to be refiled', vim.fn.getline(1))
+    vim.cmd('edit' .. vim.fn.fnameescape(destination_file))
+    assert.are.same({
+      '* foobar',
+      '* baz',
+      '** foo',
+      '*** to be refiled',
+    }, vim.api.nvim_buf_get_lines(0, 0, 5, false))
   end)
 end)
--- a/lua/orgmode/capture/init.lua
+++ b/lua/orgmode/capture/init.lua
@@ -187,7 +187,7 @@ function Capture:refile_to_headline(destination_file, lines, item, headline_titl
   local agenda_file = Files.get(destination_file)
   local headline
   if headline_title then
-    headline = agenda_file:find_headline_by_title(headline_title)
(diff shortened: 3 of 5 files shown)


kristijanhusak:
@jgollenz There is no easy way to write a test that is using prompts, which refiling does use. I would suggest to just create a test file, and call the code directly to assert if it returns what you expect.

tests/plenary/ui/mappings_spec.lua
```diff
@@ -1568,4 +1568,28 @@ describe('Mappings', function()
...
+    })
+
+    local source_file = helpers.load_file_content({
+      '* to be refiled',
+      '* not to be refiled',
+    })
+
+    vim.fn.cursor(1, 1)
+    vim.cmd([[norm ,or]])
+    -- TODO: this is wrong, not a command. how to do the actual refiling?
```
TravonteD, line 1586: The function for this test is likely 
```lua
Capture:refile_to_headline()
```

tests/plenary/ui/mappings_spec.lua
```diff
@@ -1568,4 +1568,28 @@ describe('Mappings', function()
...
+    vim.cmd([[exe "destination.org/foo\<CR>"]])
+    assert.are.same('* not to be refiled', vim.fn.getline(1))
+    -- TODO: how to switch to the destination file?
```
TravonteD, line 1589: you could use a fixture for the destination file and load it using `vim.cmd[[ edit <destfile> ]]` then from there use `vim.api.nvim_buf_get_lines` to get the list of lines to assert against

jgollenz:
Thanks @TravonteD, I got it working :) however, I'm unsure if requiring `Files` is necessary here. I don't want to clutter the imports. What do you think @kristijanhusak?

kristijanhusak:
@jgollenz it's ok to use whatever you need in the test.

tests/plenary/ui/mappings_spec.lua
kristijanhusak: This works because Capture "class" doesn't have a lot of it's own internal state, but we should avoid calling these things directly. Capture is instantiated on the main Org class.
I just pushed a small change that can be used here like this:
```lua
require('orgmode').instance().capture:refile_to_headline()
```

tests/plenary/ui/mappings_spec.lua
```diff
@@ -1568,4 +1570,29 @@ describe('Mappings', function()
...
+    })
+
+    local source_file = helpers.load_file_content({
+      '* to be refiled',
+      '* not to be refiled',
+    })
+
+    source_file = Files.get_current_file()
+    local item = source_file:get_closest_headline()
+    Capture:refile_to_headline(destination_file, source_file:get_headline_lines(item), item, 'foo')
```
kristijanhusak, line 1588:
```suggestion
    local org = require('orgmode') -- you can move this up outside of tests
    org.instance().capture:refile_to_headline(destination_file, source_file:get_headline_lines(item), item, 'foo')
```

jgollenz: When I add this change, `instance` is a nil value field (which is also not callable, but removing the parenthesis makes no difference, still nil)

kristijanhusak: You need to rebase from master branch. I added `instance()` function recently.

jgollenz: Thanks, fixed it

lua/orgmode/parser/file.lua
```diff
@@ -204,7 +204,7 @@ end
 ---@return Section[]
 function File:find_headlines_by_title(title)
```
kristijanhusak, line 205: I just figured out that this method is also used for autocompletion, which should actually do matching only by the beginning of the string. Let's add a second argument called `strict` which would default to true, and do matching as it is done now. If `strict` is false, do it the old way.
Once that's added, just pass `false` [here](https://github.com/nvim-orgmode/orgmode/blob/master/lua/orgmode/org/hyperlinks.lua#L70).
Another option, maybe even a bit cleaner, is to introduce a method called `File:find_headlines_matching_title` that would behave in the old way, and use that in hyperlinks autocompletion.

jgollenz: I'd prefer the version where we pass a second argument, as I'd rather reduce than increase the amount of functions we have. Also `exact` would be better than `strict` in my opinion. Just let me know which you want to settle on.

kristijanhusak: Sure that works. Exact is better than strict.

jgollenz: What about here: https://github.com/nvim-orgmode/orgmode/blob/master/lua/orgmode/org/hyperlinks.lua#L106 ?

kristijanhusak: There we don't need strict matching, just start needs to match.

jgollenz: @kristijanhusak pushed some changes, should be working now

kristijanhusak (approved):
Looks good, thanks!

Issue #249, opened 2022-04-08
Paragraph navigation differs in normal & visual modes


iexos:
### Describe the bug

I like using `{` and `}` to navigate from heading to heading, and it works well in normal mode. However, in visual mode it is using the regular vim behaviour and is thus inconsistent.

### Steps to reproduce

File:
```
* Heading 1
Text
* Heading 2
* Heading 3
Text

Text
* Heading 4

```
- From first line, press `}` in normal mode
- Go back to first line, press `v` and then `}`

### Expected behavior

Both times, the cursor should go to line 3. However, in visual mode it jumps to line 6.

### Minimal init.lua

https://gist.github.com/kristijanhusak/a0cb5f4eb2bad3e732a1d18d311ebe2f

### OS / Distro

Arch linux

### Neovim version/commit

0.6.1

jgollenz:
@kristijanhusak is this likely to be a bug inside nvim-orgmode or does this come from treesitter (or the grammar)?

kristijanhusak:
@jgollenz there is no mappings for visual mode, that's why it's not working.

jgollenz:
Ah, thanks :) I vaguely recall a conversation about visual mode mappings. Is that a `wontfix` or can I take a look at it?

kristijanhusak:
It's not a `wontfix`, you can look at it.

jgollenz:
Found the convo: #164 (at the bottom). Hence my question about wontfix. Do you have a general vision on how this should play out? Just so I know what to keep in mind ;)

kristijanhusak:
I think we could go with something like this. Taking this mapping from this issue as an example https://github.com/nvim-orgmode/orgmode/blob/master/lua/orgmode/config/mappings.lua#L70.
We could extend the right hand side to be a table with mode as key, and value as a path to a function that needs to be called (as it is working now). So for example this is how we would have normal, visual and operator mapping for this:
```lua
    org_previous_visible_heading = { 
      n = 'org_mappings.previous_visible_heading',
      x = 'org_mappings.previous_visible_heading_visual', -- different function if needed. If not, use the same value as for normal mapping
      o = {'org_mappings.previous_visible_heading_operator', true }, -- pass down the arguments to a function if needed.
      i = { callback = 'org_mappings.previous_visible_heading_insert', args = { true }, opts = { expr = true }, -- In case we need to provide something to the 4th argument of `vim.keymap.set`, like `expr` or `silent`
    },
[...]

PR #268, opened 2022-04-24
Enable mappings for different modes

jgollenz:
This PR adds support for mappings in other modes than normal-mode, as discussed in #249 .

@kristijanhusak fyi I'm still working on the part concerning your `vim.keymap.set` example ;)

### DIFF
--- a/lua/orgmode/config/init.lua
+++ b/lua/orgmode/config/init.lua
@@ -202,22 +202,56 @@ function Config:setup_mappings(category)
     return
   end
 
-  for name, key in pairs(self.opts.mappings[category]) do
+  for name, lhs in pairs(self.opts.mappings[category]) do
     if mappings[category] and mappings[category][name] then
-      local map = vim.tbl_map(function(i)
-        return string.format('"%s"', i)
-      end, mappings[category][name])
-      local keys = key
-      if type(keys) == 'string' then
-        keys = { keys }
+      local map = {}
+      if type(mappings[category][name]) == 'table' and not vim.tbl_islist(mappings[category][name]) then
+        -- multi-mode mapping
+        for mode, mapping in pairs(mappings[category][name]) do
+          map[mode] = {}
+          if type(mapping) == 'table' and not vim.tbl_islist(mapping) then
+            local action = {}
+            table.insert(action, mapping['callback'])
+            vim.list_extend(action, mapping['args'])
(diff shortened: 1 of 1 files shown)


lua/orgmode/config/init.lua
```diff
@@ -200,21 +200,32 @@ function Config:setup_mappings(category)
...
+  for name, lhs in pairs(self.opts.mappings[category]) do
     if mappings[category] and mappings[category][name] then
-      local map = vim.tbl_map(function(i)
-        return string.format('"%s"', i)
-      end, mappings[category][name])
-      local keys = key
-      if type(keys) == 'string' then
-        keys = { keys }
+      local map = {}
+      if #mappings[category][name] == 0 then -- we are in a multi-mode mapping
```
kristijanhusak, line 206: I think we should use this check:
```lua
type(mappings[category][name]) == 'table' and not vim.tbl_islist(mappings[category][name]
```
Because we have situation where we have a mapping that is a table list, but is actually a single normal mapping with an additional parameter, like [this one](https://github.com/nvim-orgmode/orgmode/blob/master/lua/orgmode/config/mappings.lua#L78)

jgollenz: If I'm not mistaken that example should work regardless, because it should still go into the else branch (since the length operator would return non-zero) :thinking:  But you are absolutely right, my check is ugly and I just didn't know how to do it properly :grin: Thanks a lot, will add the change :)

kristijanhusak (approved):
This looks good now. You will cover the other options also in this PR?

jgollenz:
> You will cover the other options also in this PR?

Yes, should find time for it this week

jgollenz:
```
some_multi_mapping = {
    x = 'org_mappings.previous_visible_heading_visual', -- different function if needed. If not, use the same value as for normal mapping
    o = {'org_mappings.previous_visible_heading_operator', true }, -- pass down the arguments to a function if needed.
}
```

@kristijanhusak we could make the first mapping equal to `x = { 'org_mappings.previous_visible_heading_visual' }`, demanding it to be a table. That would make it more consistent with the basic mapping and prune one additional way to do things.

kristijanhusak:
@jgollenz yeah, do that. More consistent is better.

jgollenz:
@kristijanhusak sorry, it took a while. Should be working now

kristijanhusak (approved):
Looks good, thanks!

Issue #271, opened 2022-04-25
Use treesitter for buffer local actions


TravonteD:
### Does this feature exist in Emacs orgmode core?

N/A

### Feature value

We can use tree-sitter to simplify some of the mappings that perform buffer modifications, and relieve the need for the buffer to be written to disk in order to perform them (currently if the file doesn't exist mappings don't work at all). I have a proof of concept of a couple of these actions [here](https://github.com/TravonteD/org-tree-actions) (changing priorities and todo_states). This would also reduce the amount of internal state that would be needed, since we could grab info about the buffer on the fly from the syntax tree. 

If this is something that we'd like to pursue I'd be glad to work on this further.

kristijanhusak:
> If this is something that we'd like to pursue I'd be glad to work on this further
I'm definitely in for simplifying things. I went through the repo you linked, but I'm having a bit hard time reading fennel.
Feel free to open up a draft PR and we can continue discussing there. Thanks.

PR #272, opened 2022-04-25
Use treesitter for buffer actions

TravonteD:
This PR is based on this discussion issue #271 
- [x] implement priority changing via treesitter
- [x] implement `TODO` keyword toggling via treesitter

### DIFF
--- a/lua/orgmode/org/mappings.lua
+++ b/lua/orgmode/org/mappings.lua
@@ -757,9 +757,10 @@ end
 ---@param use_fast_access boolean
 ---@return string
 function OrgMappings:_change_todo_state(direction, use_fast_access)
-  local item = Files.get_closest_headline()
-  local todo = item.todo_keyword
-  local todo_state = TodoState:new({ current_state = todo.value })
+  local headline = tree_utils.closest_headline()
+  local todo = tree_utils.get_todo(headline)
+  local current_keyword = todo and vim.treesitter.query.get_node_text(todo, 0) or ''
+  local todo_state = TodoState:new({ current_state = current_keyword })
   local next_state = nil
   if use_fast_access and todo_state:has_fast_access() then
     next_state = todo_state:open_fast_access()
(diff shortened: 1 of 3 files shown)


kristijanhusak:
This looks and works great!

This should simplify a lot of manual modification that I wrote initially before we had ts parser.

I would just like to have the code structured a bit differently. My idea is to have an encapsulated class for this. For example, instead of doing `tree_utils.set_todo(headline, next_state.value)` to have something like this:

``` lua
local headline = tree_utils.headline()
headline:set_todo(next_state.value)

-- Or

headline:set_priority(new_priority)
```

This is how it currently works with `Section`, but we could have something similar for these tree-sitter stuff, that doesn't require re-parsing file or anything, and maybe slowly migrating more things to this. What do you think?

TravonteD:
> My idea is to have an encapsulated class for this.

Converting it to class style should be straight forward, I'll give it an update.


I agree with your last suggestion, I'd like to keep implementing small ui changes that can be done with treesitter, and we can migrate when it makes since.

TravonteD:
I got a chance to look into the encapsulated style.
This:
```lua
tree_utils.set_todo(headline, next_state.value)
```
and this:
```lua
headline:set_todo(next_state.value)
```
are not equivalent because `tree_utils.get_closest_headline()` returns a `tsnode` (`:help lua-treesitter-node`). 
I'd prefer to have this not be something that holds state because the treesitter nodes already do that (via incremental parsing). After implementing a few functions I think that it'll make this much more complex to maintain.

kristijanhusak:
Yeah, it's an instance by itself, but I would like to have a small wrapper around it so it's more readable, and later more extensible. For example to introduce something like `Headline` "class":
```lua
---@param headline userdata tree sitter headline node
function Headline:new(headline)
  local data = { headline = headline }
  setmetatable(data, self)
  self.__index = self
  return data
end

-- Using get_stars for sake of simplicity
function Headline:get_stars()
  return self.headline:field('stars')[1]
end
```

And in your tree_utils you can just initiate it after you find it:

```lua
-- returns the Headline instance
function M.headline()
  return Headline:new(M.get_closest_headline())
end
```

And then in the mappings you could do this:
```lua
local headline = tree_utils.headline()
return headline:set_priority(value)
```

headline tsnode is basically a dependency for most of these ts util functions, so it's easier to provide it once and have it available inside then to pass it each time as a dependency to each function.

TravonteD:
I believe that I'm getting the gist of what you're saying, and I think it's just a matter of style preference. I mostly write in functional programming languages where encapsulation is achieved via different means. I'll finish up the implementation and set for review.

kristijanhusak (approved):
Looks good!

Thanks!

PR #280, opened 2022-05-05
Allow configurable org tags column

joaomsa:
Configuration setting mimics the emacs behavior, which allows aligning tags starting at a specific column or flush to a specific column.

The current default was 80 columns while the emacs default is -77 (for being nearly flush within 80 characters while still having extra room for the folded `...`).

[org-mode/org.el at main · bzg/org-mode](https://github.com/bzg/org-mode/blob/main/lisp/org.el#L2900)

I didn't want to break anybodies workflow but personally set it to -77. What do you think the default should be?


kristijanhusak:
I wasn't aware of this option. Do you know if it only applies to the org file itself or also for the agenda view?

joaomsa:
Slightly reworked it to use the helpers added in #272, so tag manipulation works in non-agenda files too

Grepping my way throughout the org-mode codebase I haven't found any use besides aligning tags on the actual orgfile
In emacs agenda view, tags are right-aligned with the right border of the buffer ignoring this setting.

kristijanhusak (approved):
Works great, thanks!

kristijanhusak:
> @kristijanhusak do you think the config should also apply to agenda or have it match the default emacs org-mode behavior?

Lets leave it as it is for now.

> In emacs agenda view, tags are right-aligned with the right border of the buffer ignoring this setting.

Oh, you're right. I didn't notice it because for me it always opens as a vertical split and I thought it has some specific width. Do you think doing the same thing would make sense here?

kristijanhusak:
I actually updated the agenda view to align tags to the right like emacs ormgode does. https://github.com/nvim-orgmode/orgmode/commit/2a5bdc73a67b60db42f021c40ae6e2152f93e4fb, and also an option to provide custom window command to have an option to open agenda in vertical split, or basically anything.

Thanks for letting me know about this.

PR #285, opened 2022-05-10
Fix for setting to DONE with no plan.

seibs:
I am seeing the following error when using `cit` to change a task to "DONE".

```
[orgmode] ...plugged/nvim-treesitter/lua/nvim-treesitter/ts_utils.lua:148: attempt to index local 'node' (a nil value)
```

Note, this change only fixes the immediate issue (the nil reference). I think there is still a disconnect with what is in the docs. If a schedule or deadline does not already exist on the headline, even when `org_log_done == 'time'`, a "CLOSED:" entry will not be added.

### DIFF
--- a/lua/orgmode/treesitter/headline.lua
+++ b/lua/orgmode/treesitter/headline.lua
@@ -82,6 +82,9 @@ end
 
 function Headline:dates()
   local plan = self:plan()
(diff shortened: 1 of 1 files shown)


seibs:
I added #286 to track the issue I noted above.

PR #293, opened 2022-05-11
Fix changing of keyword

TravonteD:
This addresses a scenario where the incorrect node is changed because it
has a matching keyword (see the regression test for an example)


kristijanhusak (approved):

kristijanhusak:
Thanks!

PR #312, opened 2022-05-28
Add explanation for `org_refile`

jgollenz:


Maltimore:
As per https://github.com/nvim-orgmode/orgmode/discussions/311, I think it should be included that one can only refile to files in `org_agenda_files`.
In the discussion linked above, you mentioned that you wanted to make a PR to include an error message when refiling goes wrong, I think that would be a great idea also! :)

jgollenz:
@Maltimore added the info about `org_agenda_files`. The PR about the warning is in the works ;)

PR #313, opened 2022-05-29
Fix wrong function access & output refile error

jgollenz:
fixes [the bug](https://github.com/nvim-orgmode/orgmode/discussions/311#discussioncomment-2839909) mentioned in #311

- refiling to a correct org-file with an incorrect headline resulted in an error, due to the function `_refile_to_end` being called with `.` instead of `:`
- display an error message when trying to refile to a file that is not part of the files set in `org_agenda_files`
- display an error message when trying to refile to a headline that is non-existent in the destination file

### DIFF
--- a/lua/orgmode/capture/init.lua
+++ b/lua/orgmode/capture/init.lua
@@ -147,9 +147,9 @@ end
 
 ---Triggered from org file when we want to refile headline
 function Capture:refile_headline_to_destination()
-  local agenda_file = Files.get_current_file()
-  local item = agenda_file:get_closest_headline()
-  local lines = agenda_file:get_headline_lines(item)
+  local destination_file = Files.get_current_file()
+  local item = destination_file:get_closest_headline()
+  local lines = destination_file:get_headline_lines(item)
   return self:_refile_content_with_fallback(lines, nil, item)
 end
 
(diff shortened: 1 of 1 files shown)


lua/orgmode/capture/init.lua
```diff
@@ -194,14 +194,19 @@ function Capture:_refile_content_with_fallback(lines, fallback_file, item)
   destination = vim.split(destination, '/', true)
 
   if not valid_destinations[destination[1]] then
+    if not default_file then -- we know that this comes from org_refile and not org_capture_refile
```
jgollenz, line 197: @kristijanhusak this is confusing imo. `default_file` is nil when this function is not triggered from a capture buffer. These semantics are quite implicit, perhaps we can rethink how this works?

lua/orgmode/capture/init.lua
```diff
@@ -216,7 +221,8 @@ function Capture:refile_to_headline(destination_file, lines, item, headline_titl
   end
 
   if not headline then
-    return self._refile_to_end(destination_file, lines, item)
+    utils.echo_info("headline '" .. headline_title .. "' does not exist in " .. destination_file .. ". Refiling to end of file instead.")
```
jgollenz, line 224: also, this is not displayed due to being overwritten by `Written to ...` message afterwards. But I do think it is valuable information in the case where users tried to specify a headline but made e.g. a typo. Perhaps we should even consider not fulfilling such a refiling action, as it probably is not what was intended by the user in the first place

lua/orgmode/capture/init.lua
```diff
@@ -216,7 +223,14 @@ function Capture:refile_to_headline(destination_file, lines, item, headline_titl
...
   if not headline then
-    return self._refile_to_end(destination_file, lines, item)
+    utils.echo_info(
+      "headline '"
+        .. headline_title
+        .. "' does not exist in "
+        .. destination_file
+        .. '. Refiling to end of file instead.'
+    )
+    return self:_refile_to_end(destination_file, lines, item)
```
jgollenz, line 233: this line was the actual culprit, fyi

lua/orgmode/capture/init.lua
kristijanhusak: It's nil in the situation when it's expected for a user to provide the destination file. I didn't handle the case where user didn't provide a valid destination. This is ok.

lua/orgmode/capture/init.lua
kristijanhusak: For now, I would show this message only when there is a `headline_title` provided and it wasn't found. This needs some refactoring to be more readable because there's a lot of confusing abstractions.

jgollenz: Added a check on `headline_title`. Still, this will never actually print, as it is overwritten by `Wrote <path/to/specified/file>`

jgollenz:
> This looks ok to me. Do you want to introduce any other changes or I can merge?

I'm still unsure what to do about that warning. As I said, it will never show, as it's immediately overwritten by the next log. Imo there should be no refiling when a user enters a non-existing headline. They clearly wanted to do something that cannot be fulfilled and our current output leaves them with the impression that their request was fulfilled, when in fact it was not. What is your opinion on this?

lua/orgmode/capture/init.lua
jgollenz: @kristijanhusak did some refactoring on the branching in this function. Let me know if its still unclear

jgollenz:
@kristijanhusak I made the attempt to refile to a non-existent heading in an existing file an error. The users intention cannot be full-filled, so we should inform them about it and abort the action

kristijanhusak (approved):
LGTM! Thanks

PR #319, opened 2022-06-04
Add handling of multi-key mappings

### LINKED ISSUE #318: Non-Global Multiple Mappings Don't Work
### Describe the bug

Following the [mapping instructions here](https://github.com/nvim-orgmode/orgmode/blob/master/DOCS.md#mappings), I've been unable to get non-global multiple mappings to work. So far, I've tried some agenda mappings and org mappings, for example:

```
mappings = {
  agenda = {
    org_agenda_quit = { 'q', '<ESC>' }
  }
```

Single mappings (not a table) work, eg. `org_agenda_quit = '<ESC>'`. But a table, as above, doesn't work for any of the specified keys.

Multiple mappings (with a table), do seem to work for global mappings, eg.:

```
global = {
  org_agenda = {'gA', '<Leader>oa'},
  org_capture = {'gC', '<Leader>oc'}
}
```

### Steps to reproduce

1. Specify a custom non-global multiple mapping (like agenda's `org_agenda_quit` or org's `org_cycle`.)
2. None of the keys specified in the multiple mapping will perform the mapped function.

### Expected behavior

When specifying a custom non-global multiple mapping, all specified keys will perform the mapped function.

### Emacs functionality

N/A

### Minimal init.lua

For this example, `q` or `<ESC>` should quit out of orgmode agenda (task view), but neither do.

```lua
vim.cmd [[set runtimepath=$VIMRUNTIME]]
vim.cmd [[set packpath=/tmp/nvim/site]]

local function load_plugins()
  require('packer').startup {
    {
      'wbthomason/packer.nvim',
      {'nvim-treesitter/nvim-treesitter'},
      {'kristijanhusak/orgmode.nvim', branch = 'master' },
    },
    config = {
      package_root = package_root,
    },
  }
end

_G.load_config = function()
    highlight = {
      enable = true,
    },
  })

  vim.cmd[[packadd nvim-treesitter]]
  vim.cmd[[runtime plugin/nvim-treesitter.lua]]
  vim.cmd[[TSUpdateSync org]]

  -- Close packer after install
  if vim.bo.filetype == 'packer' then
    vim.api.nvim_win_close(0, true)
  end

  require('orgmode').setup({
    mappings = {
      agenda = {
        org_agenda_quit = { 'q', '<ESC>' }
      }
    }
  })

  -- Reload current file if it's org file to reload tree-sitter
  if vim.bo.filetype == 'org' then
    vim.cmd[[edit!]]
  end
end

  vim.fn.system { 'git', 'clone', 'https://github.com/wbthomason/packer.nvim', install_path }
end
load_plugins()
require('packer').sync()
vim.cmd [[autocmd User PackerCompileDone ++once lua load_config()]]
```

_No response_

### OS / Distro

Pop!_OS 21.10, MacOS 12.3.1

### Neovim version/commit

0.7.0

### Additional context

_No response_

jgollenz:
fixes #318

@kristijanhusak I'm giving up on giving those variables a meaningful name. `map` contains the rhs + opts, any idea what it should be called?

### DIFF
--- a/lua/orgmode/config/init.lua
+++ b/lua/orgmode/config/init.lua
@@ -208,8 +208,24 @@ function Config:setup_mappings(category)
 
   for name, lhs in pairs(self.opts.mappings[category]) do
     if mappings[category] and mappings[category][name] and lhs then
+      -- lhs
+      local keys = {}
+      if type(lhs) == 'string' or vim.tbl_islist(lhs) then
+        -- no mode specified, default to normal-mode
+        keys['n'] = lhs
+      else
+        keys = lhs
+      end
+      -- rhs
       local map = {}
-      if type(mappings[category][name]) == 'table' and not vim.tbl_islist(mappings[category][name]) then
+      if not type(mappings[category][name]) == 'table' or vim.tbl_islist(mappings[category][name]) then
+        -- no mode specified, default to normal-mode
+        map['n'] = {}
+        map['n']['action'] = vim.tbl_map(function(i)
+          return string.format('"%s"', i)
+        end, mappings[category][name])
+        map['n']['opts'] = {}
+      else
         -- multi-mode mapping
         for mode, mapping in pairs(mappings[category][name]) do
           map[mode] = {}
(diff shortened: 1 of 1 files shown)


kristijanhusak:
I don't think we need to worry about it too much here, but maybe `rhs` could be used? I don't see it being used, but I'm on my phone so I might miss something

jgollenz:
yeah, it's just that `rhs` isn't accurate either, as it also contains opts :/ anyway, it's just bugging me, no big deal

kristijanhusak:
`rhs_and_opts` 😛 I think this is ok, thanks for fixing it!

PR #321, opened 2022-06-06
Make Help:show() more explicit

jgollenz:
I think this way the function is more consistent with the rest of the project.
@kristijanhusak I hope you don't mind


kristijanhusak:
Thanks!

PR #323, opened 2022-06-07
Fixed time stamp based sorting

gerazov:
For some reason time stamp based sorting was failing (some merge broke it?):

Screenshot_20220607_102617

I've fixed this adding a precedence sort if both items are on the same day and have a time stamp.


jgollenz:
Tested and confirmed working. Thanks @gerazov!

kristijanhusak:
I found some bug and fixed it here https://github.com/nvim-orgmode/orgmode/commit/a5fab63c43dd0dd78104acb2f54f7e1e80be013c, but I probably missed this scenario. Thanks!

PR #324, opened 2022-06-07
Resolve path for `default_notes_file` to support symlinks.

cheif:
This fixes an issue where you would end up with two references to the default file when using symlinks, since all other files had their symlinks resolved but not this.
This led to having one reference to the full path of the file, and one reference to the symlink, giving symptoms such as duplicate entries in the agenda.

This fix is very similar to [this](https://github.com/nvim-orgmode/orgmode/commit/60a101616bbf9b0dec11e44a46e8395509fc373e) one.

### DIFF
--- a/lua/orgmode/capture/init.lua
+++ b/lua/orgmode/capture/init.lua
@@ -134,7 +134,7 @@ end
 ---@private
 function Capture:_get_refile_vars()
   local template = vim.api.nvim_buf_get_var(0, 'org_template') or {}
-  local file = vim.fn.fnamemodify(template.target or config.org_default_notes_file, ':p')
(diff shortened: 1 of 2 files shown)


kristijanhusak:
Good catch. I just don't know if we need to do same thing here https://github.com/nvim-orgmode/orgmode/blob/master/lua/orgmode/capture/init.lua#L137

cheif:
> Good catch. I just don't know if we need to do same thing here https://github.com/nvim-orgmode/orgmode/blob/master/lua/orgmode/capture/init.lua#L137

Good catch to you as well sir, fixed now! (Although I don't think a missing `resolve` was as serious there)

kristijanhusak:
Thanks!

PR #330, opened 2022-06-10
Custom mapping prefix

### LINKED ISSUE #329: Provide a way to override the mapping prefix all at once

N/A

### Orgmode link

_No response_

### Feature value

nvim-orgmode uses `<Leader>o` for lots of its default mappings.

I see there's `disable_all` to disable it from doing so, and that I can explicitly remap all the keys in a similar way in its config, but I'd like to essentially *move* all the mappings to an unused key, rather than explicitly map them. Specifically I'd want all of orgmode's mappings to be on [`<localleader>`](https://neovim.io/doc/user/map.html#%3CLocalLeader%3E) which is more suitable for per-ft mappings.

Would it be reasonable to offer an additional config setting, such as:

```
mappings = {
    prefix = '<Leader>o',
}
```

allowing for overriding the above via `prefix = '<localleader>'`, without needing to repeat all the mappings?

Thanks for the plugin.

### Additional context

_No response_

jgollenz:
fixes #329

This kinda works. Unfortunately `<Leader>o` is hardcoded in some tests. I tried
replacing the mappings with e.g. `norm ,<prefix>k` but one error remains.


kristijanhusak:
> This kinda works. Unfortunately <Leader>o is hardcoded in some tests. I tried
replacing the mappings with e.g. norm ,<prefix>k but one error remains.

This is perfectly fine, no need to change tests. `<Leader>o` is the default, and I don't think that will ever change.

jgollenz:
> This is perfectly fine, no need to change tests. `<Leader>o` is the default, and I don't think that will ever change.

Yeah, unfortunately the tests fail. Any ideas?

kristijanhusak:
This is because some mappings are not going through this function https://github.com/nvim-orgmode/orgmode/blob/master/lua/orgmode/objects/edit_special/init.lua#L93-L110. 
Since you recently introduced ability to set custom mappings in the config, you can refactor this to:

* Have mappings as a table with the `callback` function that would call these functions, since they are custom.
* Call `config:setup_mappings('edit_src')` to properly bind mappings by the type in that file

jgollenz:
> Have mappings as a table with the callback function that would call these functions, since they are custom

I think this is not possible, because the custom function for the `org_edit_src_abort` mapping is using a variable that is only known inside `edit_special/init.lua`. The hack I made  works, but I can't say that I like it

Julian:
Thanks for jumping on this so quick! Seems this may already be nearly working, but just in case it's helpful, I *think* another way you might implement this is via a `:h <Plug>` key -- where you'd use `<Plug>(orgmode)` everywhere, and then map `<Plug>(orgmode)` to `<Leader>o` by default or otherwise to the configured value. (The above is slightly from memory, I haven't used `<Plug>` mappings from a plugin in a bit).

But if this is close to working which it sounds like it may be obviously ignore the above.

jgollenz:
@Julian that sounds like a good idea. I'll try if it would work. Thank you!

kristijanhusak:
> Thanks for jumping on this so quick! Seems this may already be nearly working, but just in case it's helpful, I think another way you might implement this is via a :h <Plug> key -- where you'd use <Plug>(orgmode) everywhere, and then map <Plug>(orgmode) to <Leader>o by default or otherwise to the configured value. (The above is slightly from memory, I haven't used <Plug> mappings from a plugin in a bit).

I don't think this can work, because `<Plug>` mapping can only be on the rhs, but here we are setting up lhs.

jgollenz:
I tried messing around with `<Plug>` a bit, but I can't get it to work. However, it's quite likely that I am not using it correctly. Documentation / examples are sparse. @Julian @kristijanhusak feel free to have a look at 363024f

kristijanhusak:
@jgollenz i don't think it can be done. Check my previous comment https://github.com/nvim-orgmode/orgmode/pull/330#issuecomment-1153095298

jgollenz:
> @jgollenz i don't think it can be done. Check my previous comment [#330 (comment)](https://github.com/nvim-orgmode/orgmode/pull/330#issuecomment-1153095298)

Pity, would have been nice. If you think my workaround in b0fc18a is valid, I'll clean up the PR for merging

lua/orgmode/objects/edit_special/init.lua
```diff
@@ -90,16 +90,18 @@ function EditSpecial:init()
 
   self:_set_context(bufnr, ctx)
 
+  --config:setup_mappings('edit_src')
+
   utils.buf_keymap(
     bufnr,
     'n',
-    config.mappings.edit_src.org_edit_src_abort,
+    config.mappings.edit_src.org_edit_src_abort:gsub('<prefix>', config.mappings.prefix),
```
kristijanhusak, line 98: Create a helper method in Config class for these `gsub('prefix')` changesand just call it here and in the setup_mappings.  Better to have it on single place

kristijanhusak (changes requested):

lua/orgmode/objects/edit_special/init.lua
jgollenz: done

jgollenz:
Seriously escalates #341. Put on hold until I figure that out

lua/orgmode/config/init.lua
```diff
@@ -34,6 +34,7 @@ function Config:extend(opts)
   opts = opts or {}
   self:_deprecation_notify(opts)
   self.opts = vim.tbl_deep_extend('force', self.opts, opts)
+  utils.keymap('n', '<Plug>(orgmode)', self.opts.mappings.prefix)
```
kristijanhusak, line 37: Do we need this for something? I don't think it's useful by itself.

jgollenz: whoops, that's a relic from earlier experimentation. Sorry

jgollenz:
from #344 

> We can just do the same gsub here https://github.com/nvim-orgmode/orgmode/pull/344/files#diff-4cdc8eef3d95a555d4cc3b73b3a23dafa438a257170a9efa3ace54a6dee6185fR91 before setting the mapping, that should do the trick.

The map entry needs the prefix to do gsub. Shall I pass that to every `attach` call or is it rather part of the opts of a map entry?

kristijanhusak:
Pass it to the attach through third `opts` parameter. I'm not 100% sure we will have proper prefix from user config in the "constructor".

lua/orgmode/config/mappings/map_entry.lua
```diff
@@ -85,7 +85,16 @@ function MapEntry:attach(default_mapping, user_mapping, opts)
 
   local map_opts = vim.tbl_extend('force', self.opts, opts or {})
 
+  local prefix = ''
+  if map_opts.prefix then
+    prefix = map_opts.prefix
+    map_opts.prefix = nil
```
jgollenz, line 91: When I pass it via `opts`, I need to remove the key or `vim.keymap` throws an error

kristijanhusak (approved):
Everything looks good to me. Once you think it's ready, move it to non-draft state and we can merge it.

jgollenz:
added a lonely unit test for good measure

kristijanhusak (approved):
Thanks!

PR #334, opened 2022-06-14
WIP: Feat/which key descriptions

### LINKED ISSUE #314: Add desc to orgmode keymap

N/A

### Orgmode link

_No response_

### Feature value

Which Key access the "desc" property to display set keymaps
it would be great to have proper descriptions for orgmode mappings
```lua
vim.keymap.set(mode, command, { desc = "Description of the mapping" })
```

### Additional context

_No response_

jgollenz:
closes #314

### TODO
- [x] multi-key rebinding
- [x] basic rebinding (user passes a string, not a table)
- [ ] discuss descriptions

### DIFF
--- a/lua/orgmode/agenda/views/tags.lua
+++ b/lua/orgmode/agenda/views/tags.lua
@@ -32,20 +32,15 @@ function AgendaTagsView:new(opts)
   return data
 end
 
-function AgendaTagsView:build(opts)
-  opts = opts or {}
-  local tags = opts.tags
-
-  if not tags then
-    tags = vim.fn.OrgmodeInput('Match: ', self.search, Files.autocomplete_tags)
-  end
+function AgendaTagsView:build()
+  local tags = vim.fn.OrgmodeInput('Match: ', self.search, Files.autocomplete_tags)
   if vim.trim(tags) == '' then
     return utils.echo_warning('Invalid tag.')
   end
   local search = Search:new(tags)
   self.items = {}
   for _, orgfile in ipairs(Files.all()) do
-    local headlines_filtered = orgfile:apply_search(search, opts.todo_only)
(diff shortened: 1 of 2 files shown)


jgollenz:
@kristijanhusak this should work. I need some confirmation on the descriptions I added

kristijanhusak:
Descriptions look fine, I just don't know if `which-key` gives some context around mappings? If a mapping has description `Earlier`, will user see something like `Orgmode - Earlier` or something like that?

jgollenz:
> I just don't know if which-key gives some context around mappings? If a mapping has 
> description Earlier, will user see something like Orgmode - Earlier or something like that?

Hm, it doesn't look like that is possible via `opts.desc` alone. It is possible to explicitly markup `o` (or whatever) with `orgmode`. Mappings with multiple keys for lhs, are organized in a tree structure:

which-key

If nothing is declared for `o` or `oi` it will be displayed as `+prefix`. So no implicit context, unless we specify it

kristijanhusak:
> So no implicit context, unless we specify it

Ok, lets add it. I think it will be more useful if users know the context, especially if they override some of the `<Leader>o` mappings.

kristijanhusak:
These mappings are really starting to complicate. Lets hold on this for a bit, I'll rewrite it today so it's more readable and testable. I'm worried we will break something with these changes.

jgollenz:
> These mappings are really starting to complicate

Absolutely. It's special cases galore. Didn't want to introduce any breaking changes, that's why I did it this way. Ideally, all the mappings have the same format. Right now there is ~5 different ways to do it

jgollenz:
And as can be seen with the help overlay, things are already starting to break

jgollenz:
new PR is #352

PR #343, opened 2022-06-18
Fix missing mappings in help overlay

jgollenz:
closes #341 

necessary for #334

PR #344, opened 2022-06-19
Refactor mappings

### LINKED ISSUE #341: Complex mappings break help overlay
### Describe the bug

Something about multi-mode mappings messes with how the help overlay works. Need to investigate further

### Steps to reproduce

1. open any `.org` file with `scripts/minimal_init.lua`
2. type `g?` to open the help overlay
3. the mappings for `Go to next heading (any level)` and `Go to previous heading (any level)` are empty

### Expected behavior

All mappings should be visible

### Emacs functionality

_No response_

### Minimal init.lua

default `minimal_init.lua`

image

### OS / Distro

Ubuntu 20.04

### Neovim version/commit

0.7.0

### Additional context

_No response_

kristijanhusak:
Should fix #341 and close https://github.com/nvim-orgmode/orgmode/pull/343


jgollenz (commented):
Looks good to me :) Good call on the MapEntry class :+1: concerning #334, the description wanders from `defaults.lua` to `mappings/init.lua`. If I'm not mistaken, that bars users from providing their own description in a custom mapping, correct? However, if you're using which-key, you probably have some custom descriptions anyway. Wouldn't consider that a big issue.

Concerning #330, do you have any plans on where and how to wedge that in?

kristijanhusak:
> If I'm not mistaken, that bars users from providing their own description in a custom mapping, correct?

Yeah that's right. Since I'm not a which-key user, can you tell me if other plugins who provide the description also allow overriding the descriptions with custom one? I'm trying to keep input from user (config) as simple as possible, and adding the option to provide custom description would require changing the data structure.

> Concerning https://github.com/nvim-orgmode/orgmode/pull/330, do you have any plans on where and how to wedge that in?

We can just do the same `gsub` here https://github.com/nvim-orgmode/orgmode/pull/344/files#diff-4cdc8eef3d95a555d4cc3b73b3a23dafa438a257170a9efa3ace54a6dee6185fR91 before setting the mapping, that should do the trick.

jgollenz:
> other plugins who provide the description also allow overriding the descriptions with custom one

thb, I don't know any plugins that use it. @danielnehrig do you know any?
Like I said, I don't think it's a huge deal. Concerning overriding: I think if a custom description is provided via which-key settings, then that gets top prio. If not, it falls back to `keymap.desc` and then `keymap.rhs`, but I did not test this yet

kristijanhusak:
Ok, you can proceed adding the descriptions with refactored code. If need arises, we will do some changes.

akinsho:
Just a drive by comment as I was looking for something like #334, a few other plugins use the `desc` key. `telescope` uses it for its `keybinds` picker, `legendary.nvim` also uses it for it's vscode like commands' picker as well as some others like maybe `keymenu.nvim`.

In the case of which key you can definitely override a `desc` key by specifying your own, telescope just reads whatever is in desc, but that's fine since that picker doesn't really attempt to do anything as fancy as which key. For which key you can also add your own prefix for the subtree label, but it seems quite longwinded for the user to do that but not the end of the world by any means. Alternatively you could wrap a call to which key 
e.g.
```lua
local function add_wk_label_if_exists(subtree_key, label)
  local ok, wk = pcall(require, 'which-key')
  wk.register({
    [subtree_key] = {name = label --[[ '+orgmode-thing' ]]},
  })
end
```

although very understandably you might not want this sort of code around 🤷🏿‍♂️

kristijanhusak:
> Just a drive by comment as I was looking for something like #334, a few other plugins use the `desc` key. `telescope` uses it for its `keybinds` picker, `legendary.nvim` also uses it for it's vscode like commands' picker as well as some others like maybe `keymenu.nvim`.
> 
> In the case of which key you can definitely override a `desc` key by specifying your own, telescope just reads whatever is in desc, but that's fine since that picker doesn't really attempt to do anything as fancy as which key. For which key you can also add your own prefix for the subtree label, but it seems quite longwinded for the user to do that but not the end of the world by any means. Alternatively you could wrap a call to which key e.g.
> 
> ```lua
> local function add_wk_label_if_exists(subtree_key, label)
>   local ok, wk = pcall(require, 'which-key')
>   wk.register({
>     [subtree_key] = {name = label --[[ '+orgmode-thing' ]]},
>   })
> end
> ```
> 
> although very understandably you might not want this sort of code around 🤷🏿‍♂️

Thanks for the detailed explanation @akinsho , this is very helpful.
@jgollenz let's go with the `desc` on each mapping as you initially added it, and if user wants to specifically override it, they can do so from the `which-key`

PR #345, opened 2022-06-19
Fix overriding of global fillchars

0xAdk:
currently the plugin will override the user's `fillchars` when setting `fold:c`, this unintentionally set's everything else beck to their defaults (such as `horiz:c` or `vert:c` for example)


kristijanhusak:
Good catch, thanks!

PR #347, opened 2022-06-21
Fixed no args change_span actions erroring out with table index is nil

gerazov:
Due to the change in `map_entry.lua` the options to `action` now need to be passed via a table with `args`. This updates the `change_span` actions for day, week, month and year to output a correct action call. Otherwise they were erroring out with a nil value for the span:

```
[orgmode] ....vim/plugged/orgmode/lua/orgmode/agenda/views/agenda.lua:115: table index is nil
```


kristijanhusak:
I was just working on a fix, thanks!

PR #354, opened 2022-06-26
Enable archiving from agenda

### LINKED ISSUE #339: Archiving from agenda

No

### Orgmode link

_No response_

### Feature value

The `<leader>o$` mapping for archiving headline(-tree)s in normal org-files would also be applicable to the agenda in my opinion.

### Additional context

_No response_

jgollenz:
closes #339

Needs some more testing. Looks to me like it may archive more than intended

- [x] docs
- [ ] unit tests

### DIFF
--- a/lua/orgmode/agenda/init.lua
+++ b/lua/orgmode/agenda/init.lua
@@ -303,6 +303,13 @@ function Agenda:priority_down()
   })
 end
 
+function Agenda:archive()
+  return self:_remote_edit({
+    action = 'org_mappings.archive',
+    redo = true,
+  })
+end
+
 function Agenda:toggle_archive_tag()
   return self:_remote_edit({
     action = 'org_mappings.toggle_archive_tag',
--- a/DOCS.md
+++ b/DOCS.md
@@ -610,6 +610,9 @@ Increase the priority of a headline item.
 #### **org_agenda_priority_down**
 *mapped to*: `-`<br />
 Decrease the priority of a headline item.
+#### **org_agenda_archive**
+mapped to: `<Leader>o$`<br />
+Archive headline item to archive location.
 #### **org_agenda_toggle_archive_tag**
 *mapped to*: `<Leader>oA`<br />
 Toggle "ARCHIVE" tag of a headline item.
--- a/lua/orgmode/config/defaults.lua
+++ b/lua/orgmode/config/defaults.lua
@@ -79,6 +79,7 @@ return {
       org_agenda_priority = '<prefix>,',
       org_agenda_priority_up = '+',
       org_agenda_priority_down = '-',
+      org_agenda_archive = '<prefix>$',
       org_agenda_toggle_archive_tag = '<prefix>A',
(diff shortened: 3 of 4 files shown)


lua/orgmode/agenda/init.lua
```diff
@@ -303,6 +303,13 @@ function Agenda:priority_down()
   })
 end
 
+function Agenda:archive_subtree()
+  return self:_remote_edit({
+    action = 'org_mappings.archive',
+    update_in_place = true,
```
kristijanhusak, line 309:
```suggestion
    redo = true,
```

kristijanhusak (changes requested):
> Looks to me like it may archive more than intended

It doesn't. It just loads the wrong state once the headline is gone. Applying the suggested change will just reload the agenda view and it should work fine.

jgollenz:
> it should work fine
It only does when the file where the headline to be archived is located is opened. Otherwise it will e.g. archive the whole top-level parent tree of a level three headline. Need to investigate

jgollenz:
[This line](https://github.com/nvim-orgmode/orgmode/blob/07d0e00baedcbe8b4f2175f8d9fbefac2a2d38b5/lua/orgmode/org/mappings.lua#L39) returns an empty string and consequently messes with the call to `refile_file_headline_to_archive`, I think

kristijanhusak:
> It only does when the file where the headline to be archived is located is opened. Otherwise it will e.g. archive the whole top-level parent tree of a level three headline

I can't reproduce this. Once you have reproducible steps let me know so I can take a look

jgollenz:
1. create a test file: 
```lua
* foo
** bar
*** heading to archive
    <2022-06-30>
```
2. nvim -u [minimal_init.lua](https://gist.github.com/jgollenz/27cebddeb8834fe5f08ab44b3ade3bb8)
3. `<leader>oa`     -- leader is mapped to <space>
4. navigate to `heading to archive`
5. `<leader>o$`

Archiving the headline works, but at this point the whole parent tree of the headline in question is removed from the test file (leaving it empty in this case, but not necessarily). So far this always went up to the top-level parent heading

jgollenz:
@kristijanhusak maybe I'm overlooking something, but we don't have unit tests for archiving yet, correct? Should we add some?

kristijanhusak:
We don't have any. I'm always in for more tests. It was a bit tricky to add them, but if you figure out a way to do some testing around it, please do so.

jgollenz:
Alright, then let's open an issue for that and handle it separately. Not sure when I get around to it.

kristijanhusak:
Let's use `org_agenda_archive` instead of `org_agenda_archive_subtree`, to follow Emacs convention https://orgmode.org/manual/Agenda-Commands.html#:~:text=short%20%24%20(-,org%2Dagenda%2Darchive,-)

jgollenz:
> Let's use `org_agenda_archive` instead of `org_agenda_archive_subtree`

done

kristijanhusak (approved):
Looks good, thanks!

PR #361, opened 2022-07-03
Enable link opening from agenda

jgollenz:
The `Agenda:open_link` function is almost identical to `OrgMappings:open_at_point`. Any suggestions on how to de-duplicate it are appreciated. The moving of `get_link_under_cursor` to `Hyperlinks.lua` is a proposal.


kristijanhusak:
Emacs orgmode functionality is actually a bit more complicated than this:
> C-c C-o (org-agenda-open-link)
Follow a link in the entry. This offers a selection of any links in the text belonging to the referenced Org node. If there is only one link, follow it without a selection prompt.

This means that it will find all the links under the headline, and show you something like `inputlist()` to select which one you want to open, or just default to first one if it's the only one.
This can be maybe done differently:
* Find the headline from the org state
* Add a method to it called `getHyperlinks()`
* Show the list or go to first if single

Regarding finding hyperlinks, we could try using a tree-sitter query with custom predicate which I'm using for highlighting (https://github.com/nvim-orgmode/orgmode/blob/master/queries/org/markup.scm#L14). I'm not sure if that would work though, would need to give it a test.

jgollenz:
Some questions:

1. If I recall correctly I was concerned about links that are part of the headline. Emacs also allows you to open headlines with links in the body, correct?
2. Does Emacs go down recursively into child headings or is it just the body of this one heading?

kristijanhusak:
1. Yes
2. I haven't tested this scenario. I can get back to you on this tomorrow

kristijanhusak:
> I haven't tested this scenario. I can get back to you on this tomorrow
It doesn't go recursive. It takes only that specific headline, at least by default. I don't know if there's some specific setting for it, I didn't find any.

PR #373, opened 2022-07-24
Fix wrong tag_column indentation

jgollenz:
The `tags_width` was being subtracted in any case, which resulted in flushright alignment per default. When a negative `tag_column` value was given, the `tags_width` was subtracted twice, resulting in a broken alignment.

### DIFF
--- a/lua/orgmode/config/defaults.lua
+++ b/lua/orgmode/config/defaults.lua
@@ -21,7 +21,7 @@ return {
   org_priority_default = 'B',
   org_priority_lowest = 'C',
   org_archive_location = '%s_archive::',
-  org_tags_column = 80,
+  org_tags_column = -80,
   org_use_tag_inheritance = true,
   org_tags_exclude_from_inheritance = {},
   org_hide_leading_stars = false,
--- a/lua/orgmode/treesitter/headline.lua
+++ b/lua/orgmode/treesitter/headline.lua
@@ -68,7 +68,7 @@ function Headline:set_tags(tags)
       to_col = math.abs(to_col) - tags_width
     end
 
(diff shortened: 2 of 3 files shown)


kristijanhusak:
`tag_column` is not meant to be negative. Flushright alignment is how Emacs aligns it. Initial implementation was wrong, but I fixed it some time ago. Agenda is also aligned that way (kind of).

jgollenz:
If it's not meant to be negative, then 1) the [`DOCS.md`](https://github.com/nvim-orgmode/orgmode/blob/master/DOCS.md#org_tags_column) is wrong and 2) there is no real reason for [this subtraction](https://github.com/nvim-orgmode/orgmode/blob/8cc6fa4599aeae171d3051570a10c94269acf05f/lua/orgmode/treesitter/headline.lua#L68), or is there? Actually, I think there should be no `math.abs` either in this case, just a default value.

jgollenz:
My emacs says this: 

```
org-tags-column is a variable defined in ‘org.el’.
Its value is -77

Documentation:
The column to which tags should be indented in a headline.
If this number is positive, it specifies the column.  If it is negative,
it means that the tags should be flushright to that column.  For example,
-80 works well for a normal 80 character screen.
When 0, place tags directly after headline text, with only one space in
between.
```

kristijanhusak:
You are completely right, I misunderstood it. I thought it always behaves as flushright.
I think what we need to do here is to change the default for `org_tags_column` to `-80` instead of `80`, and to apply your fix. Test with `:ARCHIVE:` tag shouldn't need to be changed since it should pass with your fix and the default value change.

jgollenz:
@kristijanhusak I applied the changes you suggested. Indeed, the change in the unit test was not necessary :+1:

kristijanhusak (approved):
Looks good, thanks!

PR #404, opened 2022-08-29
fix(capture): Do not yank cursor placeholder

anton-fomin:
When capturing an item with cursor placeholder `%?` the placeholder
itself is getting yanked.


kristijanhusak:
Thanks!

Issue #366, opened 2022-07-15
Cant open file by external link with `<leader>oo`


VolkovIlia:
### Describe the bug

Got error when trying to open files by `<leader>oo`

`.../packer/start/orgmode.nvim/lua/orgmode/org/mappings.lua:869: attempt to index local 'item' (a nil value)`

### Steps to reproduce

1. Enter external link to file `[[todo.org]]`
2. Try to open it with `<leader>oo`

### Expected behavior

Open file in new tab.

### Minimal init.lua

Standard

### Screenshots and recordings

![image](https://user-images.githubusercontent.com/43730119/179225970-0b470bda-63f4-4bc2-a264-5ce2560c5f08.png)

### OS / Distro

ArtixLinux Kernel: 5.18.8-artix1-1

### Neovim version/commit

v0.7.2

jgollenz:
I think the issue is due to you not having any org headline at all in that file. When I change line 11 to `* TODO [[todo.org]]` it works for me. I am unsure whether a file without any root headline is considered a valid org-file in emacs. @kristijanhusak?

jgollenz:
@kristijanhusak I think this can be closed

kristijanhusak:
> I am unsure whether a file without any root headline is considered a valid org-file in emacs

@jgollenz  did you check this? I think this should work even without headline, but I didn't check how emacs works.

jgollenz:
Oh well, looks like Emacs has no trouble with that. A file just containing `[[file:~/path/to/file.org]]` will open that file in a new buffer upon `C-c C-o`.

jgollenz:
Perhaps the buffer itself may be considered the "root heading", with depth -1.

zdcthomas:
I'm also hitting the same issue with any relative path link

jgollenz:
Yes, this will happen for all links. It is because the logic will first check whether the cursor is on a date and that function tries to access the closest headline, which does not exist. Another interesting implication of that: 

![link_priorities](https://user-images.githubusercontent.com/12050098/188423159-1d13257b-8065-4714-842a-0c74d23e7ad2.gif)

A link name that looks like a date will be interpreted as a date. I guess we should first check whether it the cursor is inside a link.

zdcthomas:
[...]

PR #409, opened 2022-09-12
Look for link before looking for date

jgollenz:
Addresses [this](https://github.com/nvim-orgmode/orgmode/issues/366#issuecomment-1236820452)

### DIFF
--- a/lua/orgmode/org/mappings.lua
+++ b/lua/orgmode/org/mappings.lua
@@ -644,13 +644,12 @@ function OrgMappings:_edit_special_callback()
 end
 
 function OrgMappings:open_at_point()
-  local date = self:_get_date_under_cursor()
-  if date then
-    return self.agenda:open_day(date)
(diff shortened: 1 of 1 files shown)


kristijanhusak:
This should already be solved with this commit https://github.com/nvim-orgmode/orgmode/commit/1f103a9b5aa15e420d249afe1a3e086a61570563. Is there any other issue that I'm not aware of?

jgollenz:
Sorry, I should have been more clear. The issue is that if a link has a name which is a valid timestamp 

`[[example.org][<2022-10-31]]`

When on the URL, `<Leader>oo` opens the file/website. When on the name/timestamp, it will open the agenda. I am under the assumption that a link-name is not a valid timestamp, but I might be wrong.

kristijanhusak:
I tested it in Emacs, and it indeed treats it as link. Thanks!

PR #410, opened 2022-09-13
Aligning tags

### LINKED ISSUE #389: Aligning tags

N/A

### Orgmode link

Is there any function that can be called to realign tags to the defined margin? When manually adding tags by just typing the keywords `:TAG1:TAG2:` or the like it's cumbersome to align them with the other tags. Would be great if there would be an easy way to realign tags in a document or marked area.

### Feature value

Ease of use and better visuals

### Additional context

_No response_

jgollenz:
fixes #389 

- [x] align after promotion, demotion and TODO state change
- [ ] add mapping to align all tags in file
- [ ] unit tests

### DIFF
--- a/lua/orgmode/org/format.lua
+++ b/lua/orgmode/org/format.lua
@@ -1,4 +1,5 @@
 local Files = require('orgmode.parser.files')
+local ts_org = require('orgmode.treesitter')
 local Table = require('orgmode.treesitter.table')
 
 local function format()
@@ -18,6 +19,11 @@ local function format()
   local ok, item = pcall(Files.get_closest_headline, line)
   if ok and item and item.logbook and item.logbook.range:is_in_line_range(line) then
     return item.logbook:recalculate_estimate(line)
+  end
+
+  if vim.fn.getline(line):match('^%*') then
+    ts_org.closest_headline():align_tags()
+    return 0
   end
 
   return 1
--- a/lua/orgmode/org/format.lua
+++ b/lua/orgmode/org/format.lua
@@ -1,6 +1,7 @@
 local Files = require('orgmode.parser.files')
 local ts_org = require('orgmode.treesitter')
 local Table = require('orgmode.treesitter.table')
+local util = require('orgmode.utils')
 
 local function format()
   if vim.tbl_contains({ 'i', 'R', 'ic', 'ix' }, vim.fn.mode()) then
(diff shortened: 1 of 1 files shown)


kristijanhusak:
@jgollenz you can either rebase and continue working on this to add alignments where you think it's missing by setting up an event, or you can close it and create a new PR if needed.

jgollenz:
I'll rebase it. The only thing missing is the mapping that aligns all the tags in the buffer. I'm trying to do this with a treesitter query (`util.get_ts_matches('(headline (tag_list))'`), but I'm still struggling with that

kristijanhusak:
What mapping have you planned to use? Since this is basically a formatting action, I think we should use `formatexpr` that is bind to `gq`. There's already a [formatexpr handler](https://github.com/nvim-orgmode/orgmode/blob/master/lua/orgmode/org/format.lua) for tables and clock times, so I think for now you can just add it in here by checking if current line starts with `*`, and if yes, call the `ts_org.closest_headline():align_tags()`. You can add it before last return, and return 0 if formatting was done. Doing it through global query is probably more difficult, and logic would be potentially split in 2 places.

jgollenz:
Ah, this makes sense, thanks! Emacs does it for the whole buffer, what would be the equivalent for that? `gq%`?

kristijanhusak:
`gq` behaves like `=`, which means that it's probably `gggqG`. I don't know if there's a shorter version.

jgollenz:
I added 

```
  if vim.fn.getline(line):match('^%*') then
    ts_org.closest_headline():align_tags()
    return 0
  end
```

and it works fine if I just select a single line in visual mode. But it breaks gloriously if I select multiple lines, especially if there are empty lines before the first heading with tags.

kristijanhusak:
`formatexpr` is not working as I expected when I initially added it. While testing this I found few bugs and rewrote the `formatexpr` to work. I also added a support for aligning tags, so this PR became obsolete, sorry. You can pull latest master and give it atry.

jgollenz:
huh, interesting. `vim.v.count` always returns zero for me :man_shrugging: anyway, glad its finally working :grinning:

kristijanhusak:
That's because there was a bug in `Table.format` which messed up the line numbering for some reason (`norm! _` caused the issue).

jgollenz:
Ah, I see. We should document this feature

kristijanhusak:
We have few mentions of `gq` in the docs for tables and clock values, but I guess we should unify it in one section.

kristijanhusak:
I added some basic explanation here https://github.com/nvim-orgmode/orgmode/blob/master/DOCS.md#formatting, should be ok for now.

PR #411, opened 2022-09-14
Add event dispatcher

kristijanhusak:
@jgollenz this is the result of https://github.com/nvim-orgmode/orgmode/pull/410/files. I thought about it a little bit, and figured out we might need to do more things when something changes, not only this one. That's why I introduced an event dispatcher and few events that are fired, and tags are aligned through them.

This will also allow us to expose listener through the api, and allow people to hook into it if they want to do something.
Let me know what you think.

### DIFF
--- a/lua/orgmode/events/init.lua
+++ b/lua/orgmode/events/init.lua
@@ -1,0 +1,44 @@
+local Events = require('orgmode.events.types')
+local Listeners = require('orgmode.events.listeners')
+
+---@class EventManager
+local EventManager = {
+  initialized = false,
+  _listeners = {},
+  event = Events,
+}
+
+---@param event Event
+function EventManager.dispatch(event)
+  if EventManager._listeners[event.type] then
+    for _, listener in ipairs(EventManager._listeners[event.type]) do
+      listener(event)
+    end
+  end
+end
+
+---@param event Event
+---@param listener fun(...)
+function EventManager.listen(event, listener)
+  if not EventManager._listeners[event.type] then
+    EventManager._listeners[event.type] = {}
+  end
+  if not vim.tbl_contains(EventManager._listeners[event.type], listener) then
+    table.insert(EventManager._listeners[event.type], listener)
+  end
+end
+
+function EventManager.init()
+  if EventManager.initialized then
+    return
+  end
+  for event, listeners in pairs(Listeners) do
+    for _, listener in ipairs(listeners) do
+      EventManager.listen(event, listener)
+    end
+  end
+  EventManager.initialized = true
+  return EventManager
+end
+
+return EventManager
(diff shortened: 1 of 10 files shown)


jgollenz (commented):
I think its a good idea :+1: code looks fine to me, the only issue I have (which is one of personal taste) is that I would rather declare a `local event = ...` and then call `eventmanager.dispatch(event)` instead of `dispatchEvent()`. Just a nitpick really.

One thing that got me thinking is how to decide when something is best run done with events? My reasoning was this: the primary intention of a user was to do promotion or whatever. The fact that alignment should also happen is secondary. It is not the actual intention. These secondary considerations are a good fit for events. For the alignment of all tags, however, there should be a dedicated mapping, thus it becomes the users primary intent (so non-event based). Do you agree with that sort of reasoning? I suspect that we might have similar situations throughout the codebase, that currently are not handled with events but really should be. Any thoughts on how to tackle that? Going through all of it sounds a bit daunting. I guess a case-by-case evaluation whenever we encounter such a situation is adequate

kristijanhusak:
> I have (which is one of personal taste) is that I would rather declare a local event = ... and then call eventmanager.dispatch(event) instead of dispatchEvent(). Just a nitpick really.

I did this in that one case only to make sure that `item` is returned as before (I don't think it matters for anything though), and we always get the "fresh" headline node and section instance as an argument to the event. I would need to type that on each return, so I just reduced duplication with it. Regarding the event in variable, it might be more readable since the line is long.

> One thing that got me thinking is how to decide when something is best run done with events? My reasoning was this: the primary intention of a user was to do promotion or whatever. The fact that alignment should also happen is secondary. It is not the actual intention. These secondary considerations are a good fit for events. 

I completely agree with this. That's the main purpose of it, to handle any kind of side effects. I will also expose it in some way through the api, so plugin developers or any user can hook into them if they want.

> For the alignment of all tags, however, there should be a dedicated mapping, thus it becomes the users primary intent (so non-event based)

Yes, if that's the main intent of the mapping, it should not be an event. Events should be only for side effects.

> I suspect that we might have similar situations throughout the codebase, that currently are not handled with events but really should be. Any thoughts on how to tackle that? Going through all of it sounds a bit daunting. I guess a case-by-case evaluation whenever we encounter such a situation is adequate

I plan to spend some time on it just to quickly scan through code to see where it would be applicable, but I will not go into the details. If I catch something, I'll extract it. If you find something, feel free to suggest it via issue or a draft PR.

jgollenz:
> I plan to spend some time on it just to quickly scan through code to see where it would be applicable, but I will not go into the details. If I catch something, I'll extract it. If you find something, feel free to suggest it via issue or a draft PR.

Agreed, sounds like a plan :+1:

PR #418, opened 2022-09-26
Fix broken priority cycling

### LINKED ISSUE #382: Priority Cycling broken when prio-state above 9
### Describe the bug

When due to custom values for `org_priority_[highest|lowest]` a prio-state above 9 becomes a valid state, the cycling is broken. See the gif.

### Steps to reproduce

- open a buffer with the provided `minimal_lua.init`
- enter `cir` on a headline until you reach a prio-state above 9 

### Expected behavior

According to [this](https://orgmode.org/manual/Priorities.html) prio-states up to 64 are valid. I highly suspect this is because 65 is the ASCII code for `A`. If we want to replicate this, we should add it to the docs.

### Emacs functionality

~~It appears, that vanilla orgmode is also buggy here, but the behavior is different. Try setting `#+PRIORITIES: 1 35 1` at the beginning of a buffer. For me, the cycling only ranges between 1 and 3. Looks like the 35 is parsed as 3. If it works for you, please let me know!~~

Edit: I had an outdated version of orgmode in my emacs. It works as expected.

### Minimal init.lua

```lua
vim.cmd([[set runtimepath=$VIMRUNTIME]])
vim.cmd([[set packpath=/tmp/nvim/site]])

local function load_plugins()
  require('packer').startup({
    {
      'wbthomason/packer.nvim',
      { 'nvim-treesitter/nvim-treesitter' },
      { 'kristijanhusak/orgmode.nvim', branch = 'master' },
    },
    config = {
      package_root = package_root,
    },
  })
end

_G.load_config = function()
    highlight = {
      enable = true,
    },
  })
  require('orgmode').setup({
    org_priority_highest = '1',
    org_priority_default = '5',
    org_priority_lowest = '15'
  })

  vim.cmd([[packadd nvim-treesitter]])
  vim.cmd([[runtime plugin/nvim-treesitter.lua]])
  vim.cmd([[TSUpdateSync org]])

  -- Close packer after install
  if vim.bo.filetype == 'packer' then
    vim.api.nvim_win_close(0, true)
  end

  require('orgmode').setup()

  -- Reload current file if it's org file to reload tree-sitter
  if vim.bo.filetype == 'org' then
    vim.cmd([[edit!]])
  end
end

  vim.fn.system({ 'git', 'clone', 'https://github.com/wbthomason/packer.nvim', install_path })
  load_plugins()
  require('packer').sync()
  vim.cmd([[autocmd User PackerCompileDone ++once lua load_config()]])
else
  load_plugins()
  load_config()
end
```

prio-cycle

The bug is even more peculiar when doing reverse cycling with `ciR`:

reverse-prio-cycle

### OS / Distro

Ubuntu 20.04

### Neovim version/commit

0.7.0

### Additional context

_No response_

jgollenz:
fixes #382 

@kristijanhusak I'd like to hear your opinion on the validation of priorities provided by the users config. It's a bit messy I think.

### DIFF
--- a/lua/orgmode/config/init.lua
+++ b/lua/orgmode/config/init.lua
@@ -34,8 +34,93 @@ function Config:extend(opts)
   self.todo_keywords = nil
   opts = opts or {}
   self:_deprecation_notify(opts)
+  if not self:_are_priorities_valid(opts) then
+    opts.org_priority_highest = self.opts.org_priority_highest
+    opts.org_priority_lowest = self.opts.org_priority_lowest
+    opts.org_priority_default = self.opts.org_priority_default
+  end
   self.opts = vim.tbl_deep_extend('force', self.opts, opts)
   return self
+end
+
+function Config:_are_priorities_valid(opts)
+  local high = opts.org_priority_highest
+  local low = opts.org_priority_lowest
+  local default = opts.org_priority_default
+
+  if high or low or default then
+    -- assert that all three options are set
+    if not (high and low and default) then
+      utils.echo_warning(
+        'org_priority_highest, org_priority_lowest and org_priority_default can only be set together.'
+          .. 'Falling back to default priorities'
+      )
+      return false
+    end
+
+    -- numbers
+    if type(high) == 'number' and type(low) == 'number' and type(default) == 'number' then
+      if high < 0 or low < 0 or default < 0 then
+        utils.echo_warning(
+          'org_priority_highest, org_priority_lowest and org_priority_default cannot be negative.'
+            .. 'Falling back to default priorities'
+        )
+        return false
+      end
+      if high > low then
+        utils.echo_warning(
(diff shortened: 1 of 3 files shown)


jgollenz:
Regarding the failed tests: if I read [this](https://orgmode.org/manual/Priorities.html) correctly, then [this](https://github.com/nvim-orgmode/orgmode/blob/77ea13332f3c811b707e902c7cb8605c471a2f4c/tests/plenary/object/priority_state_spec.lua#L15) is not correct. The highest priority should be the lowest number.

jgollenz:
See, the input validation already paid off. Those unit tests were only working due to the bug I'm trying to fix :grin:

jgollenz:
@kristijanhusak this should now work

kristijanhusak (approved):

kristijanhusak:
Thanks!

PR #419, opened 2022-09-27
Enable expansions within lua expression

jgollenz:
It says [here](https://orgmode.org/manual/Template-expansion.html) that `[the expansion of lisp expressions] happens after expanding non-interactive “%-escapes”, those can be used to fill the expression`. Fine, but this messes with Lua expressions, since its escape sequences overlap with those of orgmode templates. 

So, I guess we could do some aliasing for either one of the two sets, when they are within a `%(...)` Lua expression: 

- `%(return string.format("Hello, %s", @x)` aliasing template escapes
- `%(return string.format("Hello, @s", %x)` aliasing lua escapes

The `@` is just for demonstrative purposes, we could use else entirely.

This is the use case that actually prompted me to investigate that: `template = "* TODO [[%x][%(return string.match('%x', '([^/]+)$'))]]"`. I use it when I come across an interesting repo. `C-l`, `C-c`, switch to neovim and start the capture. Before that I had the cursor where the link name goes and typed the repo name in by hand.

### DIFF
--- a/lua/orgmode/capture/templates.lua
+++ b/lua/orgmode/capture/templates.lua
@@ -56,13 +56,9 @@ function Templates:compile(template)
     content = table.concat(content, '\n')
   end
   content = self:_compile_dates(content)
-  content = self:_compile_prompts(content)
+  content = self:_compile_expansions(content)
   content = self:_compile_expressions(content)
-  for expansion, compiler in pairs(expansions) do
-    if content:match(vim.pesc(expansion)) then
-      content = content:gsub(vim.pesc(expansion), compiler())
-    end
-  end
+  content = self:_compile_prompts(content)
   return vim.split(content, '\n', true)
 end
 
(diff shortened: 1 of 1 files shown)


kristijanhusak:
> So, I guess we could do some aliasing for either one of the two sets, when they are within a %(...) Lua expression

Isn't this change actually fixing the issue?  Expansions populate the expression, and expression gets evaluated. We could maybe have issue with some of the `string.format` placeholders, but most common ones (`%d`, `%s`, `%f`) aren't actually "mapped" to anything in the Orgmode itself, according to the docs you linked.

jgollenz:
>  Expansions populate the expression, and expression gets evaluated

That is the issue and it gets introduced by this MR. Before the change, the Lua expression would be evaluated before the other expansions. However, that is not how it's done in emacs orgmode. Yes, the overlap is small, but it exists. Actually `%f` is mapped to the file from which the capture was started. Personally I think that the benefit of being able to populate the expression outweighs the string formatting, but I don't really want to cause a functionality regression.

kristijanhusak:
We could probably do some regex matching to make sure string.format values are not replaced, but that would be pretty nasty.

I think a potential workaround could be to define function in another place, and have the expression evaluate it. I just don't know if that would actually work until I try it:

Example lua:
```lua
--- lua/myconfig/org_functions.lua
local M = {}
function M.my_string_format_fn(value)
  return string.format('Hello %s', value)
end

return M
```

Calling the function from expression
```
%(return require('myconfig.org_functions').my_string_format_fn('%x'))
```

kristijanhusak:
> return string.match('%x', '([^/]+)$'))

Using external function works perfectly fine for me, so I think this PR is good to go.

kristijanhusak (approved):

PR #421, opened 2022-10-03
Advance agenda span with count

jgollenz:
It says [here](https://orgmode.org/manual/Agenda-Commands.html): `Go forward in time to display the span following the current one. [...] With a prefix argument, repeat that many times.`


kristijanhusak:
Thanks!

PR #426, opened 2022-10-17
Inline math highligthing

gzagatti:
This PR adds support to in-line math highlighting, it builds on previous work done for highlighting inline markups.

In following text:

```org
* Circle

Let a circle with radius \( r \), then the perimeter is given by \(2\pir\) and the area by \( π * r^2 \). 

Some examples:

One more thing \(\).
```

All of the in-line math should be highlighted with `OrgTSLatex`.

As I write a lot of technical texts, this features is very important. Feel free to provide some feedback to improve the implementation.


lua/orgmode/colors/markup_highlighter.lua
```diff
@@ -110,15 +114,51 @@ local function is_valid_markup_range(match, _, source, _)
...
+  end
   if not is_valid_start then
     return false
   end
+
   local end_text = get_node_text(end_node, source, -1, 1)
-  return (end_text:len() < 3 or vim.tbl_contains(valid_post_marker_chars, end_text:sub(3, 3)))
-    and end_text:sub(1, 1) ~= ' '
+  local end_len = end_text:len()
+  is_valid_end = false
```
jgollenz, line 138: same here

lua/orgmode/colors/markup_highlighter.lua
```diff
@@ -110,15 +114,51 @@ local function is_valid_markup_range(match, _, source, _)
 
   local start_text = get_node_text(start_node, source, -1, 1)
   local start_len = start_text:len()
-
-  local is_valid_start = (start_len < 3 or vim.tbl_contains(valid_pre_marker_chars, start_text:sub(1, 1)))
-    and start_text:sub(start_len, start_len) ~= ' '
+  is_valid_start = false
+  if start_len < 3 then
```
jgollenz, line 118: If you know what this magic number stands for, please add a comment above ;)

jgollenz (changes requested):
Thanks for the contribution :+1: Please run `make format` in the root directory to format your PR correctly.

I can't say much about the added code, as I'm not familiar with that part of the project. However, I think the if statements for `is_valid_[start|end]` could benefit from some comments that explain what each case stands for.

kristijanhusak (changes requested):
Thanks for the PR.

Even though this seems like it could be put in markup predicate, I don't think it really is. We should have a separate predicate for math, like we do for hyperlinks. That way we can separate this logic out, since markup always have exactly 1 char before and 1 char after, where this math have more chars. Check how hyperlinks are handled and follow the similar path.

lua/orgmode/colors/markup_highlighter.lua
kristijanhusak: This is a situation where markup starts at the beginning of the line :). It definitely needs a comment.

We take markup start with offset char before and after (example: `test *bold*` would extract start as ` *b`. This allows to properly compare if before and after chars are suitable or valid for this region to be treated as valid markup.

gzagatti:
Thanks for the feedback. I will implement the changes before next week, they are all very sensible.

PR #428, opened 2022-10-19
Inline math highlighting (Continued from PR #426)

gzagatti:
This PR adds support to in-line math highlighting, it builds on previous work done for highlighting inline markups.

In following text:

```org
* Circle

Let a circle with radius \( r \), then the perimeter is given by \(2\pir\) and the area by \( π * r^2 \). 

Some examples:

One more thing \(\).
```

All of the in-line math should be highlighted with OrgTSLatex.

As I write a lot of technical texts, this features is very important. Feel free to provide some feedback to improve the implementation.

I am re-opening it here, as I messed up with Git and had to force push my changes.


lua/orgmode/colors/markup_highlighter.lua
```diff
@@ -173,7 +205,16 @@ local function get_matches(bufnr, first_line, last_line)
   for _, match, _ in query:iter_matches(root, bufnr, first_line, last_line) do
     for _, node in pairs(match) do
       local char = node:type()
+      if char == '\\' then
```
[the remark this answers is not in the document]
gzagatti: I think this was a bit of premature optimization. I have reverted most of the additions and simplified the table with matching elements. Rather than '\\(' we can just match on '\\'. 

In the future if we want to add markup for entities like `\cent` or latex fragments like `\newline` we might need to reconsider how to distinguish between those.

lua/orgmode/colors/markup_highlighter.lua
```diff
@@ -173,15 +221,38 @@ local function get_matches(bufnr, first_line, last_line)
...
-      local range = ts_utils.node_to_lsp_range(node)
-      local linenr = tostring(range.start.line)
-      taken_locations[linenr] = taken_locations[linenr] or {}
-      if not taken_locations[linenr][range.start.character] then
-        table.insert(ranges, {
-          type = char,
-          range = range,
-        })
-        taken_locations[linenr][range.start.character] = true
+      -- markups that start with a backslash \\ will be followed by another
```
kristijanhusak, line 224: This for loop still has "char" specific logic which does not belong here. It should just populate the ranges, and next loop should be in charge on any char specific logic (See how links are handled form line 305 - 315. Also, there should be a separate result entry for the math, like we have for the markup (`result`) and links (`link_result`).

kristijanhusak (changes requested):
It is much better now, but we should still have some logic organized differently.

gzagatti:
So I went ahead and implemented all of the Latex fragment highlight as described in the [Org Syntax](https://orgmode.org/worg/dev/org-syntax.html#LaTeX_Fragments).

The implementation is a bit more tricky but seems to work. I am not so sure about the performance on a large document.

The implementation still needs some improvement as `*\( 1 + *2* \)*` is not yet correctly parsed. I fix that soon.

This is a test file to check the markup:

```org
* Circle

Let a circle with radius \( r \), then the perimeter is given by \(2\pir+\) and the area by \( π * r^2 \).

Some examples:
 - this is /slanted/
 - a \test
 - a \goof{test}

One more thing \(\).

The book is on the table \enlargethispage{2\baselineskip} and \( \mathbb{R} + 1 \) and \foo{bar} 

And \foo{ bar } and \foo{bar } and \foo{bar} and \foo{} and \foo{bar and foo}

hello \foo{bar} and (\(1 - 2\)) and \alpha and \boo and \foo. And (\foo)

Foobar \(n\)-th character, -th should not be higlighted. 

This should only highlight =\ab= : \ab12

This should not be math: =\( 1 + 1\)= ~\(1 + 1\)~ 

This should only be verbatim: =*1*=

This should only be latex: \( 1 + =*2*= \) \(a ~2~ \)

This should not get higlighted: \2ab

T : *\( 1 + *2* \)*

This is *important* and ~foo(x) = x + 1~ foo /bar/

This is \( *1 + /1/ \)   *ab =a= cd* *fo /f f f/ =abc= a* *=ab=*

abc =a *a* b=
```

As I was testing the implementation, I noticed

lua/orgmode/colors/markup_highlighter.lua
gzagatti: I think it’s possible to move the char logic further down. Now, I understand the purpose of this loop better. I will fix that.

gzagatti:
I think I have addressed the comments and have fixed some remaining issues:

- Move char logic to appropriate loop.

- Returns latex results separately.

- Allow for proper nested markup highlight. The [syntax specification](https://orgmode.org/worg/dev/org-syntax.html#Emphasis_Markers) says text markup except for code and verbatim should accept any elements from the standard set, which sort of contradicts their own specification that says that text markup should only be preceded by blank space or a set of pre-defined characters (perhaps we can open an issue on the syntax specification upstream). This ensures that `*=foo=*` and `*foo =bar= foo*` have the same highlight. Also, Emacs and Pandoc export `*=foo=*` as bolded code.

Everything seems to be properly highlighted on this test file:

```org
* Circle

Let a circle with radius \( r \), then the perimeter is given by \(2\pi r\) and the area by \( π * r^2 \).

Some examples:
 - this is /slanted/
 - a \test
 - a \goof{test}

One more thing \(\).

The book is on the table \enlargethispage{2\baselineskip} and \( \mathbb{R} + 1 \) and \foo{bar} 

And \foo{ bar } and \foo{bar } and \foo{bar} and \foo{} and \foo{bar and foo}

hello \foo{bar} and (\(1 - 2\)) and \alpha and \boo and \foo. And (\foo)

Foobar \(n\)-th character, -th should not be higlighted. 

This should only highlight =\ab= : \ab12

This should not be math: =\( 1 + 1\)= ~\(1 + 1\)~ 

This should only be verbatim: =*1*=

This should only be latex: \( 1 + =*2*= \) \(a ~2~ \)

This should not get higlighted: \2ab

T : *\( 1 + *2* \)*

This is *important* and ~foo(x) = x + 1~ foo /bar/

Bold and verbatim *=abc=* slanted and verbatim /=foo=/ bold and slanted */foo/* bold, slanted and verbatim */=abc=/*

This is \( *1 + /1/ \) and   *ab =a= cd* *fo /f f f/ =abc=  and =abc *ab*= a* *=ab=*

abc =a *a* b=
```

lua/orgmode/colors/markup_highlighter.lua
```diff
@@ -197,30 +271,77 @@ local function get_matches(bufnr, first_line, last_line)
   local seek_link = {}
   local result = {}
   local link_result = {}
+  local latex_result = {}
 
+  local nested = {}
+  local can_nest = true
   for _, item in ipairs(ranges) do
+    if item.type == '(' then
```
kristijanhusak, line 279: Could we have these types in some kind of map? Something like this:
```
  local type_map = {
    ['('] = '\\(',
    [')'] = '\\(',
    ['}'] = '\\{'
  }

  for _, item in ipairs(ranges) do
    item.type = type_map[item.type] or item.type
    -- ...
```

This would remove some of these ifs and elseifs and make code slightly more readable. Of course, for the blocks where we have additional logic we need to have them.

kristijanhusak (changes requested):
I gave this a test, it works great!
Just one bug with the result entry and few small refactorings.

lua/orgmode/colors/markup_highlighter.lua
gzagatti: Done.

lua/orgmode/colors/markup_highlighter.lua
```diff
@@ -197,30 +271,77 @@ local function get_matches(bufnr, first_line, last_line)
...
+      item.range.start.character = item.range.start.character - 1
+      item.type = '\\('
+    elseif item.type == 'str' then
+      item.range.start.character = item.range.start.character - 1
+      local char = vim.api.nvim_buf_get_text(
```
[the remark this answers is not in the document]
gzagatti: Done.

kristijanhusak (approved):
Looks great, thanks!

gzagatti:
Thanks for the help! Glad I could contribute this PR. 

Btw, is there any reason why the in-line markups cannot be highlighted directly with Treesitter? I am asking that because it seems to be the case with other languages like Markdown and Latex. 

Also, do you think we could add the nodes that we highlight to Treesitter? I am just thinking of a situation where I want to get all bold text or all URLs in the document to do something else with them. Treesitter would make it very easy.

kristijanhusak:
> Btw, is there any reason why the in-line markups cannot be highlighted directly with Treesitter? I am asking that because it seems to be the case with other languages like Markdown and Latex.


There was an initial attempt to do that, but it was complicated to handle all these cases that you also fixed here with nesting and handling conflicts.
You can read through several of these issues on the grammar repo to get a better understanding of the problem:
* https://github.com/milisims/tree-sitter-org/issues/22
* https://github.com/milisims/tree-sitter-org/issues/3
* https://github.com/milisims/tree-sitter-org/issues/13


> Also, do you think we could add the nodes that we highlight to Treesitter? I am just thinking of a situation where I want to get all bold text or all URLs in the document to do something else with them. Treesitter would make it very easy.

I don't know if it's possible to have a second tree-sitter parser on the same filetype. Having one separately only for markup/latex would probably be very helpful, but I didn't think into that direction since I don't have much experience with tree-sitter.

gzagatti:
Thanks for the detailed explanation. I need to learn more about Treesitter, but I have a feeling that it should be able to capture all these markups that we are collecting manually.

PR #432, opened 2022-10-24
Fix nesting markup in links

kristijanhusak:
Related PR: https://github.com/nvim-orgmode/orgmode/pull/431

### DIFF
--- a/lua/orgmode/colors/markup_highlighter.lua
+++ b/lua/orgmode/colors/markup_highlighter.lua
@@ -10,55 +10,46 @@ local markers = {
   ['*'] = {
     hl_name = 'org_bold',
     hl_cmd = 'hi def org_bold term=bold cterm=bold gui=bold',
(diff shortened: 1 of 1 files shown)


kristijanhusak:
@gzagatti can you give this PR a test and see if it fixes the issues? It removes some of the code and should fix nesting of the markup and links, at least from what I saw how Emacs behaves.

kristijanhusak:
Ok I figured out this breaks all the LateX stuff that you did, so I need to reconsider this.

gzagatti:
The code does not seem to tackle cases like:

```
This is =a *1  b= and *ab cd* foo
```

Where `=a *1  b=` should be verbatim and `*ab cd*` should be bold. We might need to handle orphans like I did PR #431. I could try to bring that logic to this branch. I will just wait for your feedback on my comments in there.

PR #436, opened 2022-11-02
Added Enter as a multipurpose action

andreadev-it:
This PR makes the Enter key a multipurpose key. It will do a different action based on the context.
It currently overrides the "org_return" action and checks for these situations (in this specific order):
- open links
- open dates in calendar
- change headline status
- toggle a list item

Otherwise, it will do the "org_return" action as a default.

I've made this PR as a way to discuss this, as proposed in #301 

Some things I'd like to discuss are:
- Should we be using this as the default action for the Enter key? It seems to be like this on Emacs, but honestly, I had only tried it in Doom Emacs, not the vanilla one, if I recall correctly.
- The order in which the actions are checked is ok? It seemed sensible to me.
- What to do about the "org_return" / "use original return" actions? Should we keep these as is?

Thank you for your time, as always, I'm open to suggestions, have a nice day

### DIFF
--- a/lua/orgmode/config/defaults.lua
+++ b/lua/orgmode/config/defaults.lua
@@ -120,7 +120,7 @@ return {
       org_promote_subtree = '<s',
       org_demote_subtree = '>s',
       org_meta_return = '<Leader><CR>', -- Add heading, item or row
-      org_return = '<CR>',
+      -- org_return = '<CR>',
       org_insert_heading_respect_content = '<prefix>ih', -- Add new headling after current heading block with same level
       org_insert_todo_heading = '<prefix>iT', -- Add new todo headling right after current heading with same level
       org_insert_todo_heading_respect_content = '<prefix>it', -- Add new todo headling after current heading block on same level
@@ -142,6 +142,7 @@ return {
       org_clock_goto = '<prefix>xj',
       org_set_effort = '<prefix>xe',
       org_show_help = 'g?',
+      org_multipurpose_action = '<CR>',
     },
     edit_src = {
       org_edit_src_abort = '<prefix>k',
--- a/lua/orgmode/config/mappings/init.lua
+++ b/lua/orgmode/config/mappings/init.lua
@@ -132,6 +132,7 @@ return {
     org_clock_goto = m.action('clock.org_clock_goto', { opts = { desc = 'org clock goto' } }),
     org_set_effort = m.action('clock.org_set_effort', { opts = { desc = 'org set effort' } }),
     org_show_help = m.action('org_mappings.show_help', { opts = { desc = 'org show help' } }),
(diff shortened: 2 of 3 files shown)


andreadev-it:
I've seen that my PR is failing with an error about generating a new table row when the Enter key is pressed. I've tried it, but this feature doesn't seem to exists even in the official plugin... I've tried with the table-mode plugin either enabled and disabled on the official master branch.

jgollenz:
> What to do about the "org_return"

@kristijanhusak please refresh my memory: what was the reason for adding `org_return`?

> The order in which the actions are checked is ok

also seems sensible to me

> Should we be using this as the default action for the Enter key?

If it is this way in vanilla Emacs, we probably should have it here as well

kristijanhusak:
@jgollenz It was added when we added support for tables, to reformat the table when you press Return.
@andreadev-it  I think we should just add all this logic to `org_return` because your code doesn't handle fallback to the old "<CR>" mapping.

andreadev-it:
I'm looking at it and managed to merge my code with the one from the original org_return. But there seems to be a catch: the org_return function will be executed while in insert mode. This multipurpose action should (in my opinion) only available in normal mode. Is it still fine to merge them? It seems like they should be separated, with "org_return" for insert mode and "multipurpose_action" for normal mode...

kristijanhusak:
> But there seems to be a catch: the org_return function will be executed while in insert mode. This multipurpose action should (in my opinion) only available in normal mode. Is it still fine to merge them? It seems like they should be separated, with "org_return" for insert mode and "multipurpose_action" for normal mode...

My mistake then. `org_return` is insert mode mapping, and should not be removed. AFAIK Emacs Orgmode doesn't have this idea of a multipurpose Return mapping, so I'm not sure it should be part of the core. 

> Should we be using this as the default action for the Enter key? It seems to be like this on Emacs, but honestly, I had only tried it in Doom Emacs, not the vanilla one, if I recall correctly

Emacs doesn't have a normal mode, at least not without Evil mode. How does Return behave with Evil when you press it in normal mode in Emacs?

kristijanhusak:
Do you know if that functionality is part of evil mode or part of Doom Emacs?

andreadev-it:
Since it doesn't seem that this functionality was implemented by default in orgmode, I've decided to create a plugin to add it instead. If you guys agree that this is the correct way to do it, you can close this PR.

Here is a link to the plugin: https://github.com/andreadev-it/orgmode-multi-key

@kristijanhusak

kristijanhusak:
I guess it's ok for now to have it as a separate plugin. We could maybe have it also in here, but we need to have tests passing and this option being somehow opt-in.

jgollenz:
@andreadev-it is it ok to close this PR for now?

andreadev-it:
Yes, I agree to close it for now.

PR #437, opened 2022-11-03
Add level and is_archived to OrgHeadline

joaomsa:


joaomsa:
Tried to run the `scripts/gendoc.lua` to refresh the docs with the new fields but this line is breaking `lemmy-help`:

https://github.com/nvim-orgmode/orgmode/blob/master/lua/orgmode/api/headline.lua#L13

I think the `|` causes it to stop parsing the fields below.

What version of `lemmy-help` are you on?

kristijanhusak (approved):

PR #444, opened 2022-11-13
Fix wrong tag offset in agenda

### LINKED ISSUE #433: In week agenda view the tag are not shown fully
### Describe the bug

The tags on the heading are going beyond screen, had to scroll to the right to see the tags.

### Steps to reproduce

open agenda to view the week agenda.

### Expected behavior

Show the tags within the screen, i.e without need for horizontal scroll

### Emacs functionality

_No response_

### Minimal init.lua

n/a

Screenshot from 2022-10-30 01-36-55
Screenshot from 2022-10-30 01-37-04

### OS / Distro

fedora 36

### Neovim version/commit

0.9

### Additional context

_No response_

jgollenz:
closes #433

### DIFF
--- a/lua/orgmode/agenda/init.lua
+++ b/lua/orgmode/agenda/init.lua
@@ -25,49 +25,54 @@ function Agenda:new(opts)
     views = {},
     content = {},
     highlights = {},
+    win_width = utils.winwidth(),
   }
   setmetatable(data, self)
   self.__index = self
   return data
 end
 
-function Agenda:agenda(opts)
-  local view = AgendaView:new(vim.tbl_deep_extend('force', opts or {}, {
+---@param View table
+---@param opts? table
+function Agenda:open_agenda_view(View, opts)
+  self:open_window()
+  self.win_width = utils.winwidth()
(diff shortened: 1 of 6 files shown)


lua/orgmode/agenda/init.lua
```diff
@@ -25,36 +25,53 @@ function Agenda:new(opts)
...
   })):build()
   self.views = { view }
   return self:_render()
 end
 
 -- TODO: Introduce searching ALL/DONE
 function Agenda:todos()
-  local view = AgendaTodosView:new({ filters = self.filters }):build()
+  self:open_window()
+  self.win_width = utils.winwidth()
```
kristijanhusak, line 49: I think we can move this into the `open_window` method, so it doesn't have to be set from each agenda view type, to avoid some duplication.
Another solution might be to have like a generic method that will accept the "View" class, and you just initialize what's passed in, since all of the views are initialized with the same parameters. Something like `self:open_agenda_view(AgendaView)`

jgollenz: I went with the latter option, as it trims of quite some lines

kristijanhusak (approved):
LGTM, thanks!

PR #449, opened 2022-11-17
Adds :itoday: and :inow: abbreviations

chuck-flowers:
Adds two additional abbreviations that allow users to easily insert the
current date or time as an inactive timestamp.


jgollenz:
We also could do this: 
```
inoreabbrev <silent><buffer> :today: <C-R>=luaeval("require('orgmode.objects.date').today():to_wrapped_string(true)")<CR>
inoreabbrev <silent><buffer> :now: <C-R>=luaeval("require('orgmode.objects.date').now():to_wrapped_string(true)")<CR>
inoreabbrev <silent><buffer> :itoday: <C-R>=luaeval("require('orgmode.objects.date').now():to_wrapped_string(false)")<CR>
inoreabbrev <silent><buffer> :inow: <C-R>=luaeval("require('orgmode.objects.date').now():to_wrapped_string(false)")<CR>
```

I think it makes things a bit clearer, if one looks at the signature of `to_wrapped_string`. Anyhow, we should add a comment what `inow/itoday` stands for

jgollenz:
But then again, it's just a fancy way of doing the same thing with an additional performance cost :man_shrugging:

chuck-flowers:
I like yours better. I think it's a bit easier to see what's going on that way. I'll update now.

chuck-flowers:
Sorry, I forced push to my branch to amend the commit but apparently GitHub didn't like that. I'll have to open a new PR

PR #457, opened 2022-11-29
Makes the property searching case insensitive

chuck-flowers:
I noticed that in order to search agenda headlines using properties, the property names had to be all lower case. In Emacs, the property names are case insensitive when searching agenda items. I made a simple change to bring the behavior of nvim-orgmode in line with emacs.

### DIFF
--- a/lua/orgmode/parser/search.lua
+++ b/lua/orgmode/parser/search.lua
@@ -63,7 +63,7 @@ function Search:_matches(val, item)
     end
     return val == item.tags
   end
-  prop_name = vim.trim(prop_name)
+  prop_name = string.lower(vim.trim(prop_name))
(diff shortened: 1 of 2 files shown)


kristijanhusak:
Good catch. Can you just please update tests to pass? `props` in tests needs to be lowercased (`CATEGORY -> category`) to make them pass. They are already lowercased in code when we search for headlines, so this is issue just in this isolated unit test.

chuck-flowers:
@kristijanhusak thanks for the feedback. I believe that I have updated all relevant tests. It seems that that the CI/CD build is passing again.

kristijanhusak (approved):
Yeah, looks great! Thanks!

PR #458, opened 2022-11-30
Don't reload orgfile twice

joaomsa:
Calling `filetype detect` after setting the treesitter TODO highlights was triggering 2 back to back File.reload calls, since setting the `filetype` triggers the FileType autocommand.

Since we check if the filetype is already set, that means We can get away with just restarting the treesitter highlighter after adding new queries

This results in a noticeable perf improvement for me on when opening of an org file.


kristijanhusak:
Makes sense. I didn't even check if there is a way to reload only TS highlights. Thanks!

PR #460, opened 2022-11-30
Adds editorconfig to maintain indentation

chuck-flowers:
I've created a basic `.editorconfig` file to reflect the current indentation settings of the repository. The repository's settings differed from the ones I had set by default so I added the config to respect the existing code style.

More information on editorconfig can be found [here](https://editorconfig.org/)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,3 @@
+[*.lua]
+indent_style = space
+indent_size = 2


kristijanhusak:
Stylua should actually take care of this, but I guess it doesn't hurt to also have editorconfig.
Thanks!

chuck-flowers:
Sorry, I'm relatively new to lua and didn't think to check for language specific tools. I'll look into stylua now

PR #463, opened 2022-12-01
Agenda help improvements

### LINKED ISSUE #445: `g?` shows the same mappings for every agenda view
### Describe the bug

Mappings like `f - Go forward one span` are displayed in views that don't have spans (search, match tags, ...)

### Steps to reproduce

Open any agenda view that is not the actual agenda (list tags, search, match tags) and type `g?` to open the help

### Expected behavior

Only the key-mappings that are actually applicable to that view should be displayed

### Emacs functionality

-

### Minimal init.lua

scripts/minimal_init.lua

_No response_

### OS / Distro

Ubuntu 20.04

### Neovim version/commit

0.7.0

### Additional context

_No response_

jgollenz:
closes #445 

I chose to just use `vim.b.org_agenda_type` after all, as it was easier than passing it through with `opts`. If it is obfuscating the logic too much, let me know

### DIFF
--- a/lua/orgmode/objects/help.lua
+++ b/lua/orgmode/objects/help.lua
@@ -57,37 +57,41 @@ local helps = {
     { key = 'org_toggle_heading', description = 'Toggle current line to headline and vice versa' },
   },
   orgagenda = {
-    { key = 'org_agenda_later', description = 'Go forward one span' },
-    { key = 'org_agenda_earlier', description = 'Go backward one span' },
-    { key = 'org_agenda_goto_today', description = "Go to today's span" },
-    { key = 'org_agenda_goto_date', description = 'Jump to specific date' },
-    { key = 'org_agenda_day_view', description = 'Show day view' },
-    { key = 'org_agenda_week_view', description = 'Show week view' },
-    { key = 'org_agenda_month_view', description = 'Show month view' },
-    { key = 'org_agenda_year_view', description = 'Show year view' },
-    { key = 'org_agenda_switch_to', description = 'Open in current window' },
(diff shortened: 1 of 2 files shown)


lua/orgmode/objects/help.lua
```diff
@@ -201,7 +205,18 @@ function Help._prepare_orgagenda(mappings, max_height)
   end
 
   local content = { string.format(' **Orgmode mappings - Agenda%s:**', scroll_more_text), '' }
-  for _, item in ipairs(helps.orgagenda) do
+  if vim.b.org_agenda_type == 'agenda' then
```
kristijanhusak, line 208: Since `for` block is actually the same as for `generic` mappings, we can just merge both generic and agenda specific into one, and let the "generic" loop handle everything. Something like this:

```lua
-- Copy to make sure we do not override any references
local mappings = vim.deepcopy(helps.orgagenda.generic)
if vim.b.org_agenda_type == 'agenda' then
  mappings = utils.concat(helps.orgagenda.agenda_specific, mappings)
end

for _, item in ipairs(mappings) do
  -- ...
endfor
```

jgollenz: Great idea :)

kristijanhusak: Note that I forgot to copy "agenda_specific" mappings, since `utils.concat` will mutate it.

jgollenz: Should be fixed. Also made some changes regarding the replacement of `<prefix>` with the actual prefix from the config.

PR #465, opened 2022-12-09
Improves type annotations

chuck-flowers:
Updates type annotations in `Logbook` and `Section`

Adds and updates type annotations to reflect the actual implementations of the
modules/classes.

### DIFF
--- a/lua/orgmode/parser/logbook.lua
+++ b/lua/orgmode/parser/logbook.lua
@@ -134,7 +134,7 @@ function Logbook:recalculate_estimate(line)
 end
 
 ---@param lines string
----@param node unknown
+---@param node userdata
 ---@param dates Date[]
 ---@return Logbook
 function Logbook.parse(lines, node, dates)
--- a/lua/orgmode/parser/logbook.lua
+++ b/lua/orgmode/parser/logbook.lua
@@ -54,8 +54,8 @@ function Logbook:get_total_minutes(from, to)
   return total_minutes
 end
 
----@param from string
----@param to string
+---@param from? string
+---@param to? string
 ---@return Duration
 function Logbook:get_total(from, to)
   return Duration.from_minutes(self:get_total_minutes(from, to))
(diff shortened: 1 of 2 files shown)


lua/orgmode/parser/section.lua
```diff
@@ -34,8 +34,37 @@ local config = require('orgmode.config')
...
+---@field own_tags string[]
+---@field parent Section
+---@field priority string
+---@field properties table
+---@field range Range
+---@field root File
+---@field tags string[]
+---@field title string
+---@field todo_keyword_node unknown
+local NewSectionOptions = {}
```
kristijanhusak, line 59: There's actually no need to define an empty variable to get type annotations. LSP should know to pick it up even without it.

kristijanhusak (changes requested):
Thanks for the PR. Just address that one comment and it's good to go.

Just a note if you are maybe trying to use the internals for a separate plugin, backward compatibility is not guaranteed for the internals. If you think you need some of these things exposed, I would suggest to open up a separate PR and introduce a public api through `orgmode.api` files.

Edit: `node` can be typed as `userdata` instead of `unknown` and `any`.

lua/orgmode/parser/section.lua
chuck-flowers: I've removed the two unused variables that I added as part of this contribution

chuck-flowers:
I added these type annotations as part of some fixes I'm contributing to this plugin. It took me a minute to figure out what types certain variables should be from the call site, so once I figured it out by looking at `Section` and `Logbook` I updated the type annotations to save myself or the next person some time.

lua/orgmode/parser/logbook.lua
```diff
@@ -134,7 +134,8 @@ function Logbook:recalculate_estimate(line)
 end
 
 ---@param lines string
----@param dates string
+---@param node unknown
```
kristijanhusak, line 137: You can type this as `userdata`, since tree-sitter nodes are of that type. Same thing for the `_parse_clocks` params.

chuck-flowers: I've update this as well

kristijanhusak (approved):

kristijanhusak:
Thanks!

PR #468, opened 2022-12-12
Cache markup tree-sitter queries by changetick to improve performance

kristijanhusak:
This PR fixes flickers and slow scrolling issues by caching the queries by changetick.

### DIFF
--- a/lua/orgmode/colors/custom_highlighter.lua
+++ b/lua/orgmode/colors/custom_highlighter.lua
@@ -5,12 +5,9 @@ local MarkupHighlighter = nil
 local valid_bufnrs = {}
 
 ---@param bufnr number
----@param first_line number
----@param last_line number
-local function apply_highlights(bufnr, first_line, last_line, tick_changed)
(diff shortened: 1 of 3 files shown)


jgollenz:
Oh boy, it really does fix it :heart_eyes: thanks a lot, this has annoyed me for a very long time. Do you mind giving me a short explanation of what the issue was?

kristijanhusak:
Sure. The flickering issue was caused by not using `on_line` callback but only `on_win`. This callback clears the highlights and re-applies them to the whole visible content, which can be slow and cause flickering. `on_line` does the same thing but on a line level. I figured this out by checking the code for the internal tree-sitter highlighter.

The second issue was performance related. Redraw is called very often and for each line, and doing `query:iter_matches` seems to be really slow, especially at the visible content that has a lot of markup matches (which is of course expected).
We go around that issue by memoizing each line by changetick, since the content is not really changing while moving around, only when editing something. If there is a lot of content around the cursor in insert mode, it still might cause some slowness, but generally should be better than before.

PR #470, opened 2022-12-13
Fixes a bug in completing future tasks

chuck-flowers:
Updates the repeater logic so that tasks which are completed before
their deadline has been reached, will properly repeat forward.

### DIFF
--- a/tests/plenary/object/date_spec.lua
+++ b/tests/plenary/object/date_spec.lua
@@ -571,6 +571,13 @@ describe('Date object', function()
     assert.are.same(inTwoWeeks:to_string(), sunday:apply_repeater_until(inTwoWeeks):to_string())
   end)
 
+  it('should apply repeater to future dates', function()
+    local tomorrow = Date.now({ adjustments = { '++1d' } }):add({ day = 1 })
+    local day_after_tomorrow = tomorrow:add({ day = 1 })
+    local updated_date = tomorrow:apply_repeater()
+    assert.are.same(updated_date, day_after_tomorrow)
+  end)
+
   it('should cache check for today', function()
     local today = Date.today()
     assert.is.Nil(today.is_today_date)
--- a/tests/plenary/object/date_spec.lua
+++ b/tests/plenary/object/date_spec.lua
@@ -575,7 +575,7 @@ describe('Date object', function()
     local tomorrow = Date.now({ adjustments = { '++1d' } }):add({ day = 1 })
     local day_after_tomorrow = tomorrow:add({ day = 1 })
     local updated_date = tomorrow:apply_repeater()
-    assert.are.same(updated_date, day_after_tomorrow)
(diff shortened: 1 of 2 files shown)


kristijanhusak:
Can you please give an example? I'm not sure what's the difference between old and new code.

chuck-flowers:
The issue that I've run into recently is that I have tasks for buying Christmas presents for friends and family. I have the DEADLINE for these TODOs set to `2022-12-25 Sun ++1y -40d>`. I've already bought some of these presents so I'd like to mark them as complete using `cit`.

In the old logic, an entry would be made in the logbook but the deadline would not be updated because the while loop which incremented the `DEADLINE` date would increment only while the deadline was less than today. Since today is before the current `DEADLINE` no actual incrementing of the `DEADLINE` would occur.

Now with my new logic, at least one increment will occur before the condition of the while loop is checked. Essentially it's the difference between a `while` loop and a `do ... while` loop from other C-like languages. So now when I `cit` these Christmas present headlines, they are updated to `<2023-12-25 Mon ++1y -40d>` just like they are in Emacs.

I should note that for my emacs testing, I'm using the official `emacs` package as its distributed in Arch Linux in case that makes a difference.

chuck-flowers:
This commit also includes some slight changes and additional comments to the relevant code to help the next programmer understand more quickly what this code is doing.

kristijanhusak:
Thanks for the explanation. It is indeed a bug. Could you also please add a test case for this in the `date_spec`, somewhere around [here](https://github.com/nvim-orgmode/orgmode/blob/master/tests/plenary/object/date_spec.lua#L565)?

chuck-flowers:
I've added a test as requested. It's the first Lua test I've ever had to write so please don't hesitate to let me know if any adjustments should be made.

tests/plenary/object/date_spec.lua
```diff
@@ -571,6 +571,13 @@ describe('Date object', function()
     assert.are.same(inTwoWeeks:to_string(), sunday:apply_repeater_until(inTwoWeeks):to_string())
   end)
 
+  it('should apply repeater to future dates', function()
+    local tomorrow = Date.now({ adjustments = { '++1d' } }):add({ day = 1 })
+    local day_after_tomorrow = tomorrow:add({ day = 1 })
+    local updated_date = tomorrow:apply_repeater()
+    assert.are.same(updated_date, day_after_tomorrow)
```
kristijanhusak, line 578: Lets use `:to_string()` on both dates just to keep it consistent with other assertions.

kristijanhusak (changes requested):
Test looks good, just a minor tweak and it's good to go.

tests/plenary/object/date_spec.lua
chuck-flowers: I've added the to_strings as requested

kristijanhusak (approved):
Everything looks great now, thanks!

PR #489, opened 2023-01-17
org_time_stamp append date after cursor instead of inserting before it

Maltimore:
I think it would be a much better default to append a date after the cursor instead of inserting it before the cursor.

### DIFF
--- a/lua/orgmode/org/mappings.lua
+++ b/lua/orgmode/org/mappings.lua
@@ -862,7 +862,7 @@ function OrgMappings:org_time_stamp(inactive)
     if date_start then
       date_string = '--' .. date_string
     end
-    vim.cmd(string.format('norm!i%s', date_string))
+    vim.cmd(string.format('norm!a%s', date_string))
   end)
 end
 


Maltimore:
I checked how emacs org-mode does this. And the answer is a bit tricky, because in emacs, the cursor positioning is different. If a line is 10 characters long, and you position the cursor to the end of the line in *vim*, then the cursor will be on the 10th character. In *emacs*, if you do the same the cursor will be on an imaginary 11th character. In `nvim-orgmode`, if the cursor is at the end of the line and you add a timestamp, the timestamp will be added before the last character of the line. In emacs orgmode, it will be after.

In the following, `abc|d|e` means that the cursor is on the `d` in normal mode.
In vim:
```
* test
  test tex|t|
```
when you add a timestamp, this becomes

```
* test
  test tex<2023-01-20 Fri|>|t
```
In *emacs* the cursor will be one character further:
```
* test
  test text| |
```
when you add a timestamp, this becomes
```
* test
  test text<2023-01-20 Fri>| |
```
I believe we should copy emacs orgmode's behavior.
------
Apart from that, as regards to the failing tests: the tests fail for some undocumented behavior regarding entering date ranges. IMHO, we could drop this date-range entering behavior for now, since it is undocumented and it isn't a lot of effort to enter the `--` manually.
------
@kristijanhusak @jgollenz I'd love your input on this! If you agree, I would delete the automatic date-range entering, and delete the two tests for this.

kristijanhusak:
I pushed a change for this. This is the summary:

1. By default, date is added after cursor if cursor is not on an existing date. For example (`| |` is cursor position):
```
* TODO Test
  Somethin|g|
```
Results in:
```
* TODO Test
  Somethin|g|<selected date>
```
2. If cursor is on existing date, it modifies that date
3. If cursor is on a white space after a date, it inserts a date range. For example (`| |` is cursor position):
```
* TODO Test
  <2023-01-20 Fri>| |
```
Results in:
```
* TODO Test
  <2023-01-20 Fri>--<selected date>
```

PR #493, opened 2023-01-26
tests/minimial_init.vim: set language such that tests pass on non en_US systems

Maltimore:
My system's locale isn't en_US and therefore I had test failures on doing `make test`.


kristijanhusak (approved):

kristijanhusak:
Thanks!

PR #494, opened 2023-01-26
Match by category

chuck-flowers:
Allows a user to match using the implicit `CATEGORY` property as mentioned in org's
documentation [here](https://orgmode.org/manual/Categories.html)

### DIFF
--- a/lua/orgmode/parser/file.lua
+++ b/lua/orgmode/parser/file.lua
@@ -206,7 +206,7 @@ function File:apply_search(search, todo_only)
       return false
     end
     return search:check({
-      props = vim.tbl_extend('force', {}, item.properties.items, {
+      props = vim.tbl_extend('keep', {}, item.properties.items, {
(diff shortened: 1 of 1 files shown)


chuck-flowers:
@kristijanhusak if you have a moment can you take a look at this? It's a very small change that would improve my workflow. I currently have several PRs open so it's difficult for me to simply reference this branch, and give up the features from my other larger PRs.

kristijanhusak:
This will actually override the category set in the properties of a headline, which should have precedence over file category. It's best to create the props hash and then override category only if it's not set.

chuck-flowers:
I've updated the code to use `keep` instead of `force`. This will force the search logic to favor what's contained within properties instead of what I'm pulling from the section itself.

kristijanhusak (approved):

kristijanhusak:
Thanks!

Issue #374, opened 2022-07-25
Include plain-timestamps in notifications


Maltimore:
### Does this feature exist in Emacs orgmode core?

N/A

### Feature value

Currently, only `DEADLINE`s and `SCHEDULED` tasks are considered for notifications. I think it would be an obvious extension to consider plain timestamps as well, because you need to be reminded of meetings and such, which are neither tasks nor deadlines.

PR #505, opened 2023-02-20
notifications: show notifications for all active dates

Maltimore:
This seems like an obvious/easy fix, but maybe (probably) there was a reason it wasn't done like this to begin with?
It works for me so far.
Closes https://github.com/nvim-orgmode/orgmode/issues/374

### DIFF
--- a/lua/orgmode/notifications/init.lua
+++ b/lua/orgmode/notifications/init.lua
@@ -91,7 +91,7 @@ function Notifications:get_tasks(time)
   local tasks = {}
   for _, orgfile in ipairs(Files.all()) do
     for _, headline in ipairs(orgfile:get_opened_unfinished_headlines()) do
-      for _, date in ipairs(headline:get_deadline_and_scheduled_dates()) do
+      for _, date in ipairs(headline:get_valid_dates_for_agenda()) do
         local reminders = self:_check_reminders(date, time)
         for _, reminder in ipairs(reminders) do
           table.insert(tasks, {


dtvillafana:
I have made the same change in my fork, and the reasoning in the linked issue is sound and I agree with it, but from a software product standpoint it would make sense to make it a configurable option for backwards compatibility and extensibility.

Maltimore:
Hey, sorry for not getting back to you @jgollenz, and thanks for your interest @dtvillafana. I'm still using this patch as well.
Making it configurable is of course always an option but to be honest I see absolutely no reason *not* to include active dates in the agenda, I mean otherwise what's the difference to passive dates?

Maltimore:
Sorry, forget what I said. This is not about the agenda but about notifications. Yes I suppose one could have a config option whether to show notifications for active timestamps. I still wonder what the reasoning is to not show notifications for timestamped events by default, but to show them for SCHEDULED and DEADLINE.

jgollenz:
I took a look at the docs and actually the behavior you expect should already be happening, so strictly speaking this is a bug fix :wink: It says `agenda tasks notifications` there, so you are absolutely right. We also already have options for this, albeit the other way round: 

```
deadline_reminder = true,
scheduled_reminder = true,
```

I tested your change and it looks good to me. When you mark the PR as ready I can merge it :+1:

jgollenz:
> it would make sense to make it a configurable option for backwards compatibility and extensibility

In this case behavior was actually not in line with what the docs say, so I don't consider it necessary to add an option. If need for it exists, we can think of introducing it but for now this is a bug-fix.

Maltimore:
Ok, ready for review!

jgollenz:
Did some more checks and it seems that
```
deadline_reminder = false,
scheduled_reminder = false,
```
have no effect any more. I get notifications for both DEADLINE and SCHEDULED tasks even when they should be turned off. Can you confirm this?

Maltimore:
Ah yes, you're right!
Sorry, I was too eager to just fix my own use-case and didn't think about these settings. I'll have to make this PR much more through. I'll switch it back to draft.

Maltimore:
Updated.
According to my tests, this now works as intended:
- notifications for plain timestamps, DEADLINE and SCHEDULED
- no notification for SCHEDULED if `scheduled_reminder = false` (I haven't tested DEADLINE to be honest but code seems symmetric to SCHEDULED)
- no notification for end times of date ranges
I invite you to test these changes yourself, as it's possible that there are other irregularities (like the "end of a date range" reminder). I'll also continue running this branch. And hopefully eventually we could merge.

PR #509, opened 2023-02-25
Add 'win_border' config option

### LINKED ISSUE #507: [Feature] Allow a user to specify the type of border that orgmode windows user

N/A

### Orgmode link

N/A

### Feature value

Currently there are several floating windows opened by orgmode e.g. the help window, the calendar, the notification etc.

Most of these windows are hardcoded to open with a `single` border, however as a user I have changed the border of every plugin I use to match a style I prefer but do not have the same option with orgmode. 

It would be nice if a user could specify `border = 'style'` and have that apply for all windows. Since consistency with the user's chosen style is the key I don't believe it needs several options but rather just one.

### Additional context

_No response_

jgollenz:
closes #507 

@akinsho could you please have a look at this. It adds a new config option `win_border` that just passes the border options that can be passed to `nvim_open_win()` through.

### DIFF
--- a/lua/orgmode/config/defaults.lua
+++ b/lua/orgmode/config/defaults.lua
@@ -42,6 +42,7 @@ local DefaultConfig = {
   org_edit_src_content_indentation = 0,
   diagnostics = true,
   win_split_mode = 'horizontal',
+  win_border = 'single',
   notifications = {
     enabled = false,
     cron_enabled = true,
--- a/DOCS.md
+++ b/DOCS.md
@@ -132,6 +132,32 @@ Always open horizontally with specific height of 20 lines:
 win_split_mode = '20split'
 ```
 
+#### **win_border**
+*type*: `string|string[]`<br />
+*default value*: `single`<br />
+Border style of floating windows.<br />
+Available options:
+* `none` - No border (default)
+* `single` - A single line box
+* `double` - A double line box
+* `rounded` - Like "single", but with rounded corners ("╭" etc.)
+* `solid` - Adds padding by a single whitespace cell
+* `shadow` - A drop shadow effect by blending with the background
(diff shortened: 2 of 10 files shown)


akinsho:
@jgollenz thanks for picking this up 🙏🏿 , just gave it a try and it works nicely.


One (possibly unrelated) thing it brought to light now that all the window borders are consistent for me that the highlights for the help menu don't match the same highlights used for the agenda/capture (if those are set to floats)

Screenshot 2023-02-25 at 09 16 44

Other than the `win_split_mode = 'float'` and the border I haven't changed anything window related so I think the highlights just have different values set 🤔

jgollenz:
Highlights, my kryptonite :grin: there's really not much that I can add to that conversation, it's a topic I am mostly unfamiliar with. @kristijanhusak does probably know more.
 
> can be dealt with as a separate issue

Absolutely. If you could open one that would be great :+1: I wouldn't know how to properly formulate the issue.

akinsho:
Done 👍🏿, from the perspective of my original issue re. borders though this works great.

lua/orgmode/config/defaults.lua
```diff
@@ -42,6 +42,7 @@ local DefaultConfig = {
   org_edit_src_content_indentation = 0,
   diagnostics = true,
   win_split_mode = 'horizontal',
+  win_border = 'rounded',
```
kristijanhusak, line 45: I think we should stick with `single` here since that is currently hardcoded everywhere.

jgollenz: done

kristijanhusak (approved):

kristijanhusak:
Thanks!

PR #512, opened 2023-02-26
Add basic `org-insert-link` functionality

jgollenz:
Described [here](https://orgmode.org/manual/Handling-Links.html#index-C_002dc-C_002dl)

There is no completion support for protocols (`http`, etc.) yet.

In light of other GH issues that are concerned with hyperlinks I'd like to discuss whether it makes sense to add a hyperlink object. It would possibly ease development in the future. Also, how about parsing them? Opening hyperlinks from the agenda requires to display all the links of a headline body to the user. There is also [`org-[next|previous]-link`](https://orgmode.org/manual/Handling-Links.html#index-C_002dc-C_002dl), that jumps to the next|previous link in the buffer

### DIFF
--- a/DOCS.md
+++ b/DOCS.md
@@ -792,6 +792,9 @@ Toggle current line checkbox state
 #### **org_toggle_heading**
 *mapped to*: `<Leader>o*`<br />
 Toggle current line to headline and vice versa. Checkboxes will turn into TODO headlines.
+#### **org_insert_link**
+*mapped to*: `<Leader>oil`<br />
+Insert a hyperlink at cursor position. When the cursor is on a hyperlink, edit that hyperlink.<br />
 #### **org_open_at_point**
 *mapped to*: `<Leader>oo`<br />
 Open hyperlink or date under cursor. When date is under the cursor, open the agenda for that day.<br />
--- a/lua/orgmode/config/defaults.lua
+++ b/lua/orgmode/config/defaults.lua
@@ -142,6 +142,7 @@ local DefaultConfig = {
       org_schedule = '<prefix>is',
       org_time_stamp = '<prefix>i.',
       org_time_stamp_inactive = '<prefix>i!',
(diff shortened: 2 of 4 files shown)


kristijanhusak:
> In light of other GH issues that are concerned with hyperlinks I'd like to discuss whether it makes sense to add a hyperlink object. It would possibly ease development in the future. Also, how about parsing them?

I definitely support this. Regarding the parsing, we can probably do parsing only when necessary, not on init.

jgollenz:
> I definitely support this

Great :slightly_smiling_face: I'll prepare another PR for it.

> do parsing only when necessary

Sounds sensible. It's also large enough to do it in a dedicated PR imo.

Is there anything left for this PR?

jgollenz:
@kristijanhusak ping

kristijanhusak (approved):
Looks good! Thanks!

And sorry for delay.

Maltimore:
@jgollenz the mapping for `insert-link` does not appear in the list of mappings that I get by pressing `g?` in an org file. Is that on purpose?
A small but fantastic feature by the way, thanks :)

Maltimore:
@jgollenz sorry to bother you again. I noticed that, similar to inserting time stamps, the link is inserted before the cursor and not after. I would find it more intuitive if the link was inserted after the cursor, for the same reasons as the time-stamps (i.e., usually you write some text, then enter a space, and then you'd like to have a link after the space. Currently you have to enter two spaces and then enter the link and then remove one of the spaces again.

jgollenz:
> Is that on purpose?

of course not :grin:

> more intuitive if the link was inserted after the cursor

I agree. Please check [this branch](https://github.com/jgollenz/orgmode/tree/fix/insert-link-after-cursor). Does it work as expected?

Maltimore:
Works like a charm! Thank you! :)

PR #517, opened 2023-02-28
File link improvements

### LINKED ISSUE #514: Allow the omission of `file:` when external linking for special cases.

Yes

### Orgmode link

[External Links](https://orgmode.org/manual/External-Links.html)

### Feature value

_No response_

### Additional context

From the link above: 

> As a special case, “file” prefix may be omitted if the file name is complete, e.g., it starts with ‘./’, or ‘/’.

For example, `[[file:./book/notes.org][notes]]`, `[[file:book/notes.org][notes]]`, and `[[./book/notes.org][notes]]` are all equivalent external links. However, `[[book/notes.org][notes]]` would be an internal link.

I'm currently unable to follow links of this form. For better compatibility with existing org-mode files, I think this type of link should be allowed.

jgollenz:
fixes #514 

Currently we have a different syntax for specifying line numbers in file links than what is used in emacs orgmode. I'll look into that

### DIFF
--- a/DOCS.md
+++ b/DOCS.md
@@ -1021,10 +1021,10 @@ And going to line `4` and pressing `gqgq`, it will format it to this:
 
 The format for links is either `[[LINK]]` or `[[LINK][DESCRIPTION]]`. If a description is provided, the actual link is concealed in favor of the description.
 
-Hyperlink types supported:
+Supported hyperlink types:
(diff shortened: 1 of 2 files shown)


kristijanhusak:
Already implemented. Thanks for PR anyway!

PR #523, opened 2023-03-09
Use Treesitter for folding

WhiteBlackGoose:
Ok so

This is a truly amazing plugin

However, folding really works like shit for some reason. It always breaks. Here's a simple example:

ts-bug

I know it can collapse properties and other neat stuff. However, I personally don't need it. For me, it now collapses sections just perfectly, and I keep my sanity.

### DIFF
--- a/ftplugin/org.vim
+++ b/ftplugin/org.vim
@@ -20,7 +20,7 @@ endfunction
 setlocal nomodeline
 setlocal fillchars+=fold:\ 
 setlocal foldmethod=expr
-setlocal foldexpr=OrgmodeFoldExpr()
+setlocal foldexpr=nvim_treesitter#foldexpr()
 setlocal foldtext=OrgmodeFoldText()
 setlocal formatexpr=OrgmodeFormatExpr()
 setlocal foldlevel=0
--- a/ftplugin/org.vim
+++ b/ftplugin/org.vim
@@ -1,10 +1,6 @@
 lua require('orgmode.config'):setup_mappings('org')
 lua require('orgmode.config'):setup_mappings('text_objects')
 
-function! OrgmodeFoldExpr()
-  return luaeval('require("orgmode.org.indent").foldexpr()')
-endfunction
-
 function! OrgmodeFoldText()
   return luaeval('require("orgmode.org.indent").foldtext()')
 endfunction
--- a/lua/orgmode/utils/treesitter.lua
+++ b/lua/orgmode/utils/treesitter.lua
@@ -42,8 +42,8 @@ function M.get_node_at_cursor(cursor, winnr, ignore_injected_langs)
   if vim.treesitter.get_node then
     return vim.treesitter.get_node({
       bufrn = buf,
-      pos = {cursor[1] - 1, cursor[2]},
-      ignore_injections = ignore_injected_langs
+      pos = { cursor[1] - 1, cursor[2] },
+      ignore_injections = ignore_injected_langs,
     })
   end
 
--- a/lua/orgmode/utils/treesitter.lua
+++ b/lua/orgmode/utils/treesitter.lua
@@ -39,9 +39,11 @@ function M.get_node_at_cursor(cursor, winnr, ignore_injected_langs)
   winnr = winnr or 0
   local buf = vim.api.nvim_win_get_buf(winnr)
   -- TODO: Use only this function when 0.8 is released
-  if vim.treesitter.get_node_at_pos then
-    return vim.treesitter.get_node_at_pos(buf, cursor[1] - 1, cursor[2], {
-      ignore_injections = ignore_injected_langs,
+  if vim.treesitter.get_node then
+    return vim.treesitter.get_node({
+      bufrn = buf,
+      pos = {cursor[1] - 1, cursor[2]},
+      ignore_injections = ignore_injected_langs
     })
   end
 
--- a/queries/org/folds.scm
+++ b/queries/org/folds.scm
@@ -1,0 +1,2 @@
+((section) @fold
+  (#trim! @fold))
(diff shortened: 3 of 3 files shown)


jgollenz:
How are you testing this? When I disable `setlocal foldexpr=OrgmodeFoldExpr()` in `ftplugin/org.vim` I get really wonky behavior for the foldtext. Do you mind sharing a gif of how it looks on your end.

Also, I might be missing something here, but wouldn't adding `(property_drawer)` to `folds.scm` take care of the properties as well? Seems™ to work

WhiteBlackGoose:
It works perfectly for me, as it only collapses headers and tables. 

Demo:
new-fold-demo

> Also, I might be missing something here, but wouldn't adding (property_drawer) to folds.scm take care of the properties as well?

Absolutely, but I personally didn't like folding properties. Ofc you can add it

jgollenz:
Do you mind sharing your config regarding `foldexpr` and `foldtext`? Unfortunately your changes don't seem to work for me. I can't fold tables and they break the foldtext.

In any case, I first need to figure out why we haven't been using treesitter for this up until now. If we can catch all the features we had so far also with treesitter, we can merge this.

One more thing: I'm not able to fold tables in emacs orgmode. The `<TAB>` is used to navigate cells there. From the gif it looks like you are folding the table from any location within the table, not just the first line. Is that correct?

jgollenz:
btw, the tests are failing due to incorrect styling. Please run `make format` in the root of the repository. Thanks :+1:

WhiteBlackGoose:
Apparently I had ufo, without it, folding the table doesn't work. :thinking: 

I'm sorry for the confusion. But that's why it's a draft, and that's why I'm not in rush with styling changes/removing orgmode's folding.

jgollenz:
No worries :grin: well, I'm not absolutely sure but I had the impression that folding tables was working for me yesterday. Might confuse it with property drawers though

WhiteBlackGoose:
I'm not trying to be a dick, but currently the folding provided by the plugin works terribly. For me, it would make perfect sense remove all such functionality and move this job to things which do it well and consistently: treesitter, ufo (we probably can do it without ufo, I just don't know how).

But the plugin is great otherwise :smile: (for now, I'm on my own fork of it).

> The <TAB> is used to navigate cells there. From the gif it looks like you are folding the table from any location within the table, not just the first line. Is that correct?

Yes, but not with tab, but with default vim keys.

jgollenz:
I am all for moving this to treesitter, but like I said, I first need to check what we are doing differently in our custom foldexpr and why.

Regarding the `(property_drawer)`: I'm not a maintainer here, so I doubt I can push to your master branch. Either you add those changes or I have to create a new branch and commit with you as author. But since I can't merge my own PRs anyway, I'd prefer the former :wink:

kristijanhusak:
I didn't give this a test, but folding in here is a bit more complicated than folding in a more standard syntax like a programming language. I would suggest to also add `drawer`, `property_drawer` and `block` to the folds, and then we can see if that maybe can work out.

kristijanhusak:
Gave it a quick test. This might actually work better than the current solution since it handles a few edge cases.
To summarize my review:
* Add `drawer`, `property_drawer` and `block` to the folds list
* Replace `setlocal foldexpr=OrgmodeFoldExpr()` with `setlocal foldexpr=nvim_treesitter#foldexpr()` in `ftplugin/org.vim` file
* Fix stylua formatting issues so build properly passes

WhiteBlackGoose:
Very nice, now it works without ufo and it doesn't annoyingly autofold.

WhiteBlackGoose:
hold up it doesn't work

WhiteBlackGoose:
After a revisit:
- With UFO, it annoyingly autofolds (can be disabled with removing foldlevel, but then default folding doesn't work)
- Without UFO, it folds wrong again

kristijanhusak:
Can you elaborate what's wrong without UFO?

WhiteBlackGoose:
another-bug-aaa

WhiteBlackGoose:
There must be something that UFO overrides that I didn't

kristijanhusak:
I think it works with UFO because it refreshes the folds fairly often, while without UFO that doesn't happen until you force edit the buffer or do `zx` manually. Do you have the same issues with the current orgmode folding?

WhiteBlackGoose:
Slightly different, but close :thinking: .

kristijanhusak:
I'll be merging this in since I don't see any difference in behavior between the current implementation and TS one. We can always revert it in case people find some issues with it. Thanks for the PR!

kristijanhusak (approved):

jgollenz:
We should not forget about removing our custom foldexpr after some time, since it's dead code now.

kristijanhusak:
> We should not forget about removing our custom foldexpr after some time, since it's dead code now.

Yeah, I plan to do this once we make sure the current solution works.

jgollenz:
Tangent: do you think we can move things from `ftplugin/orgmode.vim` into lua? The issue I had about foldtext was the the fillchar was set to `\ `, backslash+space, which my config stripped silently. A trailing space without a comment is not very explicit and in lua we would at least have `'fold: '` wrapped in quotes. I tried and it worked fine for things that are just options but it failed on the `foldtext` and `foldexpr` functions. I guess it's not possible to pass such lua functions to `foldexpr` without wrapping it in a vimscript function?

kristijanhusak:
Since we are on v0.8+, I think we could move all the `vim` files to `lua`, except maybe syntax files. I'm just not sure how to handle these options that require a vim function, like `foldtext`, foldexpr, etc.. I assume we can do something like `vim.opt_local.foldtext = 'v:lua.require("orgmode.org.indent").foldtext()'` but I didn't give it a test. If you have something prepared create a draft PR and we can see if it works out.

PR #525, opened 2023-03-12
Replace 'org.vim' with 'org.lua'

jgollenz:
As discussed in #523

I'm not aware of anything to make the abbreviations nicer.
Also, I haven't tested omnifunc yet


jgollenz:
@kristijanhusak what's your opinion on that so far?

ftplugin/org.lua
```diff
@@ -0,0 +1,38 @@
...
+-- options
+vim.opt_local.modeline = false
+vim.opt_local.fillchars:append('fold: ')
+vim.opt_local.foldmethod = 'expr'
+vim.opt_local.foldexpr = 'nvim_treesitter#foldexpr()'
+vim.opt_local.foldtext = 'v:lua.require("orgmode.org.indent").foldtext()'
+vim.opt_local.formatexpr = 'v:lua.require("orgmode.org.format")()'
+vim.opt_local.foldlevel = 0
+vim.opt_local.commentstring = '# %s'
+vim.opt_local.omnifunc = 'function! OrgmodeOmni(findstart, base)'
```
kristijanhusak, line 13: Seems like this is not working. We can probably add this to `plugin/orgmode.vim` temporarily until we drop the support for versions < 0.9.0, because 0.9.0 introduced [lua_omnifunc](https://neovim.io/doc/user/lua.html#vim.lua_omnifunc()) which we can leverage here

kristijanhusak (changes requested):
Mostly looks good, just some issues around omnifunc

PR #528, opened 2023-03-17
Add error message for missing `requrie('orgmode').setup_ts_grammar()`

KentoOgata:
It is necessary to call setup_ts_grammar() before setup() to specify the revision, but the current error message does not make that clear.

Adding this PR will ease the introduction for new users.


kristijanhusak (approved):

kristijanhusak:
Thanks!

PR #531, opened 2023-03-19
Add org-log-into-drawer

### LINKED ISSUE #455: Add support for `org-log-into-drawer`, `org-add-note` and `org-agenda-add-note`

Yes

### Orgmode link

https://orgmode.org/org.html#index-org_002dlog_002dinto_002ddrawer

https://orgmode.org/org.html#Drawers

https://orgmode.org/org.html#index-org_002dagenda_002dadd_002dnote

### Feature value

_No response_

### Additional context

Related issue: https://github.com/nvim-orgmode/orgmode/issues/149

Jesse-Bakker:
Partially resolve #455

### DIFF
--- a/lua/orgmode/treesitter/headline.lua
+++ b/lua/orgmode/treesitter/headline.lua
@@ -475,6 +475,41 @@ function Headline:parse(pattern)
   return matching_nodes[1], match
 end
 
+function Headline:get_drawer(name)
+  local section = self.headline:parent()
+  local body = section:field('body')
+  if not body then
+    return nil
+  end
+
+  for node, node_name in body[1]:iter_children() do
+    if node:type() == 'drawer' then
+      local drawer_name = node:field('name')
+      if #drawer_name and string.lower(query.get_node_text(drawer_name[1], 0)) == string.lower(name) then
+        return node
+      end
+    end
+  end
+end
+
+---Return the line number where content can be appended within
+---the drawer with the given name, matched case-insensitively
+---@param name string
+---
+---@return number
+function Headline:get_drawer_append_line(name)
+  local drawer = self:get_drawer(name)
+
+  if not drawer then
+    local append_line = self:get_append_line()
+    local new_drawer = self:_apply_indent({ ':' .. name .. ':', ':END:' })
+    vim.api.nvim_buf_set_lines(0, append_line, append_line, false, new_drawer)
+    drawer = self:refresh():get_drawer(name)
+  end
+  local name_row = drawer:field('name')[1]:end_()
+  return name_row + 1
+end
+
 ---@param type string | "DEADLINE" | "SCHEDULED" | "CLOSED"
 ---@return Date|nil
 function Headline:_get_date(type)
--- a/DOCS.md
+++ b/DOCS.md
@@ -241,6 +241,13 @@ Possible values:
 * `note` - adds `CLOSED` date as above, and prompts for closing note via capture window. Confirm note with `org_note_finalize` (Default `<C-c>`), or ignore providing note via `org_note_kill` (Default `<Leader>ok`)
 * `nil|false` - Disable any logging
 
+#### **org_log_into_drawer**
+*type*: `string|nil`<br />
+*default value*: `nil`<br />
+Possible values:
+Log TODO state changes into a drawer with the given name. The recommended value is `LOGBOOK`.
+If `nil`, log into the section body.
+
 #### **org_highlight_latex_and_related**
 *type*: `string|nil`<br />
 *default value*: `nil`<br />
--- a/lua/orgmode/config/defaults.lua
+++ b/lua/orgmode/config/defaults.lua
@@ -30,6 +30,7 @@ local DefaultConfig = {
   org_hide_emphasis_markers = false,
   org_ellipsis = '...',
   org_log_done = 'time',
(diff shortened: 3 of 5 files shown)


Jesse-Bakker:
The test shows the behavior I deduced from the org mode docs, so could you give me some pointers as to what is not working yet?

jgollenz:
Sure. First off: do you know how to reproduce this in emacs? I was able to set the `org-log-into-drawer` variable, but to actually do logs on a TODO-state change, I had to `C-u C-c C-t`, not just a `C-c C-t` (which is the equivalent of our `cit`). So at least I wasn't able to log everytime I do a simple TODO-state change, that's something we should discuss. Perhaps add something like `ucit` or `uct` :thinking: 

The other thing is, that for some reason the TODO-state doesn't actually change with your PR. It stays at `TODO`, but should change to `DONE`. That's why you get the same log twice, when it should actually be something like
```
'  - State "DONE" from "TODO" [' .. Date.now():to_string() .. ']',
'  - State "" from "DONE" [' .. Date.now():to_string() .. ']',
```

And I am only able to get a log into `LOGBOOK` when there is a timestamp on that headline. Emacs also logs on plain headlines. I guess you were going for habits here?

Jesse-Bakker:
From reproducing in emacs, our current behavior matches that in emacs. For non-repeating TODO, nothing is logged, for repeating it is logged and for repeating with org-log-into-drawer set, it logs into the given drawer. For non-repeating, the state is changed, for repeating it is not, just as in emacs.

lua/orgmode/treesitter/headline.lua
```diff
@@ -475,6 +475,28 @@ function Headline:parse(pattern)
...
 end
 
+---Return the line number where content can be appended within
+---the drawer with the given name
+---@param name string
+---
+---@return number
+function Headline:get_drawer_append_line(name)
+  local section = self.headline:parent()
+  local q = tree_utils.parse_query('(body (drawer) @drawer)')
```
kristijanhusak, line 485: Let's extract this part to a separate method called `Headline:get_drawer(name)`, which would find the drawer node by the given name, case-insensitively.  Then here you can find the drawer with that method and return the appropriate line number. You can closely mimic how `properties()` and `set_property` methods work.

kristijanhusak (changes requested):
This looks very good, thanks!
I think we should just slightly refactor the headline methods so they are more reusable/readable.

lua/orgmode/treesitter/headline.lua
Jesse-Bakker: Something like this?

lua/orgmode/treesitter/headline.lua
```diff
@@ -475,6 +475,37 @@ function Headline:parse(pattern)
   return matching_nodes[1], match
 end
 
+function Headline:get_drawer(name)
+  local section = self.headline:parent()
+  local q = tree_utils.parse_query('(body (drawer) @drawer)')
```
kristijanhusak, line 480: I think drawer is always in the first level of the headline, so you can probably just iterate the named children like we do when finding property drawer. Parsing queries is usually slow for the first time until it gets cached, so we try to avoid them where not necessary.

Jesse-Bakker: Right now, it's below `body`, so one level down, but should not be difficult to find by iterating

jgollenz:
@Jesse-Bakker thanks for the clarification, the thing I was missing was how the repeating date actually affects the expected behavior :+1:

Jesse-Bakker:
> @Jesse-Bakker thanks for the clarification, the thing I was missing was how the repeating date actually affects the expected behavior :+1:
> 
> 

No problem. I think it would be nice to actually have a comparative test suite that does the same operations in both neovim and emacs and compares the results. I might look into this a little bit "soon"(tm)

jgollenz:
YES, I also thought about that recently. There probably is a test suite that the orgmode/emacs team uses. Ideally we take that and can use it as a measure of feature-completeness as well.

lua/orgmode/treesitter/headline.lua
```diff
@@ -475,6 +475,38 @@ function Headline:parse(pattern)
   return matching_nodes[1], match
 end
 
+function Headline:get_drawer(name)
+  local section = self.headline:parent()
+  local body = section:field("body")[1]
+
+  for node, node_name in body:iter_children() do
```
kristijanhusak, line 482: I think we should check if `body` is not nil here, since that is also possible.

lua/orgmode/treesitter/headline.lua
```diff
@@ -475,6 +475,43 @@ function Headline:parse(pattern)
   return matching_nodes[1], match
 end
 
+function Headline:get_drawer(name)
+  local section = self.headline:parent()
+  local body = section:field('body')
+  if body and #body then
```
kristijanhusak, line 481: I think something like this would be a bit more readable:
```suggestion
  if not body then
    return nil
  end
  body = body[1]
```

#body is always truthy in lua, no matter if it's 0 or not

Jesse-Bakker: TIL, thanks

kristijanhusak (approved):
Looks good, thanks!

PR #533, opened 2023-03-24
No shada during tests

Jesse-Bakker:
Having shada enabled sometimes breaks when running the tests from within a neovim session. Also, shada would get clobbered with synthetic test files. Therefore, disable it during testing.


kristijanhusak:
Thanks!

PR #534, opened 2023-03-25
fix: update treesitter functions

danilshvalov:
See neovim/neovim#22761 (maybe we should wait for the 0.9 release)

### DIFF
--- a/lua/orgmode/org/indent.lua
+++ b/lua/orgmode/org/indent.lua
@@ -1,3 +1,4 @@
+local ts = require('orgmode.treesitter.compat')
 local config = require('orgmode.config')
 local ts_utils = require('nvim-treesitter.ts_utils')
 local query = nil
@@ -25,7 +26,7 @@ local get_matches = ts_utils.memoize_by_buf_tick(function(bufnr)
       }
 
       if type == 'headline' then
-        opts.stars = vim.treesitter.get_node_text(node:field('stars')[1], bufnr):len()
+        opts.stars = ts.get_node_text(node:field('stars')[1], bufnr):len()
         opts.indent = opts.indent + opts.stars + 1
         matches[range.start.line + 1] = opts
       end
@@ -52,7 +53,7 @@ local get_matches = ts_utils.memoize_by_buf_tick(function(bufnr)
         end
         if parent then
           local headline = parent:named_child('headline')
-          local stars = vim.treesitter.get_node_text(headline:field('stars')[1], bufnr):len()
+          local stars = ts.get_node_text(headline:field('stars')[1], bufnr):len()
           opts.indent = stars + 1
           for i = range.start.line, range['end'].line - 1 do
             matches[i + 1] = opts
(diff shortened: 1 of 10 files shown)


jgollenz:
> maybe we should wait for the 0.9 release

That depends on whether we want to ... um ... depend on it :grin: Sooner or later we will, but I let @kristijanhusak decide that.

mcepl (approved):
Works for me (and fixes https://github.com/nvim-orgmode/orgmode/issues/536 for me).

jgollenz:
@danilshvalov do you mind squashing your commits? I'm not sure why the tests are not re-running after I changed the target branch, but a force push from you should do the trick.

kristijanhusak:
We are usually trying to support two latest versions + nightly, which will be 0.8.3, 0.9 and nightly, once the 0.9 is released. I'm ok with this, but we should probably make some wrapper around these and use appropriate one depending on the Neovim version.

jgollenz:
@danilshvalov in this case you can set the base branch back to `master`

jgollenz:
On `0.8.3` I also need a shim for `set_query` (used in `todo_highlighter.lua`) and use `ts.get_query` in `markup_highlighter.lua` and `todo_highlighter.lua` in order to not get errors.
On `nightly` I get a notice about deprecation of `vim.treesitter.query.get_query()` upon every startup. Can we avoid that?

jgollenz (dismissed):
- On `0.8.[0|3]` I also need 
	- a shim for `set_query` (used in `todo_highlighter.lua:55`) 
	- use `ts.get_qery` in `markup_highlighter.lua:230` 
	- use `ts.get_query_files` in `todo_highlighter.lua:8`

  in order to not get errors.
- On `nightly` I get a notice about deprecation of `vim.treesitter.query.get_query()` upon every startup. Can we avoid that?

danilshvalov:
> * On `0.8.[0|3]` I also need
>   
>   * a shim for `set_query` (used in `todo_highlighter.lua:55`)
>   * use `ts.get_qery` in `markup_highlighter.lua:230`
>   * use `ts.get_query_files` in `todo_highlighter.lua:8`

Done

>   in order to not get errors.
> * On `nightly` I get a notice about deprecation of `vim.treesitter.query.get_query()` upon every startup. Can we avoid that?

I can't reproduce it. Could you provide a log please?

danilshvalov:
@jgollenz, is there anything else?

lua/orgmode/org/indent.lua
```diff
@@ -25,7 +25,7 @@ local get_matches = ts_utils.memoize_by_buf_tick(function(bufnr)
       }
 
       if type == 'headline' then
-        opts.stars = vim.treesitter.query.get_node_text(node:field('stars')[1], bufnr):len()
+        opts.stars = vim.treesitter.get_node_text(node:field('stars')[1], bufnr):len()
```
kristijanhusak, line 28: I think we need to use compat here and below in whole file

danilshvalov: Done

jgollenz:
@danilshvalov when I run the `minimal_init` with your branch I don't get the error, so it looks like it's just on my side. I'll investigate further, but let's merge this in the meantime :+1:

PR #541, opened 2023-03-29
fix(calendar): remove key waiting

danilshvalov:
This patch removes the input lag when the `<` or `>` key is pressed.

#### Before

https://user-images.githubusercontent.com/57654917/228636125-b7a70556-da2d-4297-a64f-0d924100af7f.mov

#### After

https://user-images.githubusercontent.com/57654917/228636148-145de163-562b-4ae4-a3d6-91d46d4ccf2f.mov


kristijanhusak:
Thanks!

PR #542, opened 2023-03-30
feat(calendar): date jump

danilshvalov:
This patch adds the ability to enter the date manually. It's a bit like how it's done in [emacs orgmode](https://orgmode.org/manual/The-date_002ftime-prompt.html):

image

#### Example
https://user-images.githubusercontent.com/57654917/228889210-439302ad-fa5c-4772-9f0d-7c25b74c3aee.mov


jgollenz:
Thanks a lot for the PR :slightly_smiling_face: I think this is a good start for the date/time prompt from which we can expand upon :+1:

Issue #490, opened 2023-01-22
Changing status in todo list causes unhandled promise rejection 


szebenyib:
### Describe the bug

Error executing vim.schedule lua callback: unhandled promise rejection: { ".../packer/start/orgmode/lua/orgmode/agenda/views/todos.lua:85: attempt to perform arithmetic on local 'win_width' (a nil value)" }
stack traceback:
        [C]: in function 'error'
        .../pack/packer/start/orgmode/lua/orgmode/utils/promise.lua:68: in function <.../pack/packer/start/orgmode/lua/orgmode/utils/promise.lua:66>
Press ENTER or type command to continue

### Steps to reproduce

1, Open neovim
2, Open agenda, list of todos
3, Change the status of a todo item
4, Right away or only after a few change toggles the error message appears, the todo view does not refresh

### Expected behavior

No error, changing of status should be visible.

### Minimal init.lua

vim.cmd('source ' .. '~/.vimrc')
local ensure_packer = function()
	local fn = vim.fn
	local install_path = fn.stdpath('data')..'/site/pack/packer/start/packer.nvim'
	if fn.empty(fn.glob(install_path)) > 0 then
		fn.system({'git', 'clone', '--depth', '1', 'https://github.com/wbthomason/packer.nvim', install_path})
		vim.cmd [[packadd packer.nvim]]
		return true
	end
	return false
end

local packer_bootstrap = ensure_packer()

return require('packer').startup(function(use)
	use 'wbthomason/packer.nvim'
	-- My plugins here
	-- use 'foo1/bar1.nvim'
	-- use 'foo2/bar2.nvim'
	use {'nvim-treesitter/nvim-treesitter'}
	use {'nvim-orgmode/orgmode', config = function()
		require('orgmode').setup{}
	end
	}

	-- Load custom treesitter grammar for org filetype
	require('orgmode').setup_ts_grammar()

	-- Treesitter configuration
	require('nvim-treesitter.configs').setup {
		-- If TS highlights are not enabled at all, or disabled via `disable` prop,
		-- highlighting will fallback to default Vim syntax highlighting
		highlight = {
			enable = true,
			-- Required for spellcheck, some LaTex highlights and
			-- code block highlights that do not have ts grammar
			additional_vim_regex_highlighting = {'org'},
		},
		ensure_installed = {'org'}, -- Or run :TSUpdate org
	}

	require('orgmode').setup({
		org_agenda_files = {'~/sync/orgzly/*'},
[...]

PR #543, opened 2023-04-06
Fix promisify error in AgendaTodosView when changing todo state

tobymelin:
I've been having the same issue as described in #490 for a while and finally had a chance to see if I could figure out the cause of it.

The view update fails every time I change the status of a todo item from AgendaTodosView. From what I can tell it seems to be caused by `win_width` not being set when calling `AgendaTodosView.generate_todo_item` from `Agenda:_remote_edit`, which leads to all sorts of weirdness in the TodosView when this call fails.


kristijanhusak:
Thanks!

PR #546, opened 2023-04-20
fix: missing nil check in open_at_point()

troiganto:
Without the check, attempting to open a link like `[[file:./file.org::*Heading]]` outside of any heading would cause a Lua error.

I'm not super proficient in Lua dev, so if there's any unit test I should add for this, I'd need some guidance.

In any case, thank you for working on this great plugin! :slightly_smiling_face:


jgollenz:
Looks good to me, thanks!

PR #547, opened 2023-04-21
Force-disable spellcheck in certain markup groups.

troiganto:
This passes the `spell` option to `nvim_buf_set_extmark()` for all markup groups. In most cases, the passed value is simply `nil`, i.e. the same as if it were not passed at all.

In the following cases, however, we pass `spell=false`, i.e. forcefully disable spell-checking:

- the URL part of `[[URL][hyperlinks]]`;
- the inside of `[[hyperlinks]]` without description;
- `~code~` and `=verbatim=` spans;
- in-line LaTeX markup using `\(parentheses\)`.

The reason in all cases is that these spans are extremely likely to contain purposefully non-orthographic spelling, in particular hyperlinks.

### DIFF
--- a/lua/orgmode/colors/markup_highlighter.lua
+++ b/lua/orgmode/colors/markup_highlighter.lua
@@ -36,18 +36,21 @@ local markers = {
     hl_name = 'org_code',
     hl_cmd = 'hi def link org_code String',
     nestable = false,
+    spell = false,
     type = 'text',
   },
   ['='] = {
     hl_name = 'org_verbatim',
     hl_cmd = 'hi def link org_verbatim String',
     nestable = false,
+    spell = false,
     type = 'text',
   },
   ['\\('] = {
     hl_name = 'org_latex',
     hl_cmd = 'hi def link org_latex OrgTSLatex',
     nestable = false,
+    spell = false,
     type = 'latex',
   },
   ['\\{'] = {
@@ -381,6 +384,7 @@ local function apply(namespace, bufnr, line_index)
       ephemeral = true,
       end_col = range.to['end'].character,
       hl_group = markers[range.type].hl_name,
+      spell = markers[range.type].spell,
       priority = 110 + range.from.start.character,
     })
 
@@ -402,6 +406,7 @@ local function apply(namespace, bufnr, line_index)
     local line = vim.api.nvim_buf_get_lines(bufnr, link_range.from.start.line, link_range.from.start.line + 1, false)[1]
     local link = line:sub(link_range.from.start.character + 1, link_range.to['end'].character)
     local alias = link:find('%]%[') or 1
+    local link_end = link:find('%]%[') or (link:len() - 1)
 
     vim.api.nvim_buf_set_extmark(bufnr, namespace, link_range.from.start.line, link_range.from.start.character, {
       ephemeral = true,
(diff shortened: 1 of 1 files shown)


troiganto:
Hmm, something is still strange about this. The change in this PR does indeed remove highlighting for bad words, but the keymaps `]s` and `[s` (go to next/prev misspelled word) _still_ jump to these now-invisible errors. I'll investigate more tomorrow, if I find the time. There might be a bug in how extmarks handle spelling. The feature is still rather new, if I remember correctly.

lua/orgmode/colors/markup_highlighter.lua
```diff
@@ -12,54 +12,63 @@ local markers = {
...
     hl_cmd = 'hi def org_italic term=italic cterm=italic gui=italic',
     nestable = true,
+    spell = nil,
     type = 'text',
   },
   ['_'] = {
     hl_name = 'org_underline',
     hl_cmd = 'hi def org_underline term=underline cterm=underline gui=underline',
     nestable = true,
+    spell = nil,
```
kristijanhusak, line 29: I believe all `spell = nil` can be removed, since it will default to `nil` when it tries to find something on table that does not exist.

kristijanhusak:
I can confirm that this works, but has that issue you mentioned that it jumps to the works that should not have spell checking. Do you maybe have any news on that?
We could also expand this and move old `syntax spell toplevel` from `syntax/org.vim` to `syntax/org_legacy.vim` and define specific captures to have the spell checking, like `(body (paragraph) @spell)`, etc.

troiganto:
Sorry, this week has been pretty busy. I only found time to sit down and look into it tonight.

As far as I can tell, this buggy behavior is universal for neovim extmarks. I've opened an issue about this with them (see the mention above), let's see what comes out of that.

Regarding a workaround on the orgmode side, I'm not quite sure how to do that. Using captures sounds promising, but my (extremely limited) understanding is that the TS grammar doesn't currently track inline markup. See this screenshot of the `:InspectTree` output (requires nvim 0.9.0+):

grafik

Naively, I would've expected a TS nodes like "code" , but instead everything is just "expr" and the actual highlighting is done with extmarks. With this set up, I don't know if it's _possible_ to define specific captures for spell checking.

kristijanhusak:
Don't worry, this is open source, no one expects you to be around on demand :)

It seems there is a draft PR that fixes the issue so I hope it will cover this.
Do you know when they added support for `spell` on extmarks? I wasn't strictly following.

I'll double check this PR and merge it if it's ok. Thanks!

I think you misunderstood what I meant. I meant to remove the old vim syntax spell checking and introduce it on the treesitter level in the queries, something like this https://github.com/nvim-treesitter/nvim-treesitter/blob/master/queries/ecma/highlights.scm#L179.

This can be done separately, so don't worry about it.

troiganto:
I had to dig a little, but I think `spell=false` was introduced in neovim/neovim#20178. (It's also mentioned in the [release notes for nvim 0.9.0](https://github.com/neovim/neovim/releases/tag/v0.9.0).) The same PR apparently also introduced `@nospell` captures, but I couldn't make those work in the orgmode queries. Not terribly surprising though, since I'm still very new to the treesitter ecosystem.

In any case, I've removed the `spell = nil` lines as requested. Let me know if there's anything else, and thanks for the feedback!

kristijanhusak (approved):
Looks good, thanks!

kristijanhusak:
Just a note that I added support for tree-sitter level spell checking here https://github.com/nvim-orgmode/orgmode/commit/4647d20e33bf14c70fd44e1a4ee071ce1351d139, so spell checking should generally be better than before.

troiganto:
Very cool! I'll give it a try and report bugs if I find any :+1:

PR #549, opened 2023-04-29
feat: modern menu

danilshvalov:
Hello everyone I don't really like the way the selection menu is made now. Therefore, I suggest updating it. 

Now I have the following developments:

Screenshot 2023-04-29 at 13 47 55


kristijanhusak:
Hey, I don't plan to change the current built in view because it looks similar to Emacs prompt. There are 2 options this can be tackled:
* Add a config option to define your own menu, and default it to current `utils.menu`. That way you can override the functionality in your config
* Monkey patch `orgmode.utils.menu` with your implementation. This is more hacky way, but I think it could work.

danilshvalov:
<details>
<summary>Emacs style</summary>

```lua
require("orgmode").setup({
  ui = {
    menu = {
      preset = "emacs",
    },
  },
})
```

Screenshot 2023-05-01 at 18 45 28

<details>
<summary>Modern style</summary>

```lua
require("orgmode").setup({
  ui = {
    menu = {
      preset = "modern",
    },
  },
})
```
Screenshot 2023-05-01 at 18 45 15
</details>

<details>
<summary>Custom style</summary>

```lua
require("orgmode").setup({
  ui = {
    menu = {
      custom = function(title)
        vim.print("Just print title...")
        vim.print("Title is '" .. title .. "'")
      end,
    },
  },
})
```

Example:
Screenshot 2023-05-01 at 18 46 54
</details>

jgollenz:
Hm, I think this opens up a larger question of UI changes and where they belong. Could this (and any UI specifics) be made accessible via the API and enable @danilshvalov and everybody else to write their own UI overhaul of the whole plugin?

danilshvalov:
@jgollenz If I understood what was written correctly, right now it is impossible to customize the UI through the API. This PR provides one of the ways of such customization (the `custom` option).

kristijanhusak (changes requested):
Thanks for the PR!

The general idea is fine, but I don't think there's a need to have presets part of the plugin.
We can either have the default one, or a custom one that the user provides.
Comments on code will explain what's missing.
If you want to expose the modern menu to more people, I would suggest creating a small plugin and just attach it through the config.

Edit: It would also be great to update the documentation with this new option so other users know how to set up a custom menu.

danilshvalov:
I have redone the code, created some classes that internally check the correctness of the entered data. So now there are two types that can be used in the menu: `MenuItem` and `MenuSeparator`. 

If everything is fine in the code, I am ready to add documentation.

kristijanhusak (approved):
Looks good, you can proceed with docs.

danilshvalov:
I now thought a little and decided that it would be nice to pass data to the menu not through function arguments, but through a table. For example:

```lua
Menu.open({
  title = 'Some title',
  items = { ... },
  prompt = 'Prompt',
})
```

What do you think about it? In the future, with possible expansion, this will make refactoring easier.

jgollenz:
> pass data to the menu not through function arguments, but through a table

I second this, but please let's document that table somewhere :+1:

kristijanhusak:
> I now thought a little and decided that it would be nice to pass data to the menu not through function arguments, but through a table. For example:
> 
> ```lua
> Menu.open({
>   title = 'Some title',
>   items = { ... },
>   prompt = 'Prompt',
> })
> ```
> 
> What do you think about it? In the future, with possible expansion, this will make refactoring easier.

I agree with this. Additionally, I would also expose `option` and `separator` methods on the menu so the menu can be built. Something like this:

```lua
local menu = Menu:new({ 
  title = 'Some title',
  items = { ... }, -- optional
  prompt = 'Prompt',
  separator = '-',
})

menu:option(label, key, action)
menu:separator(length, char_override)
menu:option(label, key, action)

return menu:open()
```

danilshvalov:
@kristijanhusak This is a great idea! But maybe it would be better to name the methods `add_option` and `add_separator` instead of `option` and `separator` for more clarity? For example, `separator` might appear to change the value of the separator itself rather than adding it to the menu.

danilshvalov:
And it's probably better to pass a delimiter to the Menu constructor like this:
```lua
local menu = Menu:new({
  -- ...
  separator = {
    icon = '~', -- optional, default is `-`
    length = 1, -- optional, default is 80
  },
})
```

kristijanhusak:
> @kristijanhusak This is a great idea! But maybe it would be better to name the methods `add_option` and `add_separator` instead of `option` and `separator` for more clarity? For example, `separator` might appear to change the value of the separator itself rather than adding it to the menu.

Sure, that also works.


> And it's probably better to pass a delimiter to the Menu constructor like this:
> 
> ```lua
> local menu = Menu:new({
>   -- ...
>   separator = {
>     icon = '~', -- optional, default is `-`
>     length = 1, -- optional, default is 80
>   },
> })
> ```

This is also ok. `add_separator` method should accept both of these options as override, `length` and then `icon`, because we have situations where we have different length separators in the menu.

danilshvalov:
I did what we discussed. Review please.

lua/orgmode/agenda/init.lua
```diff
@@ -88,45 +89,50 @@ end
...
-    },
-    { label = 'Quit', key = 'q' },
-    { label = '', separator = ' ', length = 1 },
-  }, 'Press key for an agenda command')
+  local menu = Menu:new({
+    title = 'Press key for an agenda command',
+    prompt = 'Press key for an agenda command',
+  })
+
+  menu:add_option({
```
kristijanhusak, line 97: I think we can go with multiple arguments here instead of single table. I don't expect this to grow.
```lua
menu:add_option('Agenda for current week or day', 'a', function() 
  return self:agenda() 
end)
```

Same thing for separators.

lua/orgmode/capture/init.lua
```diff
@@ -54,12 +55,15 @@ function Capture:_create_menu_items(templates)
...
-  table.insert(menu_items, { label = 'Quit', key = 'q' })
-  table.insert(menu_items, { label = '', separator = ' ', length = 1 })
-
-  return utils.menu('Select a capture template', menu_items, 'Template key')
+  local menu = Menu:new({
+    title = 'Select a capture template',
+    items = self:_create_menu_items(templates),
+    prompt = 'Template key',
+  })
+  menu:add_separator({ icon = '-' })
```
kristijanhusak, line 63: We don't need to provide `icon` here since `-` is the default.

kristijanhusak (changes requested):
Looks good, just few small changes regarding the arguments.

lua/orgmode/agenda/init.lua
danilshvalov: I think tables improve readability. If I read the code for the first time, I would not understand what the first and second lines are responsible for.

kristijanhusak: I agree it does. Generally, I also prefer tables when the number of args goes over 3 or if more than 1 is optional. Here we must provide the first two, or even all of them.
Also, documenting them via annotations is very helpful for DX.

Since this is a matter of preference, I'll let @jgollenz give his vote on this and we'll go from there.

jgollenz: Since `MenuOption` has proper annotations, I'm in favor of the tables. I agree, it does make things a bit clearer

danilshvalov:
Ready for review

kristijanhusak (approved):
Looks great, thanks!

Issue #553, opened 2023-05-03
org-return and telescope


danilshvalov:
### Describe the bug

Error occurs when opening org file through `telescope.nvim` and pressing `<CR>` in insert mode (see video example).

### Steps to reproduce

This is a terribly hard to reproduce bug. So far I have reduced it to the following steps:
1. Open neovim
2. Open `Telescope oldfiles` (I do it with `SPC f r`)
3. Open org file via telescope
4. Press `<CR>` in insert mode

### Expected behavior

An error should not occur.

### Minimal init.lua

```lua
local root = vim.fn.fnamemodify("./.repro", ":p")

-- set stdpaths to use .repro
for _, name in ipairs({ "config", "data", "state", "cache" }) do
  vim.env[("XDG_%s_HOME"):format(name:upper())] = root .. "/" .. name
end

-- bootstrap lazy
local lazypath = root .. "/plugins/lazy.nvim"
if not vim.loop.fs_stat(lazypath) then
  vim.fn.system({
    "git",
    "clone",
    "--filter=blob:none",
    "--single-branch",
    "https://github.com/folke/lazy.nvim.git",
    lazypath,
  })
end
vim.opt.runtimepath:prepend(lazypath)

local plugins = {
  {
    "nvim-telescope/telescope.nvim",
    dependencies = "nvim-lua/plenary.nvim",
    config = function()
      vim.keymap.set("n", "<leader>fr", "<Cmd>Telescope oldfiles<CR>")
      vim.keymap.set("n", "<Space>", "<leader>", { remap = true })
    end,
  },
  {
    "nvim-orgmode/orgmode",
    build = ":TSUpdate",
    dependencies = "nvim-treesitter/nvim-treesitter",
    config = function()
      require("orgmode").setup_ts_grammar()
      require("nvim-treesitter.configs").setup({
        highlight = {
          enable = true,
          additional_vim_regex_highlighting = { "org" },
        },
        ensure_installed = { "org" },
      })
      require("orgmode").setup({})
    end,
  },
}
require("lazy").setup(plugins, {
  root = root .. "/plugins",
})

vim.opt.termguicolors = true
```

### Screenshots and recordings

I was only able to reproduce the bug on the first and last try on this video

https://user-images.githubusercontent.com/57654917/235885439-1736c241-8f06-487e-a57a-01e791a21b71.mp4

### OS / Distro

macOS Ventura 13.2

### Neovim version/commit

0.9

[...]

PR #554, opened 2023-05-03
fix(mappings): properly get old mapping

danilshvalov:
I played around a bit with the problem described in #553. It seems to me that the point is that we are calling the wrong `old_cr_mapping`. I replaced `vim.fn.maparg` with `vim.api.nvim_get_keymap` and the problem went away.

### DIFF
--- a/lua/orgmode/config/init.lua
+++ b/lua/orgmode/config/init.lua
@@ -260,9 +260,13 @@ function Config:get_todo_keywords()
   return types
 end
 
-function Config:setup_mappings(category, bufnr)
+function Config:setup_mappings(category, buffer)
   if not self.old_cr_mapping then
-    self.old_cr_mapping = vim.fn.maparg('<CR>', 'i', false, true)
+    self.old_cr_mapping = utils.get_keymap({
+      mode = 'i',
+      lhs = '<CR>',
+      buffer = buffer,
+    })
   end
   if self.opts.mappings.disable_all then
     return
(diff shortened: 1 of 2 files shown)


lua/orgmode/utils/init.lua
```diff
@@ -167,6 +167,27 @@ function utils.menu(title, items, prompt)
   return entry.action()
 end
 
+---@class KeymapData
+---@field mode string|table
+---@field lhs string
+---@field buffer integer?
+
+---@param data KeymapData
```
jgollenz, line 175: Please add a `---@returns rhs string` annotation here

lua/orgmode/config/init.lua
```diff
@@ -260,9 +260,13 @@ function Config:get_todo_keywords()
   return types
 end
 
-function Config:setup_mappings(category, bufnr)
+function Config:setup_mappings(category, buffer)
```
jgollenz, line 263: While we're at it, please annotate the function. Thanks!

jgollenz (changes requested):
Looks good to me, thank you! Just some annotations and good to merge :+1:

danilshvalov:
Done

jgollenz (approved):

kristijanhusak:
I think we will need to expand on this a bit more. This functionality now always tries to grab the buffer level mapping, but some plugins that map <CR> actually map it globally, so we will never have an "old_mapping" here in those case.
Code should do this:
* Try to grab buffer level mapping
* If there isn't any, try to grab global level mapping

kristijanhusak:
Note that I added some additional logic around this to make sure it grabs both local and global mapping and I also moved it to store the mapping per buffer level https://github.com/nvim-orgmode/orgmode/commit/07d412eb15155c8391945a5bbbfd1caa996888bc

Issue #447, opened 2022-11-14
org fold does not work when `.org` file is opened by `telescope`


milanglacier:
### Describe the bug

org fold does not work when `.org` file is opened by `telescope`

i.e

keybindings like `zM`, `zR`, `<Tab>` do not work.

However, it is noted that if after opening the `.org` buffer by telescope, typing `:e` to refresh the buffer content will make the fold works

### Steps to reproduce

1. Open an org file via `:Telescope find_files` or `:Telescope oldfiles`
2. go to a heading and typing `<tab>`
3. find that folding does not work

### Expected behavior

Folding should work

### Emacs functionality

folding should work

### Minimal init.lua

```lua

vim.cmd [[set runtimepath=$VIMRUNTIME]]
vim.cmd [[set packpath=/tmp/nvim/site]]

local package_root = '/tmp/nvim/site/pack'
local install_path = package_root .. '/packer/start/packer.nvim'

local function load_plugins()
    require('packer').startup {
        {
            'wbthomason/packer.nvim',
            { 'nvim-treesitter/nvim-treesitter' },
            { 'kristijanhusak/orgmode.nvim', branch = 'master' },
            { 'nvim-telescope/telescope.nvim' },
            { 'nvim-lua/plenary.nvim' },
        },
        config = {
            package_root = package_root,
            compile_path = install_path .. '/plugin/packer_compiled.lua',
        },
    }
end

_G.load_config = function()
    require('orgmode').setup_ts_grammar()
    require('nvim-treesitter.configs').setup {
        highlight = {
            enable = true,
            additional_vim_regex_highlighting = { 'org' },
        },
    }

    vim.cmd [[packadd nvim-treesitter]]
    vim.cmd [[runtime plugin/nvim-treesitter.lua]]
    vim.cmd [[TSUpdateSync org]]

    -- Close packer after install
    if vim.bo.filetype == 'packer' then
        vim.api.nvim_win_close(0, true)
    end

    require('orgmode').setup()
    require('telescope').setup()

    -- the following line should be commented out
    -- because :e reloads the buffer which makes the fold work
    -- however if a buffer is directly opened by telescope
    -- then folds does not work

    -- Reload current file if it's org file to reload tree-sitter
    -- if vim.bo.filetype == 'org' then
    --     vim.cmd [[edit!]]
    -- end
end

if vim.fn.isdirectory(install_path) == 0 then
[...]

PR #558, opened 2023-05-09
Revert "Remove checks for loaded ftplugin and indent"

milanglacier:
Use `did_ftplugin` for early return is a standard practice in vim. The commit that introduced this change was aiming to solve a treesitter bug which should be outside of orgmode's scope (see #447).

At the time #447 was issued, I was thinking about this is a telescope bug, but indeed this is a treesitter's bug, see https://github.com/nvim-treesitter/nvim-treesitter/issues/4754.

And indeed this commit only solves #447 partially. The first time you opened an org file via telescope, the fold works correctly, but the second time you opened a file via telescope, the fold does not work again.

Based on above description, I think we should revert the changes and follow the standard practice of vim's ftplugin. And let the bugs to be fixed in upstream.

### DIFF
--- a/ftplugin/org.vim
+++ b/ftplugin/org.vim
@@ -1,3 +1,8 @@
+if exists('b:did_ftplugin')
+  finish
+endif
+let b:did_ftplugin = 1
+
 lua require('orgmode.config'):setup_mappings('org')
 lua require('orgmode.config'):setup_mappings('text_objects')
 
--- a/indent/org.vim
+++ b/indent/org.vim
@@ -1,3 +1,8 @@
+if exists('b:did_indent')
+  finish
+endif
+let b:did_indent = 1
+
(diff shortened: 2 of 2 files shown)


kristijanhusak:
Thanks for investigating this, it's a very interesting issue.

Do you have an idea what's actually causing the bug in nvim-treesitter?

milanglacier:
> Thanks for investigating this, it's a very interesting issue.
> 
> Do you have an idea what's actually causing the bug in nvim-treesitter?

The fold method of treesitter uses cache to store fold information (fold level at each line).

And I suppose the autocmds that triggers treesitter to recalculate the fold are not activated in insert mode file switching.

so when you are switching files at insert mode, the cache are not updated.

kristijanhusak:
Thanks!

PR #561, opened 2023-05-10
feat(ui): virtual indent

danilshvalov:
### Preview

https://github.com/nvim-orgmode/orgmode/assets/57654917/b8429fba-bf52-4cfc-8d04-7243418e8099

### How to try it

1. Install neovim with neovim/neovim#23426

2. Add the following code to your `orgmode` configuration:
```lua
local VirtualIndent = require("orgmode.ui.virtual-indent"):new()
VirtualIndent:attach()
```

### TODO

- [x] Customizable indentation size (hidden or visible stars, possibly custom indentation)
- [x] Documentation

### DIFF
--- a/lua/orgmode/config/defaults.lua
+++ b/lua/orgmode/config/defaults.lua
@@ -175,6 +175,9 @@ local DefaultConfig = {
     menu = {
       handler = nil,
     },
+    virtual_indent = {
+      handler = nil,
+    },
   },
 }
 
--- a/lua/orgmode/org/indent.lua
+++ b/lua/orgmode/org/indent.lua
@@ -1,6 +1,7 @@
 local ts = require('orgmode.treesitter.compat')
 local config = require('orgmode.config')
 local ts_utils = require('nvim-treesitter.ts_utils')
+local VirtualIndent = require('orgmode.ui.virtual-indent')
 local query = nil
 
 local get_matches = ts_utils.memoize_by_buf_tick(function(bufnr)
@@ -194,8 +195,15 @@ local function foldtext()
   return line .. config.org_ellipsis
 end
 
+local function setup()
+  if config.org_indent_mode == 'virtual_indent' then
+    VirtualIndent:new(config.ui.virtual_indent):attach()
+  end
+end
+
 return {
   foldexpr = foldexpr,
   indentexpr = indentexpr,
   foldtext = foldtext,
+  setup = setup,
 }
--- a/lua/orgmode/ui/virtual-indent.lua
+++ b/lua/orgmode/ui/virtual-indent.lua
@@ -58,8 +58,6 @@ function VirtualIndent:set_indent(start_line, end_line)
         virt_text = { { string.rep(' ', indent) } },
         virt_text_pos = 'inline',
         right_gravity = false,
-        virt_text_hide = true,
-        priority = 1,
       })
     end
   end
--- a/lua/orgmode/ui/virtual-indent.lua
+++ b/lua/orgmode/ui/virtual-indent.lua
@@ -1,0 +1,86 @@
+local Headline = require('orgmode.treesitter.headline')
+
+local VirtualIndent = {}
+
+function VirtualIndent:new(data)
+  data = data or {}
+
+  local default_name = 'orgmode.ui.indent'
+
+  local opts = {}
+  opts.ns_id = vim.api.nvim_create_namespace(data.namespace or default_name)
+  opts.augroup = vim.api.nvim_create_augroup(data.augroup_name or default_name, {})
+
+  setmetatable(opts, self)
+  self.__index = self
+  return opts
+end
+
+function VirtualIndent:_delete_old_extmarks(start_line, end_line)
+  local old_extmarks = vim.api.nvim_buf_get_extmarks(
+    0,
+    self.ns_id,
+    { start_line, 0 },
+    { end_line, 0 },
+    { type = 'virt_text' }
+  )
+  for _, ext in ipairs(old_extmarks) do
+    vim.api.nvim_buf_del_extmark(0, self.ns_id, ext[1])
+  end
+end
+
+function VirtualIndent:_get_indent_size(line)
+  local headline = Headline.from_cursor({ line + 1, 1 })
+  if headline then
+    local level = headline:level()
+    local headline_line, _, _ = headline.headline:start()
+
+    if headline_line == line then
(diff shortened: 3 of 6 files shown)


danilshvalov:
It has happened! Virtual text is merged into the master branch! So I've made some changes, now you can enable virtual indent as follows:
```lua
require("orgmode").setup({
  org_indent_mode = "virtual_indent",
  ui = {
    virtual_indent = {
      handler = nil,
      --- or
      handler = function(buffer, start_line, end_line)
        -- ...
      end,
    },
  },
})
```

danilshvalov:
Ready for review

lua/orgmode/config/defaults.lua
```diff
@@ -175,6 +175,9 @@ local DefaultConfig = {
     menu = {
       handler = nil,
     },
+    virtual_indent = {
```
kristijanhusak, line 178: Why would we need a custom handler for this?

lua/orgmode/org/indent.lua
```diff
@@ -194,8 +195,15 @@ local function foldtext()
   return line .. config.org_ellipsis
 end
 
+local function setup()
+  if config.org_indent_mode == 'virtual_indent' then
```
kristijanhusak, line 199: We need to check if value is `noindent` here.

lua/orgmode/ui/virtual-indent.lua
```diff
@@ -0,0 +1,101 @@
...
+function VirtualIndent:attach(buffer)
+  vim.validate({
+    buffer = { buffer, 'number', true },
+  })
+
+  buffer = buffer or 0
+
+  self:set_indent(buffer, 0, vim.api.nvim_buf_line_count(buffer) - 1)
+
+  vim.api.nvim_buf_attach(buffer, true, {
```
kristijanhusak, line 92: Why do we need `vim.schedule` in the callback?

Also, is there any advantage of using `nvim_buf_attach` over `nvim_set_decoration_provider` ?
We use decoration provider for markup, so we could maybe use the same for this while using the `ephermal` ?

danilshvalov: We need `vim.schedule`, because otherwise, when undoing, the virtual indent handler is called too early and processed incorrectly.

As far as I know, ephemeral virtual text has not been implemented yet.

lua/orgmode/config/defaults.lua
danilshvalov: For example, if the user wants to make virtual indents with indent lines (something like [`indent-blankline.nvim`](https://github.com/lukas-reineke/indent-blankline.nvim)). Or change the size of the indentation. Or make the indentation colored depending on its size or title. It is quite difficult to cover all the options, so the user can provide his own handler.

lua/orgmode/org/indent.lua
danilshvalov: Why? And if the user doesn't want any indentation? I see three options for using indentation:
1. No indentation
2. Virtual indentation
3. Real indentation

lua/orgmode/config/defaults.lua
kristijanhusak: Let's remove this option for the time being. Once the feature is officially released and there is a proper Neovim release with the support, we can expand on it.

lua/orgmode/org/indent.lua
kristijanhusak: No indentation was initially meant to be used as virtual indentation.
If you check in the Emacs docs, there are two values, `indent` and `noindent` (https://orgmode.org/manual/Org-Indent-Mode.html)

Problem is that I set the `indent` as default, where it should actually be noindent. `indent` option is actually a `virtual indent`. I'll fix that and announce it on the breaking change issue so people are aware. Once I do that you will actually have to check for `indent` value.

danilshvalov: Why not have 3 options? Yes, this is somewhat at odds with emacs, but it makes the configuration more flexible.

kristijanhusak: `noindent` option as it is now is not useful at all. Everything is aligned to the left. As far as I know, people are now using it only to get proper formatting with other tools, like Orgzly. What we have now by default will become `noindent`, and this PR will introduce a proper `indent` option. There's no need for 3rd.

lua/orgmode/ui/virtual-indent.lua
kristijanhusak: I gave it a quick test, and it seems to work. Switching to it should reduce some code that does the cleanup. Lets try switching to decoration provider and see if it works out.

lua/orgmode/org/indent.lua
danilshvalov: I know it may be weird, but some people don't use indentation. Some time ago I came across [this discussion](https://www.reddit.com/r/emacs/comments/97naje/what_is_everyones_org_mode_indentation_preferences) and found out that not everyone needs indents (see the first comment).

In addition, if you break backward compatibility with indent, quite a lot of org files can be visually broken. I'm not sure if this is a good idea.

lua/orgmode/ui/virtual-indent.lua
danilshvalov: Is it true? I get `E5108: Error executing lua [string ":lua"]:1: not yet implemented` error.

Here is the code I run:
```lua
vim.api.nvim_buf_set_extmark(0, vim.api.nvim_create_namespace("test"), 0, 1, {virt_text_pos = "inline", virt_text = {{"    "}}, ephemeral = true})
```

My neovim version is `NVIM v0.10.0-dev-410+gc48f94d1f`

danilshvalov: Okay, I get the point. I'll try to redo it

lua/orgmode/org/indent.lua
kristijanhusak: We can allow that functionality by doing the same thing that Emacs does, via `org-adapt-indentation` option. The discussion has a few comments explaining how it works.

milanglacier:
This is one of the most missing UI features of emacs in neovim! I am glad that finally, this is implemented!

kristijanhusak:
Closing this since it was resolved in https://github.com/nvim-orgmode/orgmode/pull/627

treatybreaker:
Thanks @danilshvalov for the initial work 😉

Issue #570, opened 2023-06-03
orgmode hangs up when working with multiple lines of brackets


bridgesense:
### Describe the bug

I use orgmode to handle all of my notes. I've noticed that anytime multiple brackets are displayed on the screen, orgmode locks up. This includes expanding the current time tracking for specific tasks, too.

### Steps to reproduce

Open up an org file. Create a topic. Enter the topic and paste the following:

[BLACK_R:19]
[BLACK_G:10]
[BLACK_B:7]
[BLUE_R:34]
[BLUE_G:74]
[BLUE_B:104]
[GREEN_R:119]
[GREEN_G:117]
[GREEN_B:24]
[CYAN_R:101]
[CYAN_G:111]
[CYAN_B:47]
[RED_R:142]
[RED_G:20]
[RED_B:6]
[MAGENTA_R:161]
[MAGENTA_G:58]
[MAGENTA_B:87]
[BROWN_R:130]
[BROWN_G:85]
[BROWN_B:43]
[LGRAY_R:99]
[LGRAY_G:86]
[LGRAY_B:71]
[DGRAY_R:104]
[DGRAY_G:38]
[DGRAY_B:8]
[LBLUE_R:56]
[LBLUE_G:146]
[LBLUE_B:178]
[LGREEN_R:195]
[LGREEN_G:193]
[LGREEN_B:61]
[LCYAN_R:102]
[LCYAN_G:175]
[LCYAN_B:96]
[LRED_R:213]
[LRED_G:72]
[LRED_B:21]
[LMAGENTA_R:213]
[LMAGENTA_G:21]
[LMAGENTA_B:143]
[YELLOW_R:214]
[YELLOW_G:156]
[YELLOW_B:56]
[WHITE_R:208]
[WHITE_G:205]
[WHITE_B:169]

This can also be reproduced by simply expanding the logbooks for frequently used tasks:
:LOGBOOK:
   CLOCK: [2023-05-29 Mon 16:21]--[2023-05-29 Mon 20:32] => 4:11
   CLOCK: [2023-05-28 Sun 16:04]--[2023-05-28 Sun 16:11] => 0:07
   CLOCK: [2023-05-27 Sat 17:24]--[2023-05-27 Sat 20:34] => 3:10
   CLOCK: [2023-05-25 Thu 15:55]--[2023-05-25 Thu 18:12] => 2:17
[...]

Issue #550, opened 2023-04-30
A few CLOCK in a :LOGBOOK: makes Neovim extremely slow


dodomorandi:
### Describe the bug

Having a relatively high amount of `CLOCK` entries makes Neovim really slow.

### Steps to reproduce

1. Create a new `.org` file
2. Add a heading (i.e. `* Test`)
3. Clock in and clock out
4. Copy paste the `CLOCK` line 20 time (`yy20p`)
5. Try to edit the file furthermore to check for slowdowns (i.e. add a newline with `O`)

### Expected behavior

No slowdowns should be expected (we are not talking about file with thousands of lines), but Neovim at this point is slow performing many simple editing tasks.

### Minimal init.lua

It is just the template, with nothing changed.
```lua
vim.cmd([[set runtimepath=$VIMRUNTIME]])
vim.cmd([[set packpath=/tmp/nvim/site]])

local package_root = '/tmp/nvim/site/pack'
local install_path = package_root .. '/packer/start/packer.nvim'

local function load_plugins()
  require('packer').startup({
    {
      'wbthomason/packer.nvim',
      { 'nvim-treesitter/nvim-treesitter' },
      { 'kristijanhusak/orgmode.nvim', branch = 'master' },
    },
    config = {
      package_root = package_root,
      compile_path = install_path .. '/plugin/packer_compiled.lua',
    },
  })
end

_G.load_config = function()
  require('orgmode').setup_ts_grammar()
  require('nvim-treesitter.configs').setup({
    highlight = {
      enable = true,
      additional_vim_regex_highlighting = { 'org' },
    },
  })

  vim.cmd([[packadd nvim-treesitter]])
  vim.cmd([[runtime plugin/nvim-treesitter.lua]])
  vim.cmd([[TSUpdateSync org]])

  -- Close packer after install
  if vim.bo.filetype == 'packer' then
    vim.api.nvim_win_close(0, true)
  end

  require('orgmode').setup()

  -- Reload current file if it's org file to reload tree-sitter
  if vim.bo.filetype == 'org' then
    vim.cmd([[edit!]])
  end
end

if vim.fn.isdirectory(install_path) == 0 then
  vim.fn.system({ 'git', 'clone', 'https://github.com/wbthomason/packer.nvim', install_path })
  load_plugins()
  require('packer').sync()
  vim.cmd([[autocmd User PackerCompileDone ++once lua load_config()]])
else
  load_plugins()
  load_config()
end
```

[...]

PR #571, opened 2023-06-05
fix: Improve parsing hyperlink like text

kristijanhusak:
This PR should slightly improve performance issues reported in #550 and #570.


cands:
Thanks a lot, this fix made a huge difference on my system. I am using Orgmode to manage habits which produces a lot of state changes with timestamps, Neovim could freeze up for a long time (like 10 s or even more) before, but now such tasks are relatively quick to edit (typically below 0.5 s).

PR #573, opened 2023-06-10
foldlevel should not be zero

WhiteBlackGoose:
Not sure if it's by design or not, but when it's zero it's extremely annoying bc it automatically folds everything once you leave insert mode.

### DIFF
--- a/ftplugin/org.vim
+++ b/ftplugin/org.vim
@@ -24,7 +24,6 @@ setlocal foldmethod=expr
 setlocal foldexpr=nvim_treesitter#foldexpr()
 setlocal foldtext=OrgmodeFoldText()
 setlocal formatexpr=OrgmodeFormatExpr()
-setlocal foldlevel=0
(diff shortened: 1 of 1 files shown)


jgollenz:
The reason why it is set to zero isn't obvious to me either. But why would the foldlevel close your folds :thinking: doesn't happen for me

kristijanhusak:
@WhiteBlackGoose are you maybe using nvim-ufo?

WhiteBlackGoose:
Yeah i am

kristijanhusak:
@WhiteBlackGoose can you check if it happens when you disable it? It should not happen without that plugin. I need to further investigate why it happens with it. I had issue with that plugin while resolving git conflicts via Neovim because it went crazy with folds in the diff files, so I'm suspecting this is more of an issue there.

kristijanhusak:
Closing as duplicate (kind of) of https://github.com/nvim-orgmode/orgmode/issues/599

PR #575, opened 2023-06-11
Open file hyperlinks starting with ~

maidl0ver:
Added a small check for hyperlinks that start with tildes to substitute them with `$HOME` on `<Leader>oo`.


jgollenz:
Looks good, thanks :+1:

PR #577, opened 2023-06-17
fix: Fix executable check in `utils.open`

gw31415:
The previous implementation of `utils.open` did not properly check if the necessary executables were present. In Lua, `0` is treated as `true`, so it always tried to run `xdg-open` even when the executable was not found, preventing the proper command from being executed. This pr updates the check to correctly use the return values of `vim.fn.executable` and `vim.fn.has` functions.


kristijanhusak:
Thanks!

Issue #563, opened 2023-05-26
Open linked files with (system default) external application


Maltimore:
### Does this feature exist in Emacs orgmode core?

Yes

### Orgmode link

https://orgmode.org/manual/Handling-Links.html#index-C_002dc-C_002

see org-open-at-point

### Feature value

It would be good to open (by default with `<Leader>oo`) linked files (like `[[file:/home/maltimore/interesting_file.png]]` with their corresponding application.
I think for now we should just default to opening with the system default programs. I think on linux and macos you can just use `open` (might work on windows as well). So in the case above we'd just have to call `open /home/maltimore/interesting_file.png`.

jgollenz:
Sorry, mis-clicked. Regarding `open`: this doesn't do what you want it to do, at least on my Ubuntu machine. Any chance you are talking about `xdg-open`? I think we could set that as default and have an option to configure it in case you want different program to handle the opening

Maltimore:
Weirdly, `open` works on my KDE-neon (which is 99% ubuntu), and I seem to remember that debian at some point in the recent past  decided to use `open` as well. Maybe I'm misremembering, but it doesn't matter anyways because `xdg-open` works for me as well, so we should just use that :)
Yes, making it configurable is sort of the extended goal I'd say. I'm sure there is a use case where you want your default program to open a certain filetype to be `A`, but if you open it from within org-mode you want it to be `B`, but I think that's already power-user level.

jgollenz:
Oh, I actually meant to make it configurable which binary takes care of opening the link:

`link_handler = xdg-open`

and then just use whatever the user has in `link_handler`. That use case you described is indeed quite elaborate and I'd like to avoid making things more complicated than they are

Maltimore:
Hi, sorry for never responding. I'm going to be too tight with time for the foreseeable future to work on `nvim-orgmode` :(.
Regarding this issue, I just saw it's a duplicate of https://github.com/nvim-orgmode/orgmode/issues/506 so I'm closing here.

Issue #556, opened 2023-05-04
Directory Links


Rimann91:
### Does this feature exist in Emacs orgmode core?

N/A

### Feature value

I do not know if emacs orgmode has this capability. It would be useful to link to directories and open them in the default file explorer in nvim (netwr, NERDTree, NvimTree, etc..). I am using nvim-orgmode to keep notes on bugs I am working on, I often acquire log files from a customer which is usually a folder of many log files. Being able to jump to such a directory and navigate through all of the files in directory would be extremely useful.

avonmoll:
I was able to achieve this behavior for the Yazi file manager by putting the following inside my nvim-orgmode config:

```lua
local LinkFolderType = {}

function LinkFolderType:get_name()
    return 'folder'
end

function LinkFolderType:follow(link)
    if not vim.startswith(link, 'folder:') then
        return false
    end
    local url = link:sub(8)
    vim.cmd('lua require("yazi").yazi(nil, "' .. url .. '")')
    return true
end

require('orgmode').setup({
    ...
    hyperlinks = {
        sources = {
            LinkFolderType,
        },
    },
})
```

Then, in some org file you can write:

```org
[[folder:~/Documents][Document folder]]
```

When I open the link, it opens Yazi inside neovim to that folder. Presumably, this kind of approach can work with any of the file explorer's mentioned in the OP by calling the appropriate command in the `follow` function.

Issue #506, opened 2023-02-21
Open non-text file links in default applications


kflak:
### Does this feature exist in Emacs orgmode core?

Yes

### Orgmode link

https://orgmode.org/manual/Handling-Links.html

### Additional context

Currently, whenever I do `<leader>oo` on a link to a file, it will open in an nvim buffer, regardless of file type. Would be great if this could invoke `xdg-open` or some such to ensure correct handling.

PR #578, opened 2023-06-18
feat: Opening hyperlinks externally

maidl0ver:
This is a stab at https://github.com/nvim-orgmode/orgmode/issues/563 and https://github.com/nvim-orgmode/orgmode/issues/506. `<Leader>oo` now opens linked files with `xdg-open`.
There is also an option to specify `custom:` links, which I would like to read your thoughts on:
```
[[custom:nsxiv:~/some-gif.gif][]] - This link opens a GIF file in nsxiv
[[file:~/some-gif.gif][]] - This link opens a GIF file with the specified default opener
[[file:~/org/notes.org]] - Regular behaviour on open
```
## Issues
This functionality depends on nvim's `vim.filetype.match()` behaviour, which returns `nil` on unrecognized filetypes inferred from the filename. The issue with this is that some plaintext (i.e. `.txt` or files without an extension) would be opened externally. Directories are also treated as externally openable (https://github.com/nvim-orgmode/orgmode/issues/556). If this is not desired behaviour for directories, this could be fixed by checking the type with `vim.loop.fs_stat` and made configurable.

### DIFF
--- a/lua/orgmode/org/mappings.lua
+++ b/lua/orgmode/org/mappings.lua
@@ -754,7 +754,6 @@ function OrgMappings:open_at_point()
   local parts = vim.split(link, '][', true)
   local url = parts[1]
   local link_ctx = { base = url, skip_add_prefix = true }
-  local ext_opener = 'xdg-open'
 
   if url:find('^file:') then
     if url:find(' +', 1, true) then
@@ -770,7 +769,7 @@ function OrgMappings:open_at_point()
       link_ctx.line = url
     elseif vim.filetype.match({ filename = url }) == nil then
       local filepath = Hyperlinks.get_file_real_path(url)
-      vim.fn.jobstart(ext_opener .. ' ' .. vim.fn.shellescape(filepath), { detach = true })
+      vim.fn.jobstart(config.org_external_opener .. ' ' .. vim.fn.shellescape(filepath), { detach = true })
     else
       vim.cmd(string.format('edit %s', Hyperlinks.get_file_real_path(url)))
       vim.cmd([[normal! zv]])
--- a/lua/orgmode/org/mappings.lua
+++ b/lua/orgmode/org/mappings.lua
@@ -789,8 +789,6 @@ function OrgMappings:open_at_point()
 
   if url:find('^shell:') then
     url = string.gsub(url, '^shell:', '')
-    url = vim.fn.shellescape(url)
-    print(url)
     vim.fn.jobstart(url, { detach = true })
     return
   end
--- a/lua/orgmode/org/mappings.lua
+++ b/lua/orgmode/org/mappings.lua
@@ -754,6 +754,8 @@ function OrgMappings:open_at_point()
   local parts = vim.split(link, '][', true)
   local url = parts[1]
   local link_ctx = { base = url, skip_add_prefix = true }
+  local ext_opener = 'xdg-open'
+
   if url:find('^file:') then
     if url:find(' +', 1, true) then
       parts = vim.split(url, ' +', true)
@@ -766,6 +768,9 @@ function OrgMappings:open_at_point()
 
     if url:find('^file:(.-)::') then
       link_ctx.line = url
+    elseif vim.filetype.match({ filename = url }) == nil then
+      local filepath = Hyperlinks.get_file_real_path(url)
+      vim.fn.jobstart(ext_opener .. ' ' .. vim.fn.shellescape(filepath), { detach = true })
     else
       vim.cmd(string.format('edit %s', Hyperlinks.get_file_real_path(url)))
       vim.cmd([[normal! zv]])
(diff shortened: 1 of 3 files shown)


kristijanhusak:
Is `custom:` something that exists in the Emacs Orgmode? 

Also, does Emacs have a way to provide this custom executable for opening things? I didn't find anything in the docs, only some custom functions that allow this https://emacs.stackexchange.com/questions/57506/open-file-in-orgmode-with-external-application-check-if-file-exists-and-create

maidl0ver:
`custom:` is not something implemented in Emacs, I was mostly looking at the types of links listed here: https://orgmode.org/guide/Hyperlinks.html.
And here https://orgmode.org/manual/External-Links.html. 
The `shell:` type is pretty much what `custom:` is, but with an admittedly prettier name (which is easily fixed on our end, of course). Most of the remaining link types are either not relevant for nvim or are already implemented.
There is also this feature: https://orgmode.org/manual/Adding-Hyperlink-Types.html.
Perhaps a more true-to-Emacs way to do this would be to have the link types and their respective openers defined in the config, which I could give a try.

lua/orgmode/org/mappings.lua
```diff
@@ -783,6 +787,13 @@ function OrgMappings:open_at_point()
     return vim.cmd(string.format('edit %s', url))
   end
 
+  if url:find('^custom:') then
```
jgollenz, line 790: If emacs org-mode uses `shell` then we should stick to that. The idea is that a `.org` file that works in emacs should also work in nvim-orgmode.

jgollenz (changes requested):

lua/orgmode/org/mappings.lua
maidl0ver: Changed it to `shell:` as requested.

maidl0ver:
Renamed `custom:` to `shell:`, added a config option (that I couldn't find a _good_ short descriptive name for) to make the external opening behaviour optional. That should make the changes non-breaking.
When it is enabled, however, the issue with some plaintext not recognized as such is still there. Do you know of a better way of inferring filetypes other than with `vim.filetype.match()`?

I could also give implementing user-defined links a shot, if this is desired functionality.

maidl0ver:
Added an option replicating what is described in [Adding Hyperlink Types](https://orgmode.org/manual/Adding-Hyperlink-Types.html), or at least a way of describing how to follow links:
```
org_link_types = {
    -- Opens shell: links
    -- For example [[shell:notify-send Ding!]] sends a notification on open-at-point
    shell = {
      handler = function(url)
        url = string.gsub(url, '^shell:', '')
        vim.fn.jobstart(url, { detach = true })
        return
      end,
    },
    -- Opens print: links
    -- An arbitrary example, just prints the entire link
    print = { 
      handler = function(url)
        print(url)
      end,
    },
  }
  ```
This allows the user to define any prefix in a table and give it any kind of handler one would want.

seflue:
@kristijanhusak, @jgollenz What should we do with this PR? Actually I would like to have the possibility to open file links in orgmode with XDG_OPEN (e.g. PDFs, images), because it doesn't make any sense to open them in Neovim as text. I am used to reference PDFs in my notes, but this is currently useless. While I already thought to implement something myself, I found this open PR, which looks promising.

kristijanhusak:
@seflue we will probably close this.
I plan to refactor hyperlinks to support multiple sources, like it's done for the autocompletion https://github.com/nvim-orgmode/orgmode/tree/master/lua/orgmode/org/autocompletion, and then expose the API to add own sources. I just didn't get to it yet.

jgollenz:
@seflue nothing to add from my part, except a sincere sorry to @maidl0ver for the radio silence on this MR.

seflue:
> @seflue we will probably close this. I plan to refactor hyperlinks to support multiple sources, like it's done for the autocompletion https://github.com/nvim-orgmode/orgmode/tree/master/lua/orgmode/org/autocompletion, and then expose the API to add own sources. I just didn't get to it yet.

I wasn't aware how much you actually refactored autocompletion - this just looks much more structured and extendable then what I remember from when I did my first refactoring on hyperlinks.

joshpetit:
Hey, I'm looking forward to this possibly landing in orgmode. A note I'd love to make is how orgmode has the hyper link types (which you can see [here](https://orgmode.org/manual/Adding-Hyperlink-Types.html)). If the refactor could in some way include adding hyperlink types, that would be awesome.

That would also (I imagine) give direction towards custom link transformation when exporting a pdf (which you can see [here](https://orgmode.org/manual/Adding-Hyperlink-Types.html#:~:text=defun%20org-man-export)). These are some powerful org-mode utils.

kristijanhusak:
This can now be done with introduction of custom hyperlink type. See https://nvim-orgmode.github.io/configuration#custom-hyperlink-types for more info.

PR #582, opened 2023-06-26
docs: Add reference how to open hyperlinks (#580)

seflue:
Provide the reference mentioned in the discussion.

### DIFF
--- a/lua/orgmode/org/hyperlinks.lua
+++ b/lua/orgmode/org/hyperlinks.lua
@@ -3,7 +3,16 @@ local utils = require('orgmode.utils')
 local Hyperlinks = {}
 
 local function get_file_from_context(ctx)
-  return (ctx.hyperlinks and ctx.hyperlinks.filepath and Files.get(ctx.hyperlinks.filepath) or Files.get_current_file())
+  local filepath = (ctx.hyperlinks and ctx.hyperlinks.filepath)
+  if not filepath then
+    return Files.get_current_file()
+  end
+  local canonical = vim.loop.fs_realpath(filepath)
+  if not canonical then
+    return Files.get_current_file()
+  end
+
+  return Files.get(canonical)
 end
 
 local function update_hyperlink_ctx(ctx)
@@ -21,6 +30,10 @@ local function update_hyperlink_ctx(ctx)
   local file_match = ctx.line:match('file:(.-)::')
   if file_match then
     file_match = Hyperlinks.get_file_real_path(file_match)
+  end
+
+  if file_match then
+    file_match = vim.loop.fs_realpath(file_match)
   end
 
   if file_match and Files.get(file_match) then
--- a/tests/plenary/org/autocompletion_spec.lua
+++ b/tests/plenary/org/autocompletion_spec.lua
@@ -218,11 +218,15 @@ describe('Autocompletion', function()
     })
 
     mock_line(api, string.format('  [[file:%s::*', filename))
+    local vim_loop = mock(vim.loop, true)
+    vim_loop.fs_realpath.returns(filename)
     result = OrgmodeOmniCompletion(0, '*')
     assert.are.same({
       { menu = '[Org]', word = '*' .. headlines[1].title },
       { menu = '[Org]', word = '*' .. headlines[2].title },
     }, result)
+
+    mock.revert(vim_loop)
 
     mock.revert(MockFiles)
 


doc/orgmode.txt
```diff
@@ -1436,6 +1436,8 @@ using the '+' character (Example: `file:/home/user/.config/nvim/init.lua +10`) o
 * Headline with `CUSTOM_ID` property within the same file (starts with `#`)
 * Fallback: If file path, opens the file, otherwise, tries to find the headline title.
 
+To open a hyperlink, see |orgmode-org_open_at_point|.
```
jgollenz, line 1439: @seflue thanks for the PR :) the files in `doc/` are auto-generated whenever `DOCS.md` is changed. Sorry, I should have told you that (or even better, it should be more explicit in our `CONTRIBUTING.md`). Please move your change to `DOCS.md` and remove it from `doc/orgmode.txt` :+1:

jgollenz (changes requested):

doc/orgmode.txt
seflue: I have two questions here. First, the `doc/orgmode.txt` is under version control. How is the file generated and checked in? When I downloaded md2vim and ran it via make docs I get a lot of error messages `StrikeThrough is unimplemented`.
Second, I don't know, if I just hove the change to DOCS.md, that this will preserve the tag reference in vim helpfile style I am aiming to create here. md2vim documentation has no hint about how links are handled and a brief look into the code doesn't reveal anything enlightening.
The tag references are a core feature of vim helpfiles, it would be a shame, if they are only auto-generated in the table of contents, but cannot be used to references between help sections.

seflue: Seems, that md2vim is somehow [half-baked](https://github.com/FooSoft/md2vim/issues/1). I don't know, if the maintainers of nvim-orgmode classify their plugin as "half-representable". But as feature-rich as it is, it deserves to have well written (and maintainable) vim helpfiles. I would prefer this actually over a extensively written markdown file. I use the helpfiles heavily, because within vim I can fuzzy-search for help tags. This is much more important for a vim poweruser than having exactly the same documentation in markdown. Why not just reference the vim documentation from the readme, if you guys don't want to maintain both files?

Btw., md2vim is archived, doesn't seem to be actively developed anymore.

jgollenz: I'm not a heavy user of helpfiles, but I see your points. Since it was @kristijanhusak who decided on the current setup, he should also decide whether we change it.

As for md2vim not being maintained and tag references missing, I could look into that. Using `|some reference|` in .md might be possible, according to the thread you shared.

DOCS.md
```diff
@@ -1054,6 +1054,8 @@ using the '+' character (Example: `file:/home/user/.config/nvim/init.lua +10`) o
 * Headline with `CUSTOM_ID` property within the same file (starts with `#`)
 * Fallback: If file path, opens the file, otherwise, tries to find the headline title.
 
+To open a hyperlink, see |orgmode-org_open_at_point|.
```
kristijanhusak, line 1057: To link to a specific point in the docs use the markdown link format, like for example here https://github.com/nvim-orgmode/orgmode/blob/7834aea491c02c0eddd2fbc94714535bf06776b3/DOCS.md?plain=1#L814

kristijanhusak (changes requested):

doc/orgmode.txt
kristijanhusak: > Seems, that md2vim is somehow [half-baked](https://github.com/FooSoft/md2vim/issues/1). I don't know, if the maintainers of nvim-orgmode classify their plugin as "half-representable". But as feature-rich as it is, it deserves to have well written (and maintainable) vim helpfiles. I would prefer this actually over a extensively written markdown file. I use the helpfiles heavily, because within vim I can fuzzy-search for help tags. This is much more important for a vim poweruser than having exactly the same documentation in markdown. Why not just reference the vim documentation from the readme, if you guys don't want to maintain both files?
> 
> Btw., md2vim is archived, doesn't seem to be actively developed anymore.

One of the biggest pain points of most software, especially open source, is documentation, and orgmode is not an exception to that. At the time when this was built, md2vim was the most decent solution that I could find. If you know any plugin that is maintained and can do everything that md2vim does, please let us know. 

Having specifically written vim help files would be helpful, but it's very time consuming, and barely anyone will take responsibility for maintaining it separately while keeping the markdown docs in sync. If you know a tool that can actually create a markdown documentation from the vimfiles documentation, that can also be a way to solve this.

Saying that maintainers classify the plugin as "half-representable" is not of any help to us or anyone using the plugin. Also, if you believe that it deserves a well written and maintainable vim helpfiles, feel free to contribute at any time. We welcome any type of contributions.

seflue: The "half-representable" was a [cite](https://github.com/FooSoft/md2vim/issues/1#issuecomment-623157378) from the md2vim-thread - that's how FooSoft characterized their targeted user base. I explicetly not mean, that nvim-orgmode is half-representable, actually I meant the opposite. Why I wanted to raise the question, if the project should rely on a half-baked tool (which md2vim is).

I decided to go all in on nvim-orgmode and I'm currently heavily working on a refactoring of the hyperlink code of nvim-orgmode, because this feature is very important for me and the feature still quite buggy. So I am willing to invest some time into the project to improve it.
Fixing external dependencies like md2vim is currently not my priority, although good documentation is very important for me.

Issue #466, opened 2022-12-12
Tracking TODO state changes


PhosCity:
### Does this feature exist in Emacs orgmode core?

Yes

### Orgmode link

https://orgmode.org/manual/Tracking-TODO-state-changes.html

chuck-flowers:
I'm taking a crack at this. The development branch for it is [here](https://github.com/chuck-flowers/orgmode/tree/logging-config)

Not sure when I'll finish it since I'll just be working on it as I have time.

chuck-flowers:
Just wanted to update what state the branch is in. I had planned to work on the feature over my Christmas vacation but unfortunately I fell ill and didn't make much progress on it. I've started working on the feature again in my off hours and I'm currently dogfooding my changes.

In its current state, a `:LOGGING:` value of `nil` will successfully turn the logging off. I've written the logic for parsing the syntax of the `:LOGGING:` property, but I still need to validate it, and use the config for affecting the TODO state change behavior. Hopefully in another few days I'll have that done though.

@PhosCity are you familiar with how `:LOGGING:` affects the state change behavior in emacs? The only feature I really needed out of it currently is being able to disable logging for some of my repeating tasks. When I'm done will you be able to test it to confirm it behaves in the same way that emacs does?

PhosCity:
Hi, thank you for working in this feature. Unfortunately I won't be of much help to you in this regard. I don't use emacs myself so I can only determine the behavior of emacs by reading docs which I'm sure you've done already. So I can't add anything further.

I hope someone else who uses emacs can help us here.

cands:
Hi, thanks a lot @chuck-flowers for taking a crack at this, it's an important feature in my opinion. I may be able to help with testing to confirm the behavior is the same as in emacs, I use the nvim-orgmode plugin most of the time but go to emacs for stuff not fully supported yet (like habits).

Besides being able to disable logging it would be great if we can optionally save state changes into the LOGBOOK drawer (or ideally user-selected drawer), which is especially important for habits that can have a lot of state changes that may clutter the view. Furthermore, it would be awesome if became possible to add configuration on a per-keyword basis as in emacs, please see the Org Manual for a description: [5.3.2 Tracking TODO state changes](https://orgmode.org/manual/Tracking-TODO-state-changes.html)

chuck-flowers:
[...]

PR #584, opened 2023-06-26
Adds support for :LOGGING: property

chuck-flowers:
Addresses issue #466 by adding support for the `:LOGGING:` property to the note/timestamp insertion logic when updating the TODO state of tasks.

I factored out the parsing framework I created for adding date property support to the agenda search. I used this framework to parse the syntax that orgmode describes in its manual for the `:LOGGING:` property. I then, determined whether to insert a timestamp or note by first checking for the `:LOGGING:` property on the affected section and then using the property defined in the user's orgmode config.


chuck-flowers:
I just realized that master has moved forward since I opened the PR. I can rebase this onto master unless that will cause a problem

chuck-flowers:
I've rebased on the latest master since git was reporting that there was a merge conflict

lua/orgmode/utils/parsing.lua
```diff
@@ -0,0 +1,69 @@
+local M = {}
```
kristijanhusak, line 1: Let's move this file to `lua/orgmode/parsers/utils.lua`.

lua/orgmode/org/mappings.lua
```diff
@@ -394,9 +397,53 @@ function OrgMappings:_todo_change_state(direction)
...
+
+  -- Determine which configuration to use
+  local global_config = config.org_log_done
+
+  --- @type nil | false | 'time' | 'note'
+  local section_config
+  local logging_prop = item.properties.items.logging
+  if logging_prop == 'nil' then
+    section_config = false
+  elseif logging_prop ~= nil then
```
kristijanhusak, line 410: Having headline or file scoped config for todo keywords requires many more changes across the codebase to be properly considered. It will work on state change, but agenda, highlights, etc. will not work as expected.

Can we have this PR only handle `@` and `!` in the todo keywords and leave the file/headline level todo keywords for a different PR?

kristijanhusak (changes requested):
Thanks for the PR! And sorry for the late review.
The code looks generally ok, but as I mentioned in one of the comments, I think we cannot do file/headline level todo keywords at this moment. It needs many more changes in order to work properly. 
Adding support for `@` and `!` in the todo keywords is already a great addition, so I would like to scope this PR to that.

lua/orgmode/utils/parsing.lua
chuck-flowers: I've renamed the file as you've requested

lua/orgmode/org/mappings.lua
chuck-flowers: I reverted the 2 commits which targeted the mappings file. If I remember right, this shouldn't affected parsing the "global" config value, but I will reconfirm this with my personal usage tomorrow.

chuck-flowers:
@kristijanhusak I think I have addressed the requested feedback. Please let me know if I have missed something

kristijanhusak (changes requested):
@chuck-flowers This PR now just introduces a `TodoConfig` parser without actually using it? Is that intended?
I thought we will incorporate it to work with the todo keywords defined in the global configuration. 
I'm ok doing that separately, and have this only as a refactor + preparation for the next PR. Just want to confirm that is the intention.

Could we have a unit test that tests `todo-config` file to ensure it parses everything properly?

PR #588, opened 2023-07-02
Refactoring/hyperlinks

seflue:
As I already mentioned, I am working on a refactoring to make the hyperlinks code more robust and easier to extend to the many use cases specified in the org manual. As a starting point I introduce Link and Url as separate objects, which can be tested independently. I added some initial test cases, which also should be easy to extend.


seflue:
@kristijanhusak, @jgollenz What do you think about this approach?

seflue:
I also split up some tests to make it easier spotting bugs and adding new tests.

lua/orgmode/utils/fs.lua
seflue, line 1: I introduced this little helper module to encapsulate all interactions with file-system related apis. This makes it easier to mock them in tests.

lua/orgmode/org/hyperlinks.lua
```diff
@@ -77,98 +43,127 @@ function Hyperlinks.find_by_filepath(ctx)
...
+  elseif url:is_dedicated_anchor_or_internal_title() then
+    result = utils.concat(
+      Hyperlinks.find_by_dedicated_target(url),
+      -- TODO replace with real fuzzy title search
+      Hyperlinks.find_by_title(url)
+    )
+    mapper = Hyperlinks.as_dedicated_anchors_or_internal_titles(url)
+  end
+
+  return result, mapper
```
seflue, line 166: I wanted to get rid of this ugly `skip_add_prefix` flag. By returning the headlines together with the mapper function, it's up to the caller to do the appropiate mapping.

lua/orgmode/org/hyperlinks.lua
```diff
@@ -77,98 +43,127 @@ function Hyperlinks.find_by_filepath(ctx)
...
+  elseif url:is_custom_id() then
+    result = Hyperlinks.find_by_custom_id_property(url)
+    mapper = Hyperlinks.as_custom_id_anchors
+  elseif url:is_headline() then
+    result = Hyperlinks.find_by_title(url)
+    mapper = Hyperlinks.as_headline_anchors
+  elseif url:is_dedicated_anchor_or_internal_title() then
+    result = utils.concat(
+      Hyperlinks.find_by_dedicated_target(url),
+      -- TODO replace with real fuzzy title search
```
seflue, line 160: I'm not sure, if this would be a use case for others, but as a last resort it could be convenient to refer between headlines within a document. Currently it has to match with the beginning of a headline.

lua/orgmode/org/hyperlinks.lua
```diff
@@ -1,61 +1,27 @@
...
+---@param url Url
+---@return string[]
+function Hyperlinks.find_by_filepath(url)
   local filenames = Files.filenames()
-  local file_base = ctx.base:gsub('^file:', '')
+  local file_base = url:get_filepath()
+  if not file_base then
+    return {}
+  end
+  --TODO integrate with orgmode.utils.fs or orgmode.objects.url
```
seflue, line 21: I'm not sure, if we need this function anymore. I have to admit, that I didn't completely understand, which use-cases it actually tries to cover. It seems to only find org-files, which are already parsed. My gut-feeling tells me, that we could simplify it like in `find_by_custom_id` and `find_by_title`, but I wasn't sure, if I would remove something. @kristijanhusak can you enlighten me?

lua/orgmode/org/autocompletion/omni.lua
```diff
@@ -33,7 +35,10 @@ local properties = {
 local links = {
   line_rgx = vim.regex([[\(\(^\|\s\+\)\[\[\)\@<=\(\*\|\#\|file:\)\?\(\(\w\|\/\|\.\|\\\|-\|_\|\d\)\+\)\?]]),
   rgx = vim.regex([[\(\*\|\#\|file:\)\?\(\(\w\|\/\|\.\|\\\|-\|_\|\d\)\+\)\?$]]),
-  fetcher = Hyperlinks.find_matching_links,
+  fetcher = function(url)
+    local hyperlinks, mapper = Hyperlinks.find_matching_links(url)
+    return mapper(hyperlinks)
```
seflue, line 40: Making the mapping explicit in the omnicompletion helped me a lot to wrap my head around what actually happens here.

lua/orgmode/org/autocompletion/omni.lua
```diff
@@ -101,7 +106,8 @@ local function omni(findstart, base)
     return -1
   end
 
-  local fetcher_ctx = { base = base, line = line }
+  local url_str = line:match('%[%[(.-)$')
+  local url = Url.new(url_str)
```
seflue, line 110: The `Url` is all what we need inside the `fetcher` callback.

lua/orgmode/org/autocompletion/omni.lua
```diff
@@ -112,7 +118,7 @@ local function omni(findstart, base)
     then
       local items = {}
       if context.fetcher then
-        items = context.fetcher(fetcher_ctx)
+        items = context.fetcher(url)
       else
         items = { unpack(context.list) }
       end
```
seflue, line 124: The following lines expose a bug which I currently don't understand completely:
When I use file urls, which don't start with the `file:` prefix, omnicompletion only finds headlines when typing `::*` after the file. If I type `::#` I would expect omnicompletion to find custom_ids. They get removed here, because `base` is `:#` while it is just `*` when I aim for headlines (which matches the pattern in the filter).

I haven't understand, how the to cases differ regarding the calculation of `base`. @kristijanhusak Any ideas?

seflue: Both cases work perfectly fine, if I have no file (the link starts with `#` or `*` respectively) or if the `file:` prefix is used.

tests/plenary/ui/mappings/headline_spec.lua
```diff
@@ -173,7 +173,7 @@ describe('Heading mappings', function()
       '       Some content for level 2',
       '******* NEXT [#1] Level 3',
       '        Content Level 3',
-    }, vim.api.nvim_buf_get_lines(0, 0, 6, false))
```
seflue: This seemed to be a copy/paste mistake I removed along the way.

seflue:
@kristijanhusak, @jgollenz Is there any chance to get this merged?

kristijanhusak:
Hey @seflue , sorry for delaying this. It's a big PR and I can't find enough time to review it properly. I'll try to look into it next week.

seflue:
@kristijanhusak Thanks for the reply. Take your time. I tried to make it a bit easier to review by adding some comments.

lua/orgmode/org/hyperlinks.lua
kristijanhusak: The idea is to show all available org files in autocomplete popup when you start typing. Basically, typing this:
```
[[file:./
```
Should give you list of all org files that are part of `org_agenda_files` in your setup. We need to leave it, since  other functions will not return that.

lua/orgmode/org/autocompletion/omni.lua
kristijanhusak: This change breaks the manual omni autocompletion (`<C-x><C-o>`). Doing this:
```lua
  local url = Url.new((base or '') .. url_str)
```
Solved part of the issue by properly autocompleting files, but autocompleting headlines and custom ids after `::` is not working.

I tested this by disabling the cmp in my config and just testing with `<C-x><C-o>`.

kristijanhusak (changes requested):
Thanks for the PR once more and sorry for delaying it so much.

This is a great refactoring and reorganization. It seems that everything works fine with cmp but when I tried manual omni autocompletion it seems it stopped working. I'm not 100% sure what's the issue but I suspect it's in the `omni` function where you create an url.

Lets make sure it works with manual omni completion and cmp. Other completion plugins are no longer maintained so we don't need to worry about them.

seflue:
> Lets make sure it works with manual omni completion and cmp. Other completion plugins are no longer maintained so we don't need to worry about them.

Sure, let's try to fix this. Actually I wasn't aware of the integration with manual omni completion and didn't know the expectation, that  it should also filter on the agenda files. I must have overlooked it in the original code.

Actually my manuall omni-completion seems to be broken in my current nvim-configuration, so first I need to figure out, what is broken in my setup. When I type <C-x><C-o> nothing happens. I set a breakpoint within `Omni.get_completions` but the function isn't called. Could you please give me some support how to setup manual omni completion properly with nvim-orgmode?

kristijanhusak:
> Could you please give me some support how to set-up manual omni completion properly with nvim-orgmode?

Just disabling the cmp should make it work. In my case, I just add `enabled = false` to my completion setup https://github.com/kristijanhusak/neovim-config/blob/afd378472390d5928d901ac4e23b51846552a842/nvim/lua/partials/plugins/completion.lua#L1 and that completely disables autocompletion. Commenting out cmp setup call should also do the trick.
Once you open up an org file omnifunc should be automatically working. You can check if `:echo &omnifunc` returns `OrgmodeOmni`. That is set here https://github.com/nvim-orgmode/orgmode/blob/d26d52bb47e8978adf4ec78e20930b8f312fa473/ftplugin/org.vim#L28 and calls underlying lua function here https://github.com/nvim-orgmode/orgmode/blob/d26d52bb47e8978adf4ec78e20930b8f312fa473/lua/orgmode/org/autocompletion/omni.lua#L90

>  That it is only the first entry seems to be a bug. But I need to understand, how this feature actually works to properly support your requirement. Can you give me some hints, which parts of the code are relevant to this?

I'm not sure what you mean by "only first entry seems to be a bug". Can you clarify?
Requirement here is to first autocomplete all files that are part of your `org_agenda_files`, no matter if you start with absolute or relative path. For example, If I type `[[file:/` and trigger autocompletion, I get something like this:

```
file:/home/kristijan/orgfiles/todos.org
file:/home/kristijan/orgfiles/work.org
file:/home/kristijan/orgfiles/notes.org
etc.
```

If I try relative with `[[file:./` I get this:
```
file:./todos.org
file:./work.org
file:./notes.org
etc.
```

> A short resarch in the code brought me no connection to the agena files, only to the tags (and for this the completion works in my opinion).

On current master, these completion entries are fetched from here https://github.com/nvim-orgmode/orgmode/blob/d26d52bb47e8978adf4ec78e20930b8f312fa473/lua/orgmode/org/hyperlinks.lua#L54, and in your PR, it's from here https://github.com/nvim-orgmode/orgmode/pull/588/files#diff-818184336fd62940c83c000ff499ec9b5a409ff654e509846b4cbbb9592ffb86R16

> Could you please try out, if the feature works as intended on the current master? If not, it would be more convenient for me to first merge this PR and then create a new one.

The current master works as intended, that's why I mentioned it. Let's try to first make sure this PR doesn't break any of the existing functionality before proceeding with other stuff. Later you should be able to rebase.

lua/orgmode/org/hyperlinks.lua
seflue: Thanks for the clarification.

seflue:
> For example, If I type `[[file:/` and trigger autocompletion, I get something like this:

This actually still works with cmp.

> I'm not sure what you mean by "only first entry seems to be a bug". Can you clarify?

I can specify multiple entries for org files in the config. Each entry accepts globbing, so it can represent multiple agenda files. In my case, only the files, which are expanded from the globbing of the first entry are visible, when I type `[[file:/`, the files matched by the other entries are not visible.

> Just disabling the cmp should make it work. In my case, I just add `enabled = false` to my completion setup https://github.com/kristijanhusak/neovim-config/blob/afd378472390d5928d901ac4e23b51846552a842/nvim/lua/partials/plugins/completion.lua#L1 and that completely disables autocompletion. Commenting out cmp setup call should also do the trick. Once you open up an org file omnifunc should be automatically working. You can check if `:echo &omnifunc` returns `OrgmodeOmni`. That is set here
> 
> https://github.com/nvim-orgmode/orgmode/blob/d26d52bb47e8978adf4ec78e20930b8f312fa473/ftplugin/org.vim#L28
> and calls underlying lua function here
> 
> https://github.com/nvim-orgmode/orgmode/blob/d26d52bb47e8978adf4ec78e20930b8f312fa473/lua/orgmode/org/autocompletion/omni.lua#L90

Thanks, this actually helps. I need to fix my setup and then test it by myself. I will get back to you when I have a fix or more questions.

kristijanhusak:
> This actually still works with cmp.

Yes, cmp works, but omni doesn't. This is because the line is parsed differently between the two.

> I can specify multiple entries for org files in the config. Each entry accepts globbing, so it can represent multiple agenda files. In my case, only the files, which are expanded from the globbing of the first entry are visible, when I type [[file:/, the files matched by the other entries are not visible.

I have the same issue when I use relative path (`[[file:./`), but I don't have the issue with absolute paths (`[[file:/`). Does the same happen to you?

seflue:
Now that I was able to configure my Vim in a way, that triggering omnicompletion actually hits a breakpoint, I did some debugging. I was suprised, that my assumption, what is found in `line` was just false for OmniCompletion.

The reason is probably, that `omni` are actually two functions, crammed into one. And which code is actually called is decided by this `findstart` flag.
I already have a [refactoring for that](https://github.com/seflue/nvim-orgmode/commit/1d27a2836c26871f8c9a816819a6eeed56da1fd3), but when I implemented this, I missed the entry point in `ftplugin/org.vim`.

I will implement a fix for this branch and also a correction for my OmniCompletion refactoring. For the latter I will create a PR after we merged this branch.

lua/orgmode/org/hyperlinks.lua
```diff
@@ -1,61 +1,27 @@
...
   local file_base_no_start_path = file_base:gsub('^%./', '') .. ''
   local is_relative_path = file_base:match('^%./')
-  local current_file_directory = vim.fn.fnamemodify(utils.current_file_path(), ':p:h')
+  local current_file_directory = fs.get_current_file_dir()
   local valid_filenames = {}
   for _, f in ipairs(filenames) do
     if is_relative_path then
```
seflue, line 27: @kristijanhusak in the next line the behavior for relative paths is specified in case of agenda files.
It is very simple: When you are in a directory with agenda files directly in it, you will see them in the completion. Otherwise they are explicitly ignored.
A simple fix would be to replace the `[^/]*` with a `.*` match org files, which are deeper nested from the current path.
To find the relative path to all agenda files, no matter where we are relative to them (e.g. a sibling file tree) would require to implement more complex logic.
So what do you think, which behavior should it be for the merge?

lua/orgmode/org/autocompletion/omni.lua
```diff
@@ -101,7 +110,26 @@ local function omni(findstart, base)
     return -1
   end
 
-  local fetcher_ctx = { base = base, line = line }
+  -- The function omni gets called twice when used by omnicompletion but when
```
kristijanhusak, line 113: I don't think we should have any completion specific logic here. I would rather introduce something like `Url.from_omni(line, base)` and figure out the URL from there.

Also, you can just do a `gsub` without checking, it will just return the string if there is no match. This will avoid one regex call:

```lua
local url_str = base:gsub('^file:', '')
```

kristijanhusak (changes requested):
I reviewed the code but didn't give it a full test. Please address this one comment and then I'll check. Thanks!

seflue:
@kristijanhusak I don't want to annoy you, but this PR is now open for quite some time and I didn't hear back from you since I did the requested changes. Does this PR still has a chance to get merged?

kristijanhusak (approved):
Everything is working great now!

Thank you very much for the PR and the hard work!

And sorry for the delay. It's a bit bigger PR so I need to devote more time to reviewing it.

PR #591, opened 2023-07-15
Allow dynamic target in capture templates

sotte:
Internally this compiles the target of the capture template the same way the template itself is compiled.

Also update the docs accordingly.

This is inspired by: https://github.com/nvim-orgmode/orgmode/discussions/540

### DIFF
--- a/DOCS.md
+++ b/DOCS.md
@@ -421,6 +421,15 @@ Journal example:<br />
   } }
   ```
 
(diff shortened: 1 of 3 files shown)


sotte:
Can I do anything to help with the review of this MR?

kristijanhusak:
I'm a bit sceptic around this since it's not something that is supported by the Emacs orgmode, but I'll merge it in and see if it will cause issue for anyone. Thanks.

PR #594, opened 2023-07-23
`vim.v.count1` is needed on fallback mapping so `<count><C-A>` mappings work outside dates

oncomouse:
Currently pressing `6<C-A>` outside of a date just adds 1 instead of 6 to a number that isn't a date. This is because the fallback mapping does not include `vim.v.count1`.


kristijanhusak (approved):

kristijanhusak:
Thanks!

PR #595, opened 2023-07-24
Fix tag parsing for search patterns.

troiganto:
Search patterns only accepted tags that match the [Lua pattern][1] `%w+`. This only includes alpha-numeric characters. However, according to [Org syntax][2], tags may also contain underscores, at signs, hash signs and percent signs.

This commit fixes the parsing pattern and adds a test case that includes each of these punctuation characters.

As always, thanks for this wonderful plugin!

[1]: https://www.lua.org/manual/5.1/manual.html#5.4.1
[2]: https://orgmode.org/worg/org-syntax.html#Headings


kristijanhusak:
Thanks!

Issue #473, opened 2022-12-17
`org_indent_mode = "noindent"` is buggy with lists


jgollenz:
### Describe the bug

~~When set to "noindent", adding a newline will still respect the indentation of the previous line~~ I had a typo in the config, but it looks like I found a different bug. See comment below

Also, the default value for emacs seems to be "noindent", not "indent"

### Steps to reproduce

1. Run `nvim -u minimal_init.lua foo.org`
2. Add the following line
```
* foo|
```
3. Press `<Enter>`

### Expected behavior

The cursor should be at the beginning of the line:

```
* foo
|
```

### Emacs functionality

Emacs in general does not add any actual indenting after a headline. When `org-indent-mode` is activated, it only adds the respective virtual indentation:

![org-indent-mode](https://user-images.githubusercontent.com/12050098/208251149-5a111085-7e9b-491b-98d8-ac2143b936ba.gif)

This is not true for lists, however, no matter whether `org-indent-mode` is active or not. Our current description of `org_indent_mode` in the docs, does not reflect what the option does in emacs.

### Minimal init.lua

```lua
vim.cmd([[set runtimepath=$VIMRUNTIME]])
vim.cmd([[set packpath=/tmp/nvim/site]])

local package_root = '/tmp/nvim/site/pack'
local install_path = package_root .. '/packer/start/packer.nvim'

local function load_plugins()
  require('packer').startup({
    {
      'wbthomason/packer.nvim',
      { 'nvim-treesitter/nvim-treesitter' },
      { 'kristijanhusak/orgmode.nvim', branch = 'master' },
    },
    config = {
      package_root = package_root,
      compile_path = install_path .. '/plugin/packer_compiled.lua',
    },
  })
end

_G.load_config = function()
  require('orgmode').setup_ts_grammar()
  require('nvim-treesitter.configs').setup({
    ensure_installed = { 'latex' },
    highlight = {
      enable = true,
      additional_vim_regex_highlighting = { 'org' },
    },
    org_indent_mode = 'noindent',
  })

  vim.cmd([[packadd nvim-treesitter]])
  vim.cmd([[runtime plugin/nvim-treesitter.lua]])
  vim.cmd([[TSUpdateSync org]])

  -- Close packer after install
  if vim.bo.filetype == 'packer' then
    vim.api.nvim_win_close(0, true)
  end

  require('orgmode').setup({
    org_highlight_latex_and_related = 'native',
  })

[...]

Issue #472, opened 2022-12-17
Indentation increases with each item


pianocomposer321:
### Describe the bug

I'm having a hard time determining if this is intended or not...but it's really annoying and I'm not sure how to disable it, or if I'm just doing something wrong. The problem is that when I'm writing a list, instead of being on the same indentation level, it adds one on the new line. This is super confusing, and it also happens when `org_indent_mode` is set to `noindent`.

### Steps to reproduce

When I start a list like so (| = cursor):

```org
- List Item|
```

Then I press enter, I would expect this:

```org
- List Item
|
```

But instead I get this:

```org
 - List Item
   |
```

### Expected behavior

Explained above.

### Emacs functionality

Not really sure...I'm not an emacs user, and I'm not in a position to install it and check at the moment, but what demo videos I can find online do not seem to function the way this plugin does...though I think they're doing something more clever that I'm not sure is possible with neovim at all at the moment (with virtual indents and whatnot).

### Minimal init.lua

The template works: https://github.com/nvim-orgmode/orgmode/blob/master/scripts/minimal_init.lua

```lua
vim.cmd([[set runtimepath=$VIMRUNTIME]])
vim.cmd([[set packpath=/tmp/nvim/site]])

local package_root = '/tmp/nvim/site/pack'
local install_path = package_root .. '/packer/start/packer.nvim'

local function load_plugins()
  require('packer').startup({
    {
      'wbthomason/packer.nvim',
      { 'nvim-treesitter/nvim-treesitter' },
      { 'kristijanhusak/orgmode.nvim', branch = 'master' },
    },
    config = {
      package_root = package_root,
      compile_path = install_path .. '/plugin/packer_compiled.lua',
    },
  })
end

_G.load_config = function()
  require('orgmode').setup_ts_grammar()
  require('nvim-treesitter.configs').setup({
    highlight = {
      enable = true,
      additional_vim_regex_highlighting = { 'org' },
    },
  })

  vim.cmd([[packadd nvim-treesitter]])
  vim.cmd([[runtime plugin/nvim-treesitter.lua]])
  vim.cmd([[TSUpdateSync org]])

  -- Close packer after install
  if vim.bo.filetype == 'packer' then
    vim.api.nvim_win_close(0, true)
  end

  require('orgmode').setup()

  -- Reload current file if it's org file to reload tree-sitter
  if vim.bo.filetype == 'org' then
    vim.cmd([[edit!]])
  end
end

[...]

PR #597, opened 2023-08-13
Refactor `indentexpr()` to fix `noindent` indentation for lists. 

troiganto:
Closes https://github.com/nvim-orgmode/orgmode/issues/473.

I tried doing small fixes to this code, but kept running into edge cases. Hence, this complete rewrite. :) The important points:

- The queries in `indent.scm` no longer match on top-level (i.e un-nested) lists, but instead on list items of all levels.
- List item indentation no longer relies on the previous non-empty line. Each list item stores whether it's in a top-level or a nested list and calculates its indent based on that.
- The check whether we are in bulleted line or not no longer uses `str.match()`, since its pattern was buggy and forgot a few kinds of bullets. (namely, indented `*` bullets and `a.` ordered bullets) Instead, we compare the current line number to `match.line_nr`. We can do that because we query list items instead of lists now.

There is an edge case when the user is appending to a list. We want that next line to be indented (see https://github.com/nvim-orgmode/orgmode/issues/472), but it's technically outside of the list. At the same time, if an unindented line follows a list, it should not become part of the list.

The best solution I found for this was to make the behavior of `indentexpr()` depend on whether we are in insert mode. If yes, the line after a list is part of the list. If not, it isn't.

The new code also correctly takes into account that two consecutive empty lines always end a preceding list.

-----

Let me know if there are any changes that should be made. It's kind of a big PR, so I assume it'll take a while to review. Thanks as always for maintaining this plugin! :+1:


troiganto:
As far as I can tell, the latest test failure is a fluke. I can't seem to rerun the test though.
```diff
Fail	||	Todo mappings should add last repeat property and state change to drawer (org_log_into_drawer)	
            .../orgmode/orgmode/tests/plenary/ui/mappings/todo_spec.lua:114: Expected objects to be the same.
-             *[4] = '  :LAST_REPEAT: [2023-10-01 Sun 23:00]'
+             *[4] = '  :LAST_REPEAT: [2023-10-01 Sun 23:01]'
```

kristijanhusak (approved):
This looks good to me, thanks! @jgollenz can you take a look at this and confirm it fixes the issue you reported?

kristijanhusak:
Thanks for the PR @troiganto!

@jgollenz if you find some issue regarding this PR please open up a separate issue, thanks!

PR #598, opened 2023-08-13
wip(colors): expose highlight groups for user-defined priorities

levouh:
Generally speaking, this PR serves to allow the user to do something like:

`hi link OrgPriorityC @text.note`

to get something like:

image

[Officially](https://orgmode.org/manual/Priorities.html) it does not seem there is support for highlighting priorities, so I'm not sure if this should play the same as something like `:h orgmode-org_todo_keyword_faces` when defining the highlights.

---

Acknowledging that there still need to be:
- [ ] Tests
- [ ] Documentation updates

but wanted to throw out a draft to see the approach that should be taken for this.

Wanted to get some feedback on:

- How the highlights should be exposed:
    1. Just add documentation for these highlights and allow the user to define them the "Neovim way".
    2. Add a new `org_priority_keyword_faces` option to allow the user to define them the "Org way". This would also be inline with the way that highlights work for TODO keywords, regardless of how ergonomic/normal it is for the Neovim user.
- Whether pulling the "highlights that come from user configuration" into a separate module in this manner makes sense.
- Other things as noted through self-comments on the PR.


lua/orgmode/colors/from_config/init.lua
```diff
@@ -0,0 +1,58 @@
+local ts = require('orgmode.treesitter.compat')
+local tree_utils = require('orgmode.utils.treesitter')
+local utils = require('orgmode.utils')
+
+local M = {}
+
+---@class OrgConfigHighlighter
+---@field get_queries function()
```
levouh, line 8: _todo_: Need to figure out how to define the right return type for the `function` field here.

lua/orgmode/colors/from_config/priority.lua
```diff
@@ -0,0 +1,12 @@
+local config = require('orgmode.config')
+
+local M = {}
+
+function M.get_queries()
+  return vim.tbl_map(function(priority)
+    return string.format('((expr) @OrgPriority%s (#eq? @OrgPriority%s "[#%s]"))', priority, priority, priority)
```
levouh, line 7: _todo_: Not sure if there is a better way to format in the same value multiple times w/ Lua's `string.format`.

lua/orgmode/config/init.lua
```diff
@@ -259,6 +261,35 @@ function Config:get_todo_keywords()
   return types
 end
 
+--- Get a list of priority values based on user configuration of the low/high values
+---
+---@return table<string> priority values in ascending order
+function Config:get_priority_values()
```
levouh, line 267: _todo_: This is a minor duplicate of the code for the priority state stuff, so not sure where the right line is between de-duplicating code versus favoring "simplicity".

lua/orgmode/colors/from_config/priority.lua
levouh, line 7: _todo_: The [official](https://orgmode.org/manual/Priorities.html) docs denote that the priority always comes second after the "TODO state", so where we could get away with something like `(item . (expr) . (expr) ...)` to extract "the second expression" rather than comparing against _all_ expressions?

PR #604, opened 2023-08-28
A new logo

rwmpelstilzchen:
Hi,

I began using `nvim-orgmode` recently and it made a huge change in the way I handle tasks and events 🙂

As a thank-you, I designed and implemented a new minimalist logo, which blends the Neovim (shape and gradients) and the Org-mode (colours and composition of colours) logos. Building on characterising features of both, it is instantly recognisable by anyone who is familiar with the two logos yet has an identity of its own that goes beyond the current logo, which overlays the Org-mode logo over the Neovim one.


rwmpelstilzchen:
> Nice! Looks great! @jgollenz let me know what you think!

Thank! I’m glad to hear you like it 🙂

> Could you add a png version that is in the same path as the old one (`assets/orgmode_nvim.png`)? I'm actually [using it to show notifications](https://github.com/kristijanhusak/neovim-config/blob/master/nvim/lua/partials/plugins/orgmode.lua#L59) so if someone copied over my config to make sure it works for them.

Sure.

kristijanhusak:
Thanks!

jgollenz:
I like it, thanks a lot :) the only thing that bugs me is that the overlays are not pixel perfect. To my surprise (and shock!) this is consistent with the actual neovim logo. So cudos to @rwmpelstilzchen for being consistent with that :grin:

rwmpelstilzchen:
> I like it, thanks a lot :)

Thanks!! (⌒‿⌒)

> the only thing that bugs me is that the overlays are not pixel perfect. To my surprise (and shock!) this is consistent with the actual neovim logo. So cudos to @rwmpelstilzchen for being consistent with that 😁

Yup, I noticed that too… ^_^;
I’ll open an issue (or better, a pull request) on the NeoVim repository.

PR #616, opened 2023-09-25
test: Improve minimal init, test launching

treatybreaker:
Hey there! This PR "luafies" the tests and removes the need to `git clone` test dependencies ahead of time by doing the clone process in the minimal init.

- There is no longer a need to `git clone` any test dependencies, instead the minimal init handles that for you.
- The minimal init cleans up after itself as much as is reasonable. This ensures previous runs have no impact on newer runs
    - This only works on Mac & Linux as it uses `rm -rf` to clean up after itself. I know there's a bit of a push to get better windows support. I noticed the issues and I think I may have a solution in another PR. If the `rm -rf` doesn't run it's not a big deal for now.
- I removed the `tesfile` command and instead allowed a testfile to be specified for the `make test` command and if one is not found it fallbacks to tests/plenary
    - So now to launch a specific file you'd do `FILE=./tests/plenary/util_spec.lua make test` for instance 😉.

I know there weren't any open issues for this, but I found it slightly annoying to have to run `git clone` before doing any testing. If this PR is undesired feel free to reject it.


treatybreaker:
I see that the tests for Neovim `0.8.3` failed, investigating currently as to why.

treatybreaker:
I am having difficulties replicating the test failure for `0.8.3`. After installing Neovim from `0.8.3` my tests are passing locally.

treatybreaker:
Seems good now, the last change was only because I caught some caching in the workflow that may have been missed prior to that commit with my changes.

Not sure why `0.8.3` is working now though 🤷, it should be the same stuff

.gitignore
```diff
@@ -39,6 +39,5 @@ luac.out
 *.x86_64
 *.hex
 
-plenary.nvim/
-nvim-treesitter/
-mini.nvim/
+# Directory containing testing dependencies downloaded during test run
+.min/
```
kristijanhusak, line 43: Just curious, why `.min` ? Is it some kind of pattern?
I'd go for `.vendor` or `.deps` or something like that.

tests/minimal_init.lua
```diff
@@ -0,0 +1,106 @@
...
+
+  vim.env.XDG_CONFIG_HOME = M.root('xdg/config')
+  vim.env.XDG_DATA_HOME = M.root('xdg/data')
+  vim.env.XDG_STATE_HOME = M.root('xdg/state')
+  vim.env.XDG_CACHE_HOME = M.root('xdg/cache')
+
+  -- NOTE: Cleanup the xdg cache on exit so new runs of the minimal init doesn't share any previous state, e.g. shada
+  vim.api.nvim_create_autocmd('VimLeave', {
+    callback = function()
+      vim.fn.system({
```
kristijanhusak, line 71: We can use `vim.loop.fs_rmdir` here. That should also cover Windows support

kristijanhusak (changes requested):
Thanks for the PR! This looks great!
I think we can fix the Windows support (haven't tested it though) with the below comment.

.gitignore
treatybreaker: `.min` because it's for the minimal init 🤷. First thing that came to mind, I'll change it over to `.deps` I guess.

tests/minimal_init.lua
treatybreaker: `vim.loop.fs_rmdr` only works on empty directories. It's basically a clone of `rmdir(2)`. We'd need to iterate the entire directory and remove every file we come across, which *would* work,  but would be quite slow.

On windows we should be doing `Remove-Item -Recurse -Force DIR`, I'll just check if `powershell` is available and if it is then use that.

kristijanhusak: And what if we use [delete](https://neovim.io/doc/user/builtin.html#delete()) ? It accepts `d` and `rf` flags, so we could use that. It notes that Windows will not allowed deleting if it's used, but maybe at that moment it is no longer used.

treatybreaker: That works on Linux, based on the docs I'm gonna guess it works on Windows too. Will update shortly.

treatybreaker: See now.

.gitignore
treatybreaker: Done

kristijanhusak: You updated the code to point to `.deps`, but gitignore and git workflow still points to `.min`.

Makefile
```diff
@@ -1,7 +1,5 @@
 test:
-	nvim --headless --clean -u tests/minimal_init.vim -c "TSUpdateSync org" -c "PlenaryBustedDirectory tests/plenary/ {minimal_init = 'tests/minimal_init.vim'}"
-testfile:
-	nvim --headless --clean -u tests/minimal_init.vim -c "TSUpdateSync org" -c "PlenaryBustedDirectory $(FILE) {minimal_init = 'tests/minimal_init.vim'}"
+	nvim --headless --clean -u tests/test.lua "$(FILE)"
```
kristijanhusak, line 2: While we are at this, could we introduce a `make clean` command that would clean up the `.deps` and all other settings/deps for the tests? I had this issue before where my local tests were failing because I had outdated version of `nvim-treesitter` locally so I had to go inside it and do a `git pull`. If we introduce a `make clean`, we can run that and just re-run `make test` to start fresh.

CONTRIBUTING.md
```diff
@@ -25,13 +25,17 @@ Please document any new code you add with [emmylua annotations](https://emmylua.
...
 ```
-git clone https://github.com/nvim-treesitter/nvim-treesitter
-git clone https://github.com/nvim-lua/plenary.nvim
 make test
 ```
 
+To run a specific test you can set a `FILE` environment variable to a specific
+spec you want to test. Example:
+```
+TEST=./tests/plenary/api/api_spec.lua make test
```
kristijanhusak, line 36: `TEST` should be renamed to `FILE` since that's whats referenced in makefile.
We can also put it after the command which seems a bit more natural.

```suggestion
make test FILE=./tests/plenary/api/api_spec.lua 
```

kristijanhusak (changes requested):
Sorry for going back and forth. Just few more fixes/changes.

Makefile
treatybreaker: Sure, no problem.

Don't worry about "going back and forth", all part of the process. It would be worse if this stuff didn't get ironed out ahead of time. 😉

CONTRIBUTING.md
treatybreaker: Oops. Done.

Makefile
treatybreaker: Done, take a look. Instead of doing a git pull, I just nuke the directory. The test command will redownload it at startup if it doesn't exist.

treatybreaker:
Those "4 commits" are rebases over previous commits to address the previous comments while keeping the git history clean as a heads up.

treatybreaker:
*6 commits now :/

Missed the CI checkout for the deps. I haven't found anything else, unless you find something I *think* (see hope) this is good to go.

.github/workflows/tests.yml
```diff
@@ -30,12 +30,12 @@ jobs:
         uses: actions/checkout@v2
         with:
           repository: nvim-lua/plenary.nvim
-          path: plenary.nvim
+          path: tests/.deps/plenary
```
kristijanhusak, line 33: These paths are not correct. CI still clones the deps instead of using this one because it should be nested inside `plugins` directory (https://github.com/nvim-orgmode/orgmode/actions/runs/6313652906/job/17142196496?pr=616#step:7:8). 

Since we fetch the deps with the test script now, I think we can just remove these two steps and let the script clone everything that it needs.

kristijanhusak (changes requested):
Let's just clean up the workflow steps that are not necessary any more and I think it's good to go.

.github/workflows/tests.yml
treatybreaker: Sounds good, I'll remove them then.

treatybreaker: Done.

kristijanhusak (approved):
Everything looks good now, thanks!

PR #620, opened 2023-10-22
Refactor/omnicompletion

seflue:
This PR decouples the two modes of omnicompletion, make them easier to understand and reuse. It also adds additional tests and includes some minor improvements (see the particular commits).


lua/orgmode/org/autocompletion/omni.lua
```diff
@@ -33,8 +32,8 @@ local properties = {
 }
 
 local links = {
-  line_rgx = vim.regex([[\(\(^\|\s\+\)\[\[\)\@<=\(\*\|\#\|file:\)\?\(\(\w\|\/\|\.\|\\\|-\|_\|\d\)\+\)\?]]),
-  rgx = vim.regex([[\(\*\|\#\|file:\)\?\(\(\w\|\/\|\.\|\\\|-\|_\|\d\)\+\)\?$]]),
+  line_rgx = vim.regex([[\(\(^\|\s\+\)\[\[\)\@<=\(\*\|#\|file:\)\?\(\(\w\|\/\|\.\|\\\|-\)\+\)\?]]),
```
kristijanhusak, line 35: Why was this regex changed?

kristijanhusak (commented):
Generally looks ok. I'm just curious why was the regex changed?

lua/orgmode/org/autocompletion/omni.lua
seflue: I was hunting a bug (for which I currently only have a workaround), which I suspect to be in the regexp. Because of that I took a deeper dive into the regexp documentationand found out, that it is not the minimal regexp for what it tries to match. Some escapes are unnecessary and the `\d` and underscore were redundant, because they are already captured by `\w`. I actually tried to explain both simplifications in the commit message.

kristijanhusak: Thanks for the clarification!

kristijanhusak:
Thanks!

PR #621, opened 2023-10-22
fix(hyperlinks): link-follow for "file:../" and "../"

seflue:
Following a relative link starting with "../" is now supported.

Examples:
file:../somedir/some.org
../somedir/some.org

Previous it was possible to follow the first link, but a file "false" got opened instead of the target file. The second type was unsupported.

Because autocompletion works, it was easy for the user to create such links and surprising, that they didn't work.

### DIFF
--- a/lua/orgmode/objects/url.lua
+++ b/lua/orgmode/objects/url.lua
@@ -50,7 +50,7 @@ function Url:is_org_link()
 end
 
 function Url:is_file()
-  return self.str:find('^file:') or self.str:find('^./') or self.str:find('^/')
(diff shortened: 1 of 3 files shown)


kristijanhusak (changes requested):
Looks good! Can we just add a test to url_spec for this case?

seflue:
> Looks good! Can we just add a test to url_spec for this case?

Done.

kristijanhusak (approved):

kristijanhusak:
Thanks!

PR #623, opened 2023-10-31
feat: allow TS users to highlight marker delimiters

treatybreaker:
This permits those of us who use treesitter to change the highlights
surrounding markers. E.g. for the \"verbatim\" marker we can now highlight
the \"=\" characters individually instead of the entire chunk being the
same highlight.

This is particularly useful in tables with lots of code elements, this
makes it easier to tell the delimiter characters apart from the primary
marker.


treatybreaker:
Hello! This PR permits (as per the current commit) a user of orgmode to set the highlights for the delimiters around
markers.

Take for instance something like:

```org
=Hello world=
```

Prior to this PR those of us who used Treesitter with Nvim-Orgmode were unable to specify a specific highlight for the
equals sign, `=`. I have added a simple syntax match for those markers and bound them according to their main marker
group name. For instance, the `org_code` group also has `org_code_delimiter` now relating to its beginning and ending
delimiters.

I'm taking advantage of the `extmarks` that were set to highlight the entire block by using offsets. If you know of a
better way, feel free share! I'm willing to implement a better implementation if I can get a pointer; otherwise this
gets the job done.

lua/orgmode/colors/markup_highlighter.lua
```diff
@@ -423,8 +442,10 @@ local function apply(namespace, bufnr, line_index)
 end
 
 local function setup()
-  for _, marker in pairs(markers) do
-    vim.cmd(marker.hl_cmd)
+  for delimiter, marker in pairs(markers) do
+    vim.cmd(string.format(marker.hl_cmd, marker.hl_name))
+    vim.cmd(string.format('syn keyword %s_delimiter %s', marker.hl_name, delimiter))
```
kristijanhusak, line 447: I don't think we need this. `syn` is used for old Vim syntax highlighting.

lua/orgmode/colors/markup_highlighter.lua
```diff
@@ -423,8 +442,10 @@ local function apply(namespace, bufnr, line_index)
...
+    vim.cmd(string.format(marker.hl_cmd, marker.hl_name .. '_delimiter'))
```
kristijanhusak, line 448: We don't need delimiters highlight for `latex` type since we do not handle those markers as markup and we do not add extmarks for them.

kristijanhusak (changes requested):
Thanks for the PR!

Generally looks good, but it needs some minor cleanup.

lua/orgmode/colors/markup_highlighter.lua
treatybreaker: You are, in fact, correct. I had a misunderstanding on how highlights work in Vim... for several years now lol. Thanks for showing me that, will resolve ASAP.

treatybreaker: Done.

lua/orgmode/colors/markup_highlighter.lua
treatybreaker: Done, added an option in the markers to enable or disable the delimiter feature and do a check for it both in the `hl` group creation and in the `extmarks`.

You may actually want to review that in particular, it feels maybe kinda hacky to me or maybe I'm just stupid. Could be one or the other 🥲.

kristijanhusak (approved):
Looks good, thanks!

PR #624, opened 2023-10-31
feat: add a way to cache state between restarts of Neovim

treatybreaker:
This is useful for features such as `org-clock-in-last` that has the ability to persist its "knowledge" of the last clocked in task. Org mode does this by saving to a cache file on system. This commit gives `nvim-orgmode` the same capability.

I'm creating this PR in what I consider a partially finished state. I'm primarily looking for feedback as to the design of this module. In the scenario we're all happy with the design then I can get some simple tests written and then promote this to a non draft PR.

Some quick notes:

1. This state management system uses the in-built promises library that `nvim-orgmode` bundles. This means there's some additional validation and steps taken than what might exist in fully synchronous code. 
2. I use a metatable to ensure any accesses against the `State` table in which we assign new values are saved into the subtable `data`. This is so later on it is easy to determine what to save into the cache. Without it, I'd have to write additional logic explicitly excluding specific keys from the `State` table.
3. Seeing as the `State` object is a singleton is may seem weird to bother with a `State:new()` function. This actually is quite important, this sets up the metatable after all other functions are read in the file, ensuring they are not bound to the `data` subtable within the `State` object.
4. Reading the cache should work on all platforms as I'm "normalizing" the paths with `vim.fs.normalize`.

Any feedback (especially critical feedback) is desired on this. I'm looking to iron out kinks and find out if this is even desired 😃.

### DIFF
--- a/lua/orgmode/state/state.lua
+++ b/lua/orgmode/state/state.lua
@@ -150,4 +150,18 @@ function State:load_sync(timeout)
   return state
 end
 
+---Reset the current state to empty
+---@param overwrite? boolean Whether or not the cache should also be wiped
+function State:wipe(overwrite)
+  overwrite = overwrite or false
+
+  self.data = {}
+  self._ctx.curr_loader = nil
+  self._ctx.loaded = false
+  self._ctx.saved = false
+  if overwrite then
+    state:save_sync()
+  end
+end
+
 return State.new()
--- a/lua/orgmode/state/state.lua
+++ b/lua/orgmode/state/state.lua
@@ -47,9 +47,7 @@ function State:load()
 
   --- If we've already loaded the state from cache we don't need to do so again
   if self._ctx.loaded then
-    return Promise.new(function(resolve, _)
-      return resolve()
-    end)
+    return Promise.resolve()
   end
 
   self._ctx.curr_loader = utils.readfile(cache_path, { raw = true }):next(
--- a/lua/orgmode/state/state.lua
+++ b/lua/orgmode/state/state.lua
@@ -1,0 +1,100 @@
+local utils = require('orgmode.utils')
+local Promise = require('orgmode.utils.promise')
+
+local State = { data = {}, _ctx = { loaded = false, curr_loader = nil } }
+
+local cache_path = vim.fs.normalize(vim.fn.stdpath('cache') .. '/org-cache.json', { expand_env = false })
+--- Returns the current State singleton
+function State:new()
+  -- This is done so we can later iterate the 'data'
+  -- subtable cleanly and shove it into a cache
+  setmetatable(State, {
+    __index = function(tbl, key)
+      return tbl.data[key]
+    end,
+    __newindex = function(tbl, key, value)
+      tbl.data[key] = value
+    end,
+  })
+  -- Start trying to load the state from cache as part of initializing the state
+  self:load()
+  return self
+end
+
+---Save the current state to cache
+---@return Promise
+function State:save()
+  --- We want to ensure the state was loaded before saving.
+  return self:load():next(function(_)
+    utils.writefile(cache_path, vim.json.encode(State.data)):next(function(_) end, function(err_msg)
+      vim.schedule_wrap(function()
+        vim.notify('Failed to save current state! Error: ' .. err_msg, vim.log.levels.WARN, { title = 'Orgmode' })
+      end)
+    end)
+  end, function(_) end)
+end
+
+---Load the state cache into the current state
+---@return Promise
+function State:load()
+  --- If we currently have a loading operation already running, return that
+  --- promise. This avoids a race condition of sorts as without this there's
+  --- potential to have two State:load operations occuring and whichever
+  --- finishes last sets the state. Not desirable.
+  if self._ctx.curr_loader ~= nil then
+    return self._ctx.curr_loader
+  end
+
+  --- If we've already loaded the state from cache we don't need to do so again
+  if self._ctx.loaded then
+    return Promise.new(function(resolve, _)
+      return resolve()
+    end)
+  end
+
+  self._ctx.curr_loader = utils.readfile(cache_path, { raw = true }):next(
+    function(data)
+      local success, decoded = pcall(vim.json.decode, data, {
+        luanil = { object = true, array = true },
+      })
+      self._ctx.curr_loader = nil
+      if not success then
+        vim.schedule(function()
+          vim.notify('State cache load failure, error: ' .. decoded, vim.log.levels.WARN, {
+            title = 'Orgmode',
+          })
+          -- Try to 'repair' the cache by saving the current state
+          self:save()
+        end)
+      end
+      -- Because the state cache repair happens potentially after the data has
+      -- been added to the cache, we need to ensure the decoded table is set to
+      -- empty if we got an error back on the json decode operation.
+      if type(decoded) ~= 'table' then
+        decoded = {}
+      end
+
+      self._ctx.loaded = true
+      -- It is possible that while the state was loading from cache values
+      -- were saved into the state. We want to preference the newer values in
+      -- the state and still get whatever values may not have been set in the
+      -- interim of the load operation.
+      self.data = vim.tbl_deep_extend('force', decoded, self.data)
+      return self
+    end, ---@param err string
+    function(err)
+      -- If the file didn't exist then go ahead and save
+      -- our current cache and as a side effect create the file
+      if type(err) == 'string' and err:match([[^ENOENT.*]]) then
+        self:save()
+      else
+        -- If the file did exist, something is wrong. Kick this to the top
+        error(err)
+      end
+    end
+  )
+
+  return self._ctx.curr_loader
+end
+
+return State:new()
--- a/lua/orgmode/state/state.lua
+++ b/lua/orgmode/state/state.lua
@@ -35,7 +35,7 @@ function State:save()
       end)
       :catch(function(err_msg)
         vim.schedule_wrap(function()
-          vim.notify('Failed to save current state! Error: ' .. err_msg, vim.log.levels.WARN, { title = 'Orgmode' })
+          utils.echo_warning('Failed to save current state! Error: ' .. err_msg)
         end)
       end)
   end)
@@ -67,9 +67,7 @@ function State:load()
       if not success then
         local err_msg = vim.deepcopy(decoded)
         vim.schedule(function()
-          vim.notify('State cache load failure, error: ' .. vim.inspect(err_msg), vim.log.levels.WARN, {
-            title = 'Orgmode',
-          })
+          utils.echo_warning('State cache load failure, error: ' .. vim.inspect(err_msg))
           -- Try to 'repair' the cache by saving the current state
           self:save()
         end)
--- a/lua/orgmode/state/state.lua
+++ b/lua/orgmode/state/state.lua
@@ -27,18 +27,17 @@ end
 function State:save()
   State._ctx.saved = false
   --- We want to ensure the state was loaded before saving.
-  return self:load():finally(function()
-    utils
-      .writefile(cache_path, vim.json.encode(State.data))
-      :next(function()
-        State._ctx.saved = true
+  self:load()
+  return utils
+    .writefile(cache_path, vim.json.encode(State.data))
+    :next(function()
+      State._ctx.saved = true
+    end)
+    :catch(function(err_msg)
+      vim.schedule_wrap(function()
+        utils.echo_warning('Failed to save current state! Error: ' .. err_msg)
       end)
-      :catch(function(err_msg)
-        vim.schedule_wrap(function()
-          utils.echo_warning('Failed to save current state! Error: ' .. err_msg)
-        end)
-      end)
-  end)
+    end)
 end
 
 ---Load the state cache into the current state
--- a/lua/orgmode/state/state.lua
+++ b/lua/orgmode/state/state.lua
@@ -4,7 +4,9 @@ local Promise = require('orgmode.utils.promise')
 local State = { data = {}, _ctx = { loaded = false, saved = false, curr_loader = nil } }
 
 local cache_path = vim.fs.normalize(vim.fn.stdpath('cache') .. '/org-cache.json', { expand_env = false })
---- Returns the current State singleton
+
+---Returns the current State singleton
+---@return State
 function State.new()
   -- This is done so we can later iterate the 'data'
   -- subtable cleanly and shove it into a cache
--- a/lua/orgmode/state/state.lua
+++ b/lua/orgmode/state/state.lua
@@ -1,16 +1,17 @@
 local utils = require('orgmode.utils')
 local Promise = require('orgmode.utils.promise')
 
-local State = { data = {}, _ctx = { loaded = false, saved = false, curr_loader = nil, savers = 0 } }
+---@class OrgState
+local OrgState = { data = {}, _ctx = { loaded = false, saved = false, curr_loader = nil, savers = 0 } }
 
 local cache_path = vim.fs.normalize(vim.fn.stdpath('cache') .. '/org-cache.json', { expand_env = false })
 
----Returns the current State singleton
----@return State
-function State.new()
+---Returns the current OrgState singleton
+---@return OrgState
+function OrgState.new()
   -- This is done so we can later iterate the 'data'
   -- subtable cleanly and shove it into a cache
-  setmetatable(State, {
+  setmetatable(OrgState, {
     __index = function(tbl, key)
       return tbl.data[key]
     end,
@@ -18,7 +19,7 @@ function State.new()
       tbl.data[key] = value
     end,
   })
-  local self = State
+  local self = OrgState
   -- Start trying to load the state from cache as part of initializing the state
   self:load()
   return self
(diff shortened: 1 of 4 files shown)


treatybreaker:
@kristijanhusak I'm primarily interested in what you think. If the above all sounds good and the code looks solid, I intend to get this merged and then turn around and implement at least `org-clock-in-last` with persistent state between Neovim

kristijanhusak:
Thanks for the PR. I didn't give it a deeper look yet. I have one question regarding this:

> Org mode does this by saving to a cache file on system

Can you point me where I can read more about this? I wasn't aware it is done in Emacs, at least not in this way. I assumed Emacs has a more advanced persistence.

treatybreaker:
image

So emacs is a *little* bit fancier than the method done above to persist state. That `index` file contains:
```
((:container
  ((elisp org-element--headline-cache)
   (elisp org-element--cache))
  :persist-file "26/0e7168-1132-40fb-9c4a-f9527c58b0eb" :associated
  (:hash "db811d99c98367fd949e030078d6c3dc" :file "/home/sam/org/sprints.org" :inode 1798784)
  :expiry 30 :last-access 1698760936.4788425 :last-access-hr "2023-10-31T09:02:16-0500")
 (:container
  ((elisp org-element--headline-cache)
   (elisp org-element--cache))
  :persist-file "16/77b350-d23c-435f-a1af-79fab01dc901" :associated
  (:hash "9b28fb51f68a22e72bfe80dd6eea5713" :file "/home/sam/.notes" :inode 1798482)
  :expiry 30 :last-access 1698754134.958719 :last-access-hr "2023-10-31T07:08:54-0500")
 (:container
  ((elisp org-element--headline-cache)
   (elisp org-element--cache))
  :persist-file "32/2a423d-7966-474a-b285-bec583e15678" :associated
  (:hash "f821ab3b4dde77f42b322d56138e8b6f" :file "/home/sam/Notes/Whitelist-API.org" :inode 1853863)
  :expiry 30 :last-access 1698847955.2776916 :last-access-hr "2023-11-01T09:12:35-0500")
 (:container
  ((elisp org-element--headline-cache)
   (elisp org-element--cache))
  :persist-file "7a/71cd55-9320-4ced-8290-64f32f790869" :associated
  (:hash "0c752e61eadaa71b7e3408264c6a2349" :file "/home/sam/Notes/sprints.org" :inode 1709198)
  :expiry 30 :last-access 1698649783.8752766 :last-access-hr "2023-10-30T02:09:43-0500")
 (:container
  ((index "3.1"))
  :persist-file "c0/b60f88-9bc6-45e0-a7f9-711167b5b71d" :associated nil :expiry never :last-access 1698847955.2797225 :last-access-hr "2023-11-01T09:12:35-0500"))
```

Which definitely is more complex than what I'm implementing here.

treatybreaker:
@kristijanhusak

Found it, source code for `org-persist` can be found in a mirrored Github repository [here](https://github.com/bzg/org-mode/blob/main/lisp/org-persist.el).

Or you can directly get it by cloning from Savannah, `git clone git://git.sv.gnu.org/emacs/org-mode.git && cd org-mode && $EDITOR lisp/org-persist.el `. It has a fair bit of its own documentation in the top of the file. I was unable to access said docs in Emacs 🤷.

lua/orgmode/state/state.lua
```diff
@@ -0,0 +1,100 @@
...
+  --- promise. This avoids a race condition of sorts as without this there's
+  --- potential to have two State:load operations occuring and whichever
+  --- finishes last sets the state. Not desirable.
+  if self._ctx.curr_loader ~= nil then
+    return self._ctx.curr_loader
+  end
+
+  --- If we've already loaded the state from cache we don't need to do so again
+  if self._ctx.loaded then
+    return Promise.new(function(resolve, _)
```
kristijanhusak, line 50: You can just return `Promise.resolve()` here.

lua/orgmode/state/state.lua
```diff
@@ -0,0 +1,100 @@
...
+  self:load()
+  return self
+end
+
+---Save the current state to cache
+---@return Promise
+function State:save()
+  --- We want to ensure the state was loaded before saving.
+  return self:load():next(function(_)
+    utils.writefile(cache_path, vim.json.encode(State.data)):next(function(_) end, function(err_msg)
```
kristijanhusak, line 29: Handle the errors through the `:catch` function instead of 2nd param to `:next`.

```lua
utils.writefile(cache_path, vim.json.encode(State.data)):catch(function(err_msg)
      vim.schedule_wrap(function()
        vim.notify('Failed to save current state! Error: ' .. err_msg, vim.log.levels.WARN, { title = 'Orgmode' })
      end)
    end)

File: lua/orgmode/utils/init.lua [hunk 3]
```diff
@@ -6,7 +6,8 @@ local debounce_timers = {}
 local query_cache = {}
 local tmp_window_augroup = vim.api.nvim_create_augroup('OrgTmpWindow', { clear = true })
 
-function utils.readfile(file)
+function utils.readfile(file, opts)
+  opts = vim.tbl_deep_extend('keep', opts or {}, { raw = false })
```
kristijanhusak, line 10: Doing this should be enough, since `nil` is considered a falsy value.
```suggestion
  opts = opts or {}
```

kristijanhusak (commented):
I didn't give it a test, but I did review the code. 
Generally, it looks ok but I need to investigate a bit more to see what and how we should persist data.

Also, it would be great to write some tests around this to ensure it works as expected.

treatybreaker:
> Also, it would be great to write some tests around this to ensure it works as expected.

Absolutely. I just wanted to check that this looked good in the first place before writing up tests etc. only to be told that it's a no go.

I'll get those tests written after resolving your comments in a sec here.

lua/orgmode/state/state.lua
treatybreaker: Done.

lua/orgmode/state/state.lua
treatybreaker: Done.

treatybreaker:
Still need to write a test for the self-healing.

treatybreaker:
As a heads up @kristijanhusak this should be good to go. If you're happy with this, I'm going to begin implementing some missing orgmode features that can use persistence.

lua/orgmode/state/state.lua
```diff
@@ -0,0 +1,100 @@
...
+
+      self._ctx.loaded = true
+      -- It is possible that while the state was loading from cache values
+      -- were saved into the state. We want to preference the newer values in
+      -- the state and still get whatever values may not have been set in the
+      -- interim of the load operation.
+      self.data = vim.tbl_deep_extend('force', decoded, self.data)
+      return self
+    end, ---@param err string
+    function(err)
```
kristijanhusak, line 85: Move this also to a `:catch` so it is consistent with other error handlers in the file.

lua/orgmode/state/state.lua
```diff
@@ -0,0 +1,100 @@
...
+      -- If the file didn't exist then go ahead and save
+      -- our current cache and as a side effect create the file
+      if type(err) == 'string' and err:match([[^ENOENT.*]]) then
+        self:save()
```
kristijanhusak, line 89: You can return early here to avoid `else` statement:

```suggestion
        return self:save()
```

lua/orgmode/state/state.lua
```diff
@@ -0,0 +1,100 @@
...
+        -- If the file did exist, something is wrong. Kick this to the top
+        error(err)
+      end
+    end
+  )
+
+  return self._ctx.curr_loader
+end
+
+return State.new
```
kristijanhusak, line 100: Since this is a singleton we should return the instance of it here `return State.new()`

kristijanhusak (changes requested):
We're almost there, just few minor changes and it's good to go.

lua/orgmode/state/state.lua
treatybreaker: Done.

lua/orgmode/state/state.lua
treatybreaker: Done.

lua/orgmode/state/state.lua
treatybreaker: Done.

treatybreaker:
@kristijanhusak Let me know if you see anything else.

kristijanhusak:
@treatybreaker how did you test this manually? 
I tried setting up a keymap like this to ensure it's working but it doesn't seem to do anything:

```lua
vim.keymap.set('n', '<leader>u', function()
  local state = require('orgmode.state.state')
  return state:load():finally(function()
    print('LOADED', state.data)
    state.field = 'value'
    state:save()
  end):next(function()
    print('SAVED')
    vim.print('UPDATED STATE', state.data)
  end)
end)
```

Prints are never shown, and `org-cache.json` file is never created.
Also, tests are not actually checking anything. For example, if you change this line https://github.com/nvim-orgmode/orgmode/pull/624/files#diff-23e5eb566ad6dac12d0c71f0d549d5e273bd9ed2c79b837688a0ad21e0946dd7R52 to expect something that it should not get, tests still pass:

```lua
-- This passes, even though it shouldn't
assert.are.equal(state.my_var, 'hello world foo')
```

Something seems to be off. Can you check? Thanks.

treatybreaker:
I was testing by shoving `require("orgmode.state.state")` into Orgmode's top level `init.lua`.

I believe I found the issue which was caused by the early return. I remember not doing an early return for a reason, I just couldn't remember why. Now I do.

In terms of testing, I believe the tests aren't valid because they're expecting a synchronous context and since the promises are in the "background" the tests complete before the actual callbacks have finished. 🫤

lua/orgmode/state/state.lua
treatybreaker: Undone, this causes a loop with `self:load`.

treatybreaker:
Alright, I've fixed some functionality in the `State:load` function so it all should be good there. Works on my side.

Now I get to figure out async testing with plenary.

treatybreaker:
@kristijanhusak 

This was uhhh, painful. Async code in lua is the very definition of not fun.

I think this should be good now, I have the tests responding correctly now by abusing `vim.wait`. And I did some minor fixes to the State module. One of those is tracking the `State:saved` context which particularly useful for tests.

I am *really* hoping this is good at this point. If it's not then I'm going to keep on grinding until it's in a good state.

Take a look and let me know your findings.

treatybreaker:
As a note, Neovim didn't seemingly create those `stdpaths` at startup with the minimal init... I had to create them during the minimal init setup.

kristijanhusak:
Awesome, now it works :)

`vim.wait` is a way to go. I would even suggest introducing `load_sync()` and `save_sync()` as part of the State class, where it would do the same thing you do in tests. Then you can leverage that in tests and we also have it as an option if needed.
Feel free to increase timeout to 1000 and use an interval of `5`. I did [similar thing](https://github.com/nvim-orgmode/orgmode/blob/master/lua/orgmode/parser/files.lua#L311-L318) to ensure that org files are loaded.

treatybreaker:
@kristijanhusak 

Added those functions as well as a `State:wipe` function. In the State module I had to do some fancy stuff with `State:save` as I realized there can be multiple saves running as once. Tracking this in the context now. For the `sync` functions I went with an interval of `20` with a default timeout of `500` ms. If that's too low, let me know and I can do a quick commit to change that.

Man... async code and multiple execution is hell on earth. Very useful, but my god.

Looking forward to more hellish findings 😉.

treatybreaker:
Working on replicating the failure in `0.9.1`.

treatybreaker:
Fixed.

lua/orgmode/state/state.lua
```diff
@@ -0,0 +1,167 @@
+local utils = require('orgmode.utils')
+local Promise = require('orgmode.utils.promise')
+
+local State = { data = {}, _ctx = { loaded = false, saved = false, curr_loader = nil, savers = 0 } }
```
kristijanhusak, line 4: Add a type to the class so it's properly read by lua ls. You can also rename everything from `State` to `OrgState`, but not necessary. I just want to make type prefixed to not potentially clash with other plugins.
```suggestion
---@class OrgState
local State = { data = {}, _ctx = { loaded = false, saved = false, curr_loader = nil, savers = 0 } }
```

tests/plenary/state/state_spec.lua
```diff
@@ -0,0 +1,144 @@
+local utils = require('orgmode.utils')
+local state = nil
+local cache_path = vim.fs.normalize(vim.fn.stdpath('cache') .. '/org-cache.json', { expand_env = false })
+
+describe('State', function()
+  before_each(function()
+    -- Ensure the cache file is removed before each run
+    state = require('orgmode.state.state')
+    vim.wait(50, function()
```
kristijanhusak, line 9: Do we still need this?

tests/plenary/state/state_spec.lua
```diff
@@ -0,0 +1,144 @@
...
+      return state._ctx.saved
+    end, 10)
+    state:wipe()
+    vim.fn.delete(cache_path, 'rf')
+  end)
+
+  after_each(function()
```
kristijanhusak, line 16: I think we don't need this anymore. Before_each does the wipe and sets these variables to false.

tests/plenary/state/state_spec.lua
```diff
@@ -0,0 +1,144 @@
+local utils = require('orgmode.utils')
+local state = nil
```
kristijanhusak, line 2:
```suggestion
---@type OrgState
local state = nil
```

tests/plenary/state/state_spec.lua
```diff
@@ -0,0 +1,144 @@
...
+  it("should create a state file if it doesn't exist", function()
+    -- Ensure the file doesn't exist
+    vim.fn.delete(cache_path, 'rf')
+    local stat = vim.loop.fs_stat(cache_path)
+    if stat then
+      error('Cache file existed before it should! Ensure it is deleted before each test run!')
+    end
+
+    -- This creates the cache file on new instances of `State`
+    state:load()
```
kristijanhusak, line 32: Use `load_sync` here and remove the `vim.wait` below.
```suggestion
    state:load_sync()
```

tests/plenary/state/state_spec.lua
```diff
@@ -0,0 +1,144 @@
...
+  after_each(function()
+    state._ctx.saved = false
+    state._ctx.loaded = false
+    state = nil
+    -- vim.fn.delete(cache_path, 'rf')
+  end)
+
+  it("should create a state file if it doesn't exist", function()
+    -- Ensure the file doesn't exist
+    vim.fn.delete(cache_path, 'rf')
```
kristijanhusak, line 25: This is already done in `before_each` so there's no need to  delete. You can leave the check below just in case.

kristijanhusak (changes requested):
Looks good! Just some minor tweaks to types and tests and it's good to go.

tests/plenary/state/state_spec.lua
treatybreaker: Negative, the `state:wipe` call below it nukes it out synchronously. Will remove.

tests/plenary/state/state_spec.lua
treatybreaker: You're right, will remove.

tests/plenary/state/state_spec.lua
treatybreaker: So this test is actually not checking the precise usage of `State:load`. It's validating that the cache file is created by `State:save` via a `State:load` call. As such `State:load_sync` doesn't actually have any value here.

This test is kinda peeking under the hood a bit to validate some "magic" functionality of `State:load`. It's important that the `vim.wait` stays. Notice that it's checking the status of `state._ctx.saved`, not `state._ctx.loaded`.

tests/plenary/state/state_spec.lua
treatybreaker: Yep, will remove.

tests/plenary/state/state_spec.lua
treatybreaker: Done. I also renamed all the `State` objects to `OrgState`.

tests/plenary/state/state_spec.lua
kristijanhusak: Ok, that makes sense. Thanks for clarifying!

lua/orgmode/state/state.lua
treatybreaker: Done. I also renamed the `State` object to `OrgState`.

tests/plenary/state/state_spec.lua
treatybreaker: Done

tests/plenary/state/state_spec.lua
treatybreaker: Done

tests/plenary/state/state_spec.lua
treatybreaker: Done.

treatybreaker:
Let me know if you see anything else 🙂

kristijanhusak (approved):
Ok, everything looks great now, thanks!

treatybreaker:
I'll look into implementing some features that take advantage of this come this weekend or next week as a heads up.

PR #627, opened 2023-11-09
Add Virtual Indent

treatybreaker:
This is largely a carry on from #561 with additional fixes and logic. I'm assuming that PR is likely dead in the water seeing as no further progress has been made since May.

Notes:
1. We *must* use `nvim_buf_attach` for the `on_lines` callback in `virtual_indent.lua`. Ephemeral marks with inline virtual text is *not* supported. It was accidentally supported by Neovim for a few dev versions previously. As such we have to use `nvim_buf_attach` so we can call `nvim_buf_del_extmark`. Using `nvim_set_decoration_provider` will error with `nvim_buf_del_extmark`.
2. I changed `org_indent_mode` to a pure boolean value like in Emacs Orgmode. I added a simple wrapper to handle folks who have `noindent` and `indent` in their configs set to that var.
3. I implemented support for realigning items in promotions and deletions.
4. You'll notice that I have a `hack` as part of the `on_lines` callback in `virtual_indent.lua`. This doesn't cause any weird behavior, but it sure does look messy. I had to do that to avoid some issues with the `on_lines` callback being triggered early (as was the case in #561).
5. ~**This is the one where I'm stumped**. Promotions and demotions on already indented lists (the list root isn't aligned at column 0) does not fully reindent that list. It only reindents the root list node and subsequent demotions & promotions reindent down the subnodes until it's all flush. Not sure how to fix this. Would like some advice here for sure. It's not a big deal imo, it's simple enough to format the lines with visual mode, but I'm hoping to avoid that as it causes the cursor to move during a promotion/demotion.~

Problem exposed in point 5 has been resolved by a rebase on top of #629

@kristijanhusak Let me know what you think, and I'd really like some feedback on the fifth note there. Once I can resolve the fifth note and you like the core of this, then I'll promote this to a full PR. 🙂


treatybreaker:
Cannot replicate the test failures above... works as expected locally. Seems the `todo_spec` is exploding because the timestamp is ever so slightly different on the properties.

treatybreaker:
Resolved test failures, fixed some logic errors. Mostly to do with respecting "legacy" behavior in promotions & demotions as well as correctly checking `vim.version()` to enable VirtualIndent.

treatybreaker:
Rebased on top of #629

treatybreaker:
TODO: Write tests to validate virt cols and then this should be ready.

treatybreaker:
@kristijanhusak 

I'll mark this ready for review if #629 gets merged. If it doesn't, then I'll need to undo my rebase here and that will reintroduce the issues in the fifth point. As is it stands though, if #629 gets merged this should be good on point 5.

treatybreaker:
I realized I totally forgot about this! I had a reminder pop up on my phone today.

I'm going to pick up where I left off some point this week and update/fix what needs fixing with any changes on master so this can get integrated. My fault, totally forgot. Dementia sets in early apparently.

treatybreaker:
I am unable to replicate the test failures occurring here in the CI. When using Neovim version `0.9.0` & `0.9.1` the tests pass as expected, even on multiple runs.

treatybreaker:
@kristijanhusak Do you have any ideas as to why these tests are inconsistent? I'm kinda grasping at straws here -- even recompiling Neovim now for the failed version(s) makes those tests work as expected locally.

The moment I can get these tests playing nice I can probably promote this from draft status.

kristijanhusak:
@treatybreaker I re-ran them now manually and it passed. I'm not sure why it's not consistent. We have some random failures in case tests overlap between 2 minutes and assertion fails, but these failed because some indentation was not properly done in the assertion. See if you can reproduce it locally by running tests multiple times in a row.

treatybreaker:
I can cause it to happen semi-consistently by running this:

> See if you can reproduce it locally by running tests multiple times in a row.

I can with the following script:
```bash
TEST_RUN=1
while :; do
  printf "==== Current Test Run: %s ====\n" "${TEST_RUN}"
  if make test; then
    printf "\n\n++++ Finished Test Run: %s ++++\n\n" "${TEST_RUN}"
  else
    printf "\n\n==== Tests Failed on Run %s ====\n\n" "${TEST_RUN}"
    break
  fi
  ((TEST_RUN++))
done
```

Might need to run for a while. It's always the same test that fails: `with "indent", "0gg=G" reindents the whole file`. I wonder if the test file is loaded fully or if Orgmode is 100% ready when the full reindent is called. Seems to be inconsistent if it fails, but I'm not 100% certain as to why. 

Definitely an issue with this PR, I've let that script run for a while on the master branch -- no issues. Gonna dig in and find out what's going on.

treatybreaker:
Correction, I can cause this failure *consistently* on master.

kristijanhusak:
@treatybreaker I'm assuming your previous PR causes the issue https://github.com/nvim-orgmode/orgmode/pull/629 . Can you check there what's happening?

treatybreaker:
With the seeming resolution of the bug above in #640, I'm going to promote this from a draft.

I'm ready for some nitpicks.

kristijanhusak (changes requested):
Thanks for the PR. I gave this a quick test, it looks ok, but I will test it thoroughly once we align on the comment below.
`org_indent_mode` in here is a bit confusing at this moment. This is how I initially thought it should work:
* `indent` means that it should apply real indentation to everything with spaces, as it is doing now by default
* `noindent` - Do not apply any indentations with spaces, and let the virtual indentation do the indentation.

I'm aware this is not correct so we should follow what Emacs does. This is how it is actually enabled from the configuration in Emacs: https://stackoverflow.com/a/66793922

treatybreaker:
Hey, I'll do my best to respond and resolve some issues in this PR on Friday. Been quite busy recently. Just a heads up 🙂.

kristijanhusak:
No worries. There's no rush on this, it's only on nightly anyway 😄

DOCS.md
```diff
@@ -266,11 +266,11 @@ Possible values:
   * between `\(` and `\)` delimiters - example: `\( b=2 \)`
 
 #### **org_indent_mode**
-*type*: `string`<br />
-*default value*: `indent`<br />
+*type*: `boolean`<br />
+*default value*: `true`<br />
```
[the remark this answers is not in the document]
treatybreaker: Done, I outright removed `org_indent_mode` and shifted that to a buffer variable: `vim.b.org_indent_mode`. I had to add a hack into the test `minimal_init.lua` to enable indents in the tests. Almost all the tests were written with indent mode being enabled in mind 🤷. Quite annoying.

There still isn't detach functionality implemented, so if they disable indent mode it won't apply until the buffer is completely closed and reopened.

treatybreaker:
Question, do we need to enable a different variable or something to allow "raw" indents in the forms of spaces instead of Virtual? If so, where would you recommend placing such an option?

kristijanhusak:
@treatybreaker you mean to convert virtual indents to actual spaces, or a command to toggle between virtual indent and regular indent? For former, we don't need to do anything. Latter, maybe later, I don't think we need it at this moment.

treatybreaker:
> @treatybreaker you mean to convert virtual indents to actual spaces, or a command to toggle between virtual indent and regular indent? For former, we don't need to do anything. Latter, maybe later, I don't think we need it at this moment.

This PR currently removes the functionality of regular indents for nightly users. There's setting to use back regular indents on nightly in this PR. There's only virtual or no indents right now.

_Regular indents meaning spaces in front of the text as part of the saved file I'm assuming._

kristijanhusak:
This is how it should work:
* `org_startup_indented = false` (default) - Work as it is working now on master. Which means, spaces are inserted for indents where applicable
* `org_startup_indented = true` - Enables Virtual indents, and does not add any spaces for indentation, except for list items (and maybe something else, I'm not 100% sure).

Does that answer your question?

treatybreaker:
> This is how it should work:
> 
> * `org_startup_indented = false` (default) - Work as it is working now on master. Which means, spaces are inserted for indents where applicable
> 
> * `org_startup_indented = true` - Enables Virtual indents, and does not add any spaces for indentation, except for list items (and maybe something else, I'm not 100% sure).
> 
> 
> 
> Does that answer your question?

Kind of. 

This is a breaking change right now. Under the model above you have there there isn't a way to disable indentation whether virtual or literal. Emacs, by default, doesn't have indents at all.

Do I need to add another setting to disable indents entirely?

kristijanhusak:
Ok sorry, I missed one configuration that I had set.
`org-adapt-indentation` enables the "regular" indentation: https://orgmode.org/manual/Hard-indentation.html
This is off by default in Emacs.
To prevent breaking things with this PR, I would say that we should add that option to the configuration and make it `true` by default:
```
org_adapt_indentation = true
```

So to summarize, this is how it should work:
* defaults (`org_startup_indented = false`, `org_adapt_indentation = true`) - Behave how the `master` branch behaves now
* `org_startup_indented = true`, `org_adapt_indentation = true` - Use Virtual indents, and do not add any actual spaces for intentations, with the exceptions I mentioned in my previous comment
* `org_startup_indented = false`, `org_adapt_indentation = false` - Do not add any spaces for indentation (with exceptions), and do not add any virtual indentation. Basically how Emacs behaves by default.

Sorry about this, I forgot some information I collected a long time ago around this.

treatybreaker:
Yeah, that's what I was wondering about. When I get home today I'll get this implemented.

On Jan 24, 2024, at 1:05 PM, Kristijan Husak ***@***.***> wrote:

﻿


Ok sorry, I missed one configuration that I had set.
org-adapt-indentation enables the "regular" indentation: https://orgmode.org/manual/Hard-indentation.html <https://orgmode.org/manual/Hard-indentation.html> 
This is off by default in Emacs.
To prevent breaking things with this PR, I would say that we should add that option to the configuration and make it true by default:


org_adapt_indentation = true


So to summarize, this is how it should work:

*	defaults (org_startup_indented = false, org_adapt_indentation = true) - Behave how the master branch behaves now
*	org_startup_indented = true, org_adapt_indentation = true - Use Virtual indents, and do not add any actual spaces for intentations, with the exceptions I mentioned in my previous comment
*	org_startup_indented = false, org_adapt_indentation = false - Do not add any spaces for indentation (with exceptions), and do not add any virtual indentation. Basically how Emacs behaves by default.

Sorry about this, I forgot some information I collected a long time ago around this.

treatybreaker:
@kristijanhusak 

Take a look now. `org_adapt_indentation` is now implemented.

tests/minimal_init.lua
```diff
@@ -134,4 +134,6 @@ require('nvim-treesitter.configs').setup({
 require('orgmode').setup({
   org_agenda_files = { base_root_path .. '/plenary/fixtures/*' },
   org_default_notes_file = base_root_path .. '/plenary/fixtures/refile.org',
+  -- Many of the tests were originally written with indent mode being enabled in mind :/
+  org_startup_indented = true,
```
kristijanhusak, line 138: We should be able to leverage new `org_adapt_indentation` option here, right?

tests/plenary/ui/mappings/headline_spec.lua
```diff
@@ -90,26 +91,52 @@ describe('Heading mappings', function()
...
-      '*** TODO [#A] Test orgmode level 2 :PRIVATE:',
-      ' Some content for level 2',
-      '**** NEXT [#1] Level 3',
-      ' Content Level 3',
-    }, vim.api.nvim_buf_get_lines(0, 2, 8, false))
+    assert.are.same(check, vim.api.nvim_buf_get_lines(0, 2, 8, false))
 
     -- Support count
+    local check
+    if VirtualIndent.enabled then
```
kristijanhusak, line 119: This is done to make tests pass on 0.9.* and 0.10 ?

DOCS.md
```diff
@@ -265,12 +265,23 @@ Possible values:
...
-* `noindent` - Disable indentation. All lines start from 1st column
+* `true` - Enables indent mode on startup. Lines that follow headlines are indented to that headline level.
+* `false` - Disables indent mode on startup. All lines start from the first column.
+
+#### **org_adapt_indentation**
+
+*type*: `boolean`<br />
+*default value*: `true`<br />
+Possible values:
+* `true` - Converts indents to *Virtual* indents. The indents are only visual, they are not saved  to the file.
```
kristijanhusak, line 281: Values are reverted here. When it's `true`, it should use real indents.
In Emacs, enabling `org-indent-mode` automatically sets this value to `false` ([docs](https://orgmode.org/manual/Org-Indent-Mode.html#:~:text=By%20default%2C%20Org%20Indent%20mode%20turns%20off%20org%2Dadapt%2Dindentation)), but we will not actually modify any of the settings, just consider a priority.

If `org_startup_indented = true`, always use virtual indents.
If `org_startup_indented = false`, follow `org_adapt_indentation` according to the describe logic above.

lua/orgmode/config/init.lua
```diff
@@ -406,7 +415,7 @@ end
 ---@param amount number
 ---@return string
 function Config:get_indent(amount)
-  if self.opts.org_indent_mode == 'indent' then
+  if vim.b.org_indent_mode then
```
kristijanhusak, line 418: Are you sure we will always have access to the buffer variable?

I would prefer if we stick only with configuration checks for now, and we can later add the buffer level toggling.

kristijanhusak (changes requested):
We're getting there, but I think the logic is mixed up a bit.
Lets just address these comments and go from there.

lua/orgmode/config/init.lua
treatybreaker: It makes it much easier to use the buffer variable later on when we add toggling functionality.

I figured I'd bite the bullet now and take care of it as it's pretty fundamental to how it works.

If you'd prefer for me to undo that and have a separate PR with those checks I can do that. I have tests locally that have been rewritten now that support the buffer variable and they're all green currently.

tests/plenary/ui/mappings/headline_spec.lua
treatybreaker: Yes, and I'll refactor it to not check that variable and instead rely on the buffer option (depending on your decision to [this comment](https://github.com/nvim-orgmode/orgmode/pull/627#discussion_r1465601412)) in combination with `org_adapt_indentation`.

tests/minimal_init.lua
treatybreaker: Unfortunately not. `org_adapt_indentation` purely attempts to enable Virtual Indent if indent mode is on.

This enables the normal raw indents (and Virtual in some cases) that we need for many of the tests that were written with them in mind.

We actually need `org_startup_indented` set to true for *many* of the tests to work. Should I set `org_startup_indented` to `true` by default with that in mind?

DOCS.md
treatybreaker: Ah, I see... https://orgmode.org/manual/Hard-indentation.html. I'm a bit off, I already have a lot of this in my Emacs config that I'm too used to 🙂. 

I'll correct this momentarily.

treatybreaker: @kristijanhusak 

Done. I made `adapt_indentation` *very* similar to Orgmode.

When `org_startup_indented` is disabled and `org_adapt_indentation` is enabled you only get hard indentation, just like on `master` currently.

When `org_startup_indented` is enabled and `org_adapt_indentation` is disabled you get only Virtual Indentation.

When `org_startup_indented` is true and `org_adapt_indentation` is enabled you get *both* Virtual indentation *and* hard indents like Emacs.

tests/plenary/ui/mappings/headline_spec.lua
treatybreaker: Refactored.

tests/minimal_init.lua
treatybreaker: Deleted comment was not valid. Working on this now.

treatybreaker: We now leverage `org_adapt_indentation` by default as `org_adapt_indentation` is *enabled* by default.

lua/orgmode/config/init.lua
treatybreaker: The latest commits are making heavy use of the `vim.b.org_indent_mode` variable, it's not that hard to substitute it for something else if you'd prefer me to not use a buffer variable for now. I can definitely use a pure config variable or something similar.

treatybreaker:
The latest commit did a refactor to make `org_adapt_indentation` responsible for hard indentation.

Not sure if I need to add something that disables it when `org_startup_indent` is enabled.

Copying from a previous reply of mine as to how `org_adapt_indentation` now works:

> When `org_startup_indented` is disabled and `org_adapt_indentation` is enabled you only get hard indentation, just like on `master` currently.
> 
> When `org_startup_indented` is enabled and `org_adapt_indentation` is disabled you get only Virtual Indentation.
> 
> When `org_startup_indented` is true and `org_adapt_indentation` is enabled you get _both_ Virtual indentation _and_ hard indents like Emacs.

kristijanhusak (changes requested):
I gave it a test, it works great, and code looks good overall.
 I only noticed one inconsistency.

If `org_startup_indented = true` and `org_adapt_indentation = true`, when I open up the file and reindent the file, it doesn't change any indentation.

For example, I have this content that is hard indented:
```
* TODO Test
  Test
  List
  - First
  - second
    - level2
      - level3
    - level2 item
```

Now when I apply above configuration (setting both to `true`) and open it, I get this:

virtual indent + hard indent:
```
* TODO Test
    Test
    List
    - First
    - second
      - level2
        - level3
      - level2 item
```

This is working as expected. But now, if I do gg=G, no indentation is actually changed.
I would expect to realign everything so it adapts to the virtual indentation.
It does that when `org_adapt_indentation = false`, so it generally works. We just need to tweak the checks.

AFAIK Emacs doesn't have this idea of (hard) reindentation, so this is something we can figure out ourselves.

If we follow the default Emacs settings for org indent mode I mentioned previously [link](https://orgmode.org/manual/Org-Indent-Mode.html#:~:text=By%20default%2C%20Org%20Indent%20mode%20turns%20off%20org%2Dadapt%2Dindentation), which states that when org-indent-mode is enabled on buffer, it disables the `org-adapt-indentation`, we can expect to apply the reindentation logic I mentioned above.

So `org_startup_indented` always has higher priority over `org_adapt_indentation`, until we maybe introduce `org_indent_mode_turns_off_org_adapt_indentation` mentioned [here](https://orgmode.org/manual/Org-Indent-Mode.html#:~:text=If%20you%20want%20to%20customize%20this%20default%20behavior%2C%20see%20org%2Dindent%2Dmode%2Dturns%2Don%2Dhiding%2Dstars%20and%20org%2Dindent%2Dmode%2Dturns%2Doff%2Dorg%2Dadapt%2Dindentation.), but we'll leave that as an improvement for later.

treatybreaker:
@kristijanhusak 

I'm on mobile currently so I can't make any changes; however, if you want to wait until 2 - 3pm CST I can add in disabling adapt indentation when indent mode is enabled by default along with the new flag you mention.

If not, I'm more than happy to just create a follow up PR with the functionality implemented then. LMK

kristijanhusak:
> I'm on mobile currently so I can't make any changes; however, if you want to wait until 2 - 3pm CST I can add in disabling adapt indentation when indent mode is enabled by default along with the new flag you mention.

Don't worry about this, work on it when you get a chance. Once we merge it in it will be considered experimental for some time until we flush out any potential issues. Ideally, I would put this out only when Neovim 0.10 comes out, but I don't think we can rely on that.

treatybreaker:
@kristijanhusak 

`org_startup_indented` now disables `org_adapt_indentation` by default, with a configuration option `org_indent_mode_turns_off_org_adapt_indentation` that is enabled by default. If that is set to false, then `org_adapt_indentation` won't be disabled when `org_startup_indented` is used.

> AFAIK Emacs doesn't have this idea of (hard) reindentation, so this is something we can figure out ourselves.

In Emacs, if I set `org-adapt-indentation` to `headline-data` then the behavior you're seeing with the hard indents and virtual indents is identical to how it's implemented in this PR. Same thing happens if I set `org-adapt-indentation` to `t` in Emacs as well. 

Not sure if I need to do anything else around indent adaption seeing as it follows Emacs now from what I can tell.

treatybreaker:
What else needs to be done for the Virtual Indents?

Off the top of my head:
- [ ] Toggling indent mode on or off in a buffer
- [ ] Toggling `org_adapt_indentation` on or off in a buffer
- [ ] Support `#+STARTUP: indent` & `#+STARTUP: noindent`

t.org
kristijanhusak, line 1: What's the purpose of this file? I don't see it being used anywhere.

ftplugin/org.lua
```diff
@@ -10,6 +10,12 @@ config:setup_mappings('org')
 config:setup_mappings('text_objects')
 config:setup_foldlevel()
 
+if config.org_startup_indented then
+  config.org_adapt_indentation = not config.org_indent_mode_turns_off_org_adapt_indentation
```
kristijanhusak, line 14: Let's move this line to `Config:extend` where we also do some validation and changes to the config.

kristijanhusak (changes requested):
Everything works great now! Just some minor changes.

Regarding the changes, I'm not sure about this:
> Toggling org_adapt_indentation on or off in a buffer

Does Emacs support this? I know it can toggle org-indent-mode, but I don't think it can change this setting in the same way.

t.org
treatybreaker: None, I accidentally included in a `git add`.

Removing it.

ftplugin/org.lua
treatybreaker: Will do. I do have to keep the conditional there though, as we use the `ftplugin` to set the buffer `org_indent_mode` variable as a heads up.

treatybreaker:
> Does Emacs support this? I know it can toggle org-indent-mode, but I don't think it can change this setting in the same way.

Open any org file in emacs and enable `org-indent-mode`. Then you can use `<M-x>eval-expression` to evalute `(setq org-adapt-indentation t)`.

That's the "toggle" I'm talking about.

kristijanhusak:
I don't think we need to worry about this. That's changing a config inline. 

Just one more thing please. Let's squash the commits into 1 or 2. There are > 30 commits , so it will not look helpful once someone looks at it in the changes.

treatybreaker:
@kristijanhusak 

Squashed all down into one commit. Sorry about how many there were 😅. Lots of tiny things got put together and I had to rebase/merge things done on master a few times.

kristijanhusak (approved):
Ok, I think this is good to go.
Thank you very much! You did some huge work here.

treatybreaker:
@kristijanhusak 

As a heads up, I'm sure folks are going to find issues that I haven't been able to. I've been daily driving off this branch since I opened this PR and it's worked relatively well for me, but I'm only one person with a single config.

If an issue comes in for the virtual indents, I'm not gonna leave you hanging -- ping me (@treatybreaker) on those issues I'll make them my problem.

PR #632, opened 2023-11-25
refactor(table): expose TsTable and reorganize table utility functions

joaomsa:
I decided to make this small refactor mainly for my own benefit. I wanted to utilize the treesitter TsTable helper in order to create a command for transposing tables (similar to `org-table-transpose-table-at-point` in Emacs). However, it was challenging to implement it because the TsTable object was not accessible from that module.

I imagine this litle helper may not be widely applicable enough to be included in the mainline code. Nevertheless, I'd love if more parts of the API exposed similar "from_current_node" or "at_cursor" functionality to facilitate ad-hoc scripting (even without committing to a stable API).

```lua
local function transpose(data)
  local data_t = {}
  for r, row in ipairs(data) do
    for c, cell in ipairs(row) do
      data_t[c] = data_t[c] or {}
      data_t[c][r] = cell
    end
  end
  return data_t
end

local function org_table_transpose()
  local ts_table = require('orgmode.treesitter.table')
  local Table = require('orgmode.parser.table')

  local ts_tbl = ts_table.from_current_node()
  if not ts_tbl then
    return
  end

  local data = vim.tbl_filter(function(row)
    return row ~= 'hr'
  end, ts_tbl.data)

  local data_t = transpose(data)
  local tbl_t = Table.from_list(data_t,
    ts_tbl.tbl.start_row,
    ts_tbl.tbl.start_col)

  local view = vim.fn.winsaveview()
  vim.api.nvim_buf_set_lines(0,
    ts_tbl.tbl.range.start_line - 1,
    ts_tbl.tbl.range.end_line,
    false, tbl_t:draw())
  vim.fn.winrestview(view)
end

vim.api.nvim_create_user_command("OrgTableTranspose", org_table_transpose, {})
```

### DIFF
--- a/lua/orgmode/org/format.lua
+++ b/lua/orgmode/org/format.lua
@@ -1,5 +1,5 @@
 local Files = require('orgmode.parser.files')
-local Table = require('orgmode.treesitter.table')
+local ts_table = require('orgmode.treesitter.table')
 local ts_org = require('orgmode.treesitter')
 
 local function format_line(linenr)
@@ -13,7 +13,9 @@ local function format_line(linenr)
     end
   end
 
-  if Table.format(linenr) then
+  local tbl = ts_table.from_current_node(linenr)
+  if tbl then
+    tbl:reformat()
     return true
   end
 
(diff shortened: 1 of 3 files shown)


kristijanhusak (changes requested):
This is ok, even though I would not suggest using internals.
Let's just do a small change in the mappings and it's good to go.

kristijanhusak:
You can now access the current table like this:

```lua
local tbl = require('orgmode.files.elements.table').from_current_node()
```

Issue #633, opened 2023-11-28
Refiling sections on the same file gives unexpected results


lyz-code:
### Describe the bug

When refiling a section to another in the same file that is below the result is not the expected

### Steps to reproduce

In the file `test.org` with the contents:

```
* Heading 1
** Subheading 1 
** Subheading 2
* Heading 2     
**** SubSubheading 2.1    
```
Place the cursor over `Subheading 1` and try to refile to `test.org:SubSubheading 2.1`

### Expected behavior

The section is well refined, so the result file is:

```
* Heading 1
** Subheading 2
* Heading 2     
**** SubSubheading 2.1   
***** Subheading 1 
```

Right now the output is:

```
* Heading 1
***** Subheading 1
** Subheading 2
* Heading 2     
**** SubSubheading 2.1    
```

### Emacs functionality

I guess it works as expected

### Minimal init.lua

```lua
vim.g.mapleader = ' '

vim.cmd([[set runtimepath=$VIMRUNTIME]])
vim.cmd([[set packpath=/tmp/nvim/site]])

local package_root = '/tmp/nvim/site/pack'
local install_path = package_root .. '/packer/start/packer.nvim'

local function load_plugins()
  require('packer').startup({
    {
      'wbthomason/packer.nvim',
      { 'nvim-treesitter/nvim-treesitter' },
      { 'kristijanhusak/orgmode.nvim', branch = 'master' },
    },
    config = {
      package_root = package_root,
      compile_path = install_path .. '/plugin/packer_compiled.lua',
    },
  })
end

_G.load_config = function()
  require('orgmode').setup_ts_grammar()
  require('nvim-treesitter.configs').setup({
    highlight = {
      enable = true,
      additional_vim_regex_highlighting = { 'org' },
    },
  })

  vim.cmd([[packadd nvim-treesitter]])
  vim.cmd([[runtime plugin/nvim-treesitter.lua]])
  vim.cmd([[TSUpdateSync org]])

  -- Close packer after install
  if vim.bo.filetype == 'packer' then
    vim.api.nvim_win_close(0, true)
  end

  require('orgmode').setup({
    org_agenda_files = {
      './*'
    }
  }
  )

  -- Reload current file if it's org file to reload tree-sitter
  if vim.bo.filetype == 'org' then
    vim.cmd([[edit!]])
  end
end

if vim.fn.isdirectory(install_path) == 0 then
[...]

PR #637, opened 2023-12-27
fix: repair refile to headline in the same file (#633)

seflue:
- use correct line position from target headline
- reset cursor back
- fix issue #633

### DIFF
--- a/lua/orgmode/capture/init.lua
+++ b/lua/orgmode/capture/init.lua
@@ -350,9 +350,9 @@ function Capture:_refile_to(opts)
   local item = opts.item
   if is_same_file and item then
     local target = opts.range.end_line
-    local pos = vim.fn.getcurpos()
+    local view = vim.fn.winsaveview()
     vim.cmd(string.format('silent! %d,%d move %s', item.range.start_line, item.range.end_line, target))
-    vim.fn.cursor(pos[2], pos[3])
+    vim.fn.winrestview(view)
     return true
   end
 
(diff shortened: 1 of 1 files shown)


seflue:
@kristijanhusak, @jgollenz: I need your aproval for this tiny change.

lua/orgmode/capture/init.lua
```diff
@@ -349,7 +349,10 @@ function Capture:_refile_to(opts)
 
   local item = opts.item
   if is_same_file and item then
-    vim.cmd(string.format('silent! %d,%d move %s', item.range.start_line, item.range.end_line, tostring(opts.file)))
+    local target = opts.range.end_line
+    local pos = vim.fn.getcurpos()
```
kristijanhusak, line 353: You can use `winsaveview()` and `winrestview()` to keep the cursor position.

kristijanhusak (changes requested):

kristijanhusak (approved):

kristijanhusak:
@seflue code looks good, but tests for nightly started to fail. Can you take a look?

seflue:
I'm already having a look into it and I don't get it. The failing test is regarding a capturing (failed expectation) which should not trigger the branch.

seflue:
I reverted the change to see, if the test failure was caused by it - unsuprisingly it wasn't. So the nightly version introduced a regression with the plugin, which is independent from this PR and needs to be investigated independently.

kristijanhusak:
Ok, thanks for checking. I'm merging this in and we can investigate the issue separately.

PR #640, opened 2023-12-28
fix: cache indent matches to avoid occasional race conditions

### LINKED ISSUE #639: Tests fail for full reindents with indent mode on non-nightly (`0.10.0`) Neovim
### Describe the bug

On non-nightly versions of Neovim (<`0.10.0`) the tests for full reindent can intermittently fail.

These tests do not fail all the time, only occasionally -- this makes this a right headache to resolve. This problem was introduced by #629.

### Steps to reproduce

Use a Neovim version in the `0.9.x` range, ideally `0.9.0` as that exhibits the problem most consistently and then run (assuming a bash shell):

```bash
git clone https://github.com/nvim-orgmode/orgmode.git Orgmode \
 && cd Orgmode

TEST_RUN=1
while :; do
  printf "==== Current Test Run: %s ====\n" "${TEST_RUN}"
  if make test FILE=./tests/plenary/org/indent_spec.lua; then
    printf "\n\n++++ Finished Test Run: %s ++++\n\n" "${TEST_RUN}"
  else
    printf "\n\n==== Tests Failed on Run %s ====\n\n" "${TEST_RUN}"
    break
  fi
  ((TEST_RUN++))
done
```

Now wait for a failure (again use a Neovim version in range `0.9.x`). It may take a while to happen.

### Expected behavior

Tests consistently pass instead of only sometimes passing.

### Emacs functionality

_No response_

### Minimal init.lua

N/A. This is a testing issue.

orgmode-testing-bug

### OS / Distro

Arch Linux -- `Linux orion 6.6.8-arch1-1 #1 SMP PREEMPT_DYNAMIC Thu, 21 Dec 2023 19:01:01 +0000 x86_64 GNU/Linux`

### Neovim version/commit

`0.9.1`, `0.9.2`, `0.9.4`

### Additional context

See [here](https://github.com/nvim-orgmode/orgmode/pull/627#issuecomment-1870565947) for where this problem was discovered.

This does not seem to affect nightly (`0.10.0`) versions of Neovim -- I can't cause any test failures on nightly at the time of writing.

This seemingly only impacts the tests, not any actual usage. I cannot replicate by opening a orgmode file in Neovim `0.9.0` several times. Seems fine from the user perspective. Purely a testing issue.

@kristijanhusak Opening this so its tracked, I've spent the last hour or so battling this and haven't made any progress. I'm going to finish up the Virtual indent stuff and try to solve this again when I'm not losing my sanity.

treatybreaker:
This is a really weird bug. I know it's an issue with how the `get_matches` function works and buftick. `indentexpr` isn't really intended for indenting many lines that all depend on each other's indents.

What this does is "freeze" the matches during a indent operation against a range of lines and only updates the previous matches' indent as necessary. By doing this we get "memory" for the indentexpr of the previous indents as part of a ranged indent operation like `norm! 0gg=G` which avoids any race conditions as to the speed of indents. Without this caching the `get_matches` function is called on every indentexpr operation, so for every line in the range, which can result in bad data back for calculating the indents.

I have added a big ol' comment above the indentexpr with further information.

Closes #639

### DIFF
--- a/lua/orgmode/org/indent.lua
+++ b/lua/orgmode/org/indent.lua
@@ -251,13 +251,42 @@ local function foldexpr()
   return '='
 end
 
+-- Some explanation as to the caching insanity inside of this function. The `get_matches` function
+-- is memoized, but that only goes so far. When a user wants to indent a large region, say with
+-- `norm! 0gg=G` every indent operation will call `get_matches` and get *new* matches. For the most
+-- part, this is fine, but on occasion the cache can end up invalidated when the indent operation doesn't
+-- occur fast enough. When the cache is invalidated new matches are returned and this leads to an
+-- issue in which the indent calculated for the line is no longer correct as it is based on bad
+-- data. This causes indents, especially for lists, to be incorrect as many indents are dependent on
+-- the previous node's indentation.
+--
+-- By caching the matches and previous line numbers matched we can effectively check if a range was
+-- requested for indentation and, if so, stop requesting new matches; then we only use the initial
+-- matches while updating the previous indent amounts as we return the new indents. We invalidate
+-- the cached matches when the user isn't in normal mode as it's likely they're modifying buffer
+-- content which requires us to get the updated matches for the changed content.
+--
+-- TLDR: The caching avoids some inconsistent race conditions with getting the Treesitter matches.
+local buf_indentexpr_cache = {}
 local function indentexpr(linenr, mode)
   linenr = linenr or vim.v.lnum
   mode = mode or vim.fn.mode()
   query = query or vim.treesitter.query.get('org', 'org_indent')
-  local matches = get_matches(0)
 
-  return get_indent_for_match(matches, linenr, mode)
+  local bufnr = vim.api.nvim_get_current_buf()
+  local indentexpr_cache = buf_indentexpr_cache[bufnr] or { prev_linenr = -1 }
+  if indentexpr_cache.prev_linenr ~= linenr - 1 or not mode:lower():find('n') then
+    indentexpr_cache.matches = get_matches(0)
+  end
+
+  local new_indent = get_indent_for_match(indentexpr_cache.matches, linenr, mode)
+  local match = indentexpr_cache.matches[linenr]
+  if match then
+    match.indent = new_indent
+  end
+  indentexpr_cache.prev_linenr = linenr
+  buf_indentexpr_cache[bufnr] = indentexpr_cache
+  return new_indent
 end
 
 local function foldtext()


treatybreaker:
Proof this works for me:

orgmode-indent-race-condition-fixed

I only let this run out to near 30 times on Neovim `0.9.0`. Max upload size is 10mb for files on github 🤷. I had another that went out to around ~500 runs before I figured that was good enough:
image


The script I ran to validate:
```bash
TEST_RUN=1; while :; do
  printf "==== Current Test Run: %s ====\n" "${TEST_RUN}"
  if make test FILE=./tests/plenary/org/indent_spec.lua; then
    printf "\n\n++++ Finished Test Run: %s ++++\n\n" "${TEST_RUN}"
  else
    printf "\n\n==== Tests Failed on Run %s ====\n\n" "${TEST_RUN}"
    break
  fi
  ((TEST_RUN++))
done
```

I recommend running the above on my branch and validating it works for you as well, the script will stop if any failures occur. AFAIK my patch resolves the issue.

kristijanhusak (changes requested):
Could we slightly simplify this by leveraging `options` on the `memoize_by_buf_tick` (https://github.com/nvim-treesitter/nvim-treesitter/blob/master/doc/nvim-treesitter.txt#L502) ? You can provide a custom `key` where the memoization knows to use either cached result or generate a new one. Same thing is done in markup highlighter here https://github.com/nvim-orgmode/orgmode/blob/92bfc3fb7ee845d9e58326b0b69f3ed89e84253f/lua/orgmode/colors/markup_highlighter.lua?plain=1#L387

treatybreaker:
So I'm not certain that `key` function actually works for this use case. I'd love to be proven wrong, but as far as I can tell the key function doesn't lend itself well to this problem.

It seems the key function has a hard requirement on using the `vim.b.changedtick` to determine if the cache should be invalidated, seen [here](https://github.com/nvim-treesitter/nvim-treesitter/blob/7a2c6211d6f550988fccd8800de8426d5e480a2d/lua/nvim-treesitter/ts_utils.lua?plain=1#L364). Unfortunately it's possible the buffer's tick gets changed during indent operations (which I suspect is what was causing the cache invalidation). There doesn't seem to be a way to avoid this using the `memoize_by_buf_tick` function. We really only want to cache entirely by previous line numbers and the vim mode.

If you know of a way to do this with the `key` function, I'm happy to implement it. But I don't see a decent way of doing it that isn't just shifting the current logic around the key function and doing our level best to ignore the changedtick - kinda hacky.

kristijanhusak (approved):
Yeah you are right. Lets go with this for now. Thanks!

PR #641, opened 2023-12-31
fix: promote headline while refile to file-end

seflue:
- promote headlines to first level when archiving (like Emacs)
- promote headlines to first level when refiling without targeting a
  headline (Emacs only allows headlines as refiling targets)

Motivation:

Refiling to file-ends is only indirectly used in Emacs Org-Mode for
archiving. In this case Org-Mode promotes headlines to the first
level in the archive file.

This fix aligns the archiving-behavior with Emacs while also adapting
the normal refiling feature when no headline was given.

This has two advantages: It prevents the user from getting wrong and
unexpected structures in org-files and archive-files and it aligns
Nvim-Orgmode more with the original Emacs Org-Mode.

Because the main logic is now consolidated in Capture:_refile_to,
the two functions Capture:refile_to_headline and Capture:_refile_to_end
are not needed anymore.

Adjust the tests accordingly.

Note:
A general refactoring of the Capture class would be helpful in the
future. The current Capture class mixes several concerns: Refiling,
capturing and archiving. Separating these into different classes might
help to make the code easier to understand and to adjust.


lua/orgmode/capture/init.lua
```diff
@@ -240,49 +223,27 @@ function Capture:_refile_content_with_fallback(opts)
...
+---@param is_same_file boolean
+function Capture:_adapt_headline_level(item, target_level, is_same_file)
+  -- Refiling in same file just moves the lines from one position
+  -- to another,so we need to apply demote instantly
+  if target_level == 0 then
+    return item:promote(item.level - 1, true, not is_same_file)
+  elseif item.level <= target_level then
+    return item:demote(target_level - item.level + 1, true, not is_same_file)
+  else
+    return item:promote(item.level - target_level - 1, true, not is_same_file)
```
kristijanhusak, line 245: Since we do early returns, I'd prefer having each if statement separately instead of elseif.

```suggestion
  if target_level == 0 then
    return item:promote(item.level - 1, true, not is_same_file)
  end
  if item.level <= target_level then
    return item:demote(target_level - item.level + 1, true, not is_same_file)
  end
  return item:promote(item.level - target_level - 1, true, not is_same_file)
```

lua/orgmode/capture/init.lua
```diff
@@ -343,16 +304,43 @@ function Capture:_refile_to(opts)
     return false
   end
 
-  apply_properties(opts)
+  local has_headline = opts.headline and opts.headline ~= ''
+  local destination_file = Files.get(opts.file)
+  local target_level
+  local target_line
```
kristijanhusak, line 310: Define these as defaults, and then we can remove the `else` block.
```suggestion
  local target_level = 0
  local target_line = -1
```

kristijanhusak (changes requested):
Looks good, just some minor code changes.

lua/orgmode/capture/init.lua
seflue: Done.

lua/orgmode/capture/init.lua
seflue: Done.

kristijanhusak (approved):

kristijanhusak:
Thanks!

PR #642, opened 2023-12-31
fix: schedule vim commands in state_spec test

seflue:
Fix test on CLI AND Neotest by executing the vim commands asynchronously.

This test was not marked as failed when it ran from the command line, although the error was visible in the output.
In Neotest it was actually marked as failed.

### DIFF
--- a/tests/plenary/state/state_spec.lua
+++ b/tests/plenary/state/state_spec.lua
@@ -79,10 +79,9 @@ describe('State', function()
   it('should be able to self-heal from an invalid state file', function()
     state:save_sync()
 
-    -- Mangle the cache
-    vim.cmd.edit(cache_path)
+    vim.cmd(('edit %s'):format(cache_path))
     vim.api.nvim_buf_set_lines(0, 0, -1, false, { '[ invalid json!' })
-    vim.cmd.write()
+    vim.cmd('write')
 
     -- Ensure we reload the state from its cache file (this should also "heal" the cache)
     state._ctx.loaded = false


seflue:
Within neotest I get the following error:
image
On CLI I actually also had 0 errors, regarding the output I might have seen a gost (a couple of tests throw some output and it's quite possible, that I misread it).
To check against neotest, you would need [my last state of neotest_plenary](https://github.com/seflue/neotest-plenary) with some funky reconfiguration:
image
Or you rename the `scripts/minimal_init.lua` into something which isn't catched by the neovim-plenary globbing.
Background: neovim-plenary is globbing for a `minimal_init.lua`, and `tests/minimal_init.lua` in the orgmode project is the correct one. But because `scripts/minimal_init.lua` matches first, so the tests don't run at all with neotest. My fix makes the globbing configurable and also fixes the configurability of neotest-plenary in a general sense. Being able to run the tests from within neovim is much more convenient during development.

tests/plenary/state/state_spec.lua
```diff
@@ -80,9 +80,11 @@ describe('State', function()
...
 
     -- Mangle the cache
-    vim.cmd.edit(cache_path)
-    vim.api.nvim_buf_set_lines(0, 0, -1, false, { '[ invalid json!' })
-    vim.cmd.write()
+    vim.schedule_wrap(function()
+      vim.cmd.edit(cache_path)
+      vim.api.nvim_buf_set_lines(0, 0, -1, false, { '[ invalid json!' })
+      vim.cmd.write()
+    end)
```
kristijanhusak, line 87: Can you try these changes instead and see if it works?

```suggestion

    vim.cmd(('edit %s'):format(cache_path))
    vim.api.nvim_buf_set_lines(0, 0, -1, false, { '[ invalid json!' })
    vim.cmd('write')

```

From the screenshot I see that neotest-plenary cannot figure out that `vim.cmd` is a table, so this might fix it.

seflue: Thanks, this actually also fixes the problem. I updated my commit with your suggestion.

kristijanhusak (approved):

PR #647, opened 2024-01-07
Fix refiling to current buffer (#644)

### LINKED ISSUE #644: Easy to lose in-progress capture that only ends up in temp dir with `:w`
### Describe the bug

If you have an in-progress capture, it can be very easy to lose it entirely. It seems to be recoverable as long as you don't close nvim, and know exactly what to do, but it doesn't seem intuitive to do so.

It appears that until you actually file the capture, it doesn't _really_ live anywhere persistent (expected), even if you write the buffer (not expected).

Before you file or write with `:w`, if you use `<TAB>` from the agenda to open a different TODO, you'll get this error: `[orgmode] vim/_editor.lua:341: nvim_exec2(): Vim(edit):E37: No write since last change (add ! to override)`. Similar story if you try to close the buffer or nvim as a whole. However, if you use `:w`, then that in-progress capture's buffer is allowed to close without being filed and without warning.

The buffer also seems to disappear entirely (`ls!` does not show it) when closed, which exacerbates the issue. I am not sure if this is just because it's unlisted or some other reason, but I was not able to identify a way to get this buffer back outside of manually identifying and opening the temp file. At that point, it's a plain file with no orgmode support (maybe a way to re-enable that?) but I think you'd have to manually file it.

If you don't do that, you will permanently lose that capture when nvim is closed as that temp file seems to get cleaned up.

The end result here is that I lost some notes for a meeting that, thankfully, I was just taking to get used to orgmode.

This is just one example workflow, but any way that you end up leaving nvim after saving an in-progress capture using `:w` will run into this, as the `:w` prevents any warnings.

### Steps to reproduce

Edit: Keeping the below for posterity, but they don't actually reproduce the issue.
See https://github.com/nvim-orgmode/orgmode/issues/644#issuecomment-1879921992 for steps that actually reproduce.

~~The simplest workflow, I think:~~

~~1. Start a new capture with `<leader>oc`~~
~~2. Write the capture buffer with `:w`, noting the temp file name it is written to~~
~~3. Close nvim with `:q`~~
~~4. See that the temp file is gone, and the capture was never filed~~

### Expected behavior

I think there are a couple of improvements that could potentially be made here:

- Warn about any un-filed captures before overwriting the capture buffer (or allowing vim to close, if possible), regardless of whether they've been written to the temp file.
- Make the capture buffers listed. This makes them way more discoverable and provides general benefits outside of just this problem.
- Provide a command/keymap to get back to an in-progress capture with full orgmode support

### Emacs functionality

_No response_

### Minimal init.lua

```lua

local nvim_root = '/tmp/nvim_orgmode'
local lazy_root = nvim_root .. '/lazy'
local lazypath = lazy_root .. '/lazy.nvim'

-- Install lazy.nvim if not already installed
if not vim.loop.fs_stat(lazypath) then
  vim.fn.system({
    'git',
    'clone',
    '--filter=blob:none',
    'https://github.com/folke/lazy.nvim.git',
    '--branch=stable', -- latest stable release
    lazypath,
  })
end
vim.opt.rtp:prepend(lazypath)

require('lazy').setup({
  {
    'nvim-orgmode/orgmode',
    dependencies = {
      { 'nvim-treesitter/nvim-treesitter', lazy = true },
    },
    event = 'VeryLazy',
    config = function()
      -- Load treesitter grammar for org

      -- Setup treesitter
        highlight = {
          enable = true,
        },
        ensure_installed = { 'org' },
      })

      -- Setup orgmode
     require('orgmode').setup({
        org_agenda_files = nvim_root .. '/org/**/*',
        org_default_notes_file = nvim_root .. '/org/refile.org',
     })
    end,
  },
}, {
  root = lazy_root,
  lockfile = nvim_root .. '/lazy.json',
  install = {
    missing = false,
  },
})

require('lazy').sync({
  wait = true,
  show = false,
})

```

_No response_

### OS / Distro

macOS 14.2

### Neovim version/commit

v0.9.4

### Additional context

Also reproduced in v0.9.2 on Windows 10

seflue:
This fixes #644. @kristijanhusak Could you please merge #641 first, because this PR builds upon it. I will then rebase this one. If you decide to merge this directly, we can close #641.

### DIFF
--- a/lua/orgmode/capture/init.lua
+++ b/lua/orgmode/capture/init.lua
@@ -296,6 +296,15 @@ local function remove_buffer_empty_lines(opts)
   range.end_line = end_line
 end
 
+--- Checks, if we refile a heading within one file or from one file to another.
+---@param opts CaptureOpts
+---@return boolean
+local function check_refile_source(opts)
+  local source_file = opts.item and opts.item.file or utils.current_file_path()
+  local target_file = opts.file
+  return source_file == target_file
+end
+
 ---@private
 ---@param opts CaptureOpts
 ---@return boolean
@@ -319,8 +328,8 @@ function Capture:_refile_to(opts)
     target_line = headline.range.end_line
   end
 
-  local is_same_file = opts.file == utils.current_file_path()
   local item = opts.item
+  local is_same_file = check_refile_source(opts)
   if item and should_adapt_headline then
     -- Refiling in same file just moves the lines from one position
     -- to another,so we need to apply demote instantly


kristijanhusak (changes requested):
@seflue just do a rebase here so we can proceed. Thanks!

seflue:
> @seflue just do a rebase here so we can proceed. Thanks!

I already did, see the force push.

kristijanhusak:
@seflue rebase from `master` branch. This PR still contains changes from https://github.com/nvim-orgmode/orgmode/pull/641, and there is conflict in `capture/init.lua` file.

seflue:
Sorry, I forgot, that you do not merge commits but squashing PRs onto master. I just did the requested rebase.

kristijanhusak (approved):
Thanks!

PR #649, opened 2024-01-11
Remove duplicate list item

adambubenicek:
"Clocking time" has been appearing twice in the Features list


kristijanhusak (approved):
Thanks!

Issue #607, opened 2023-09-04
Extend date picker (Calendar) with time


seflue:
### Does this feature exist in Emacs orgmode core?

Yes

### Orgmode link

I'm not sure, if Emacs really has a time picker for scheduling dates. But it is [mentioned](https://orgmode.org/manual/Deadlines-and-Scheduling.html), that times are allowed to be used for scheduling and deadline entries.

### Feature value

I missed a convenient picker for times, so I sat down and [implemented it](https://github.com/seflue/nvim-orgmode/commit/6b83392eefcebb569216515cf2d8c755b3da35a5).

![image](https://github.com/nvim-orgmode/orgmode/assets/952313/c4e40260-8eca-43da-9a24-56635f6da920)

I will add an PR as soon as #588 is merged.

azycat:
are you still working on this? i was just thinking today about how convenient this feature would be... thanks

seflue:
I have it on my daily used branch, but not driven it forward. There are still some usability quirks I need to fix. But if you are curious, feel free to try it out on my fork, I rebase it onto upstream master regularly. And I would be happy for some feedback.

seflue:
@azycat I filed a PR, let's see if @kristijanhusak finds it also useful. 🙏

seflue:
Implemented, closed.

Issue #396, opened 2022-08-17
org_timestamp_{up,down} with count behaves unexpectedly on minutes


Ram-Z:
### Describe the bug

When using `org_timestamp_{up,down}` with a count on minutes, the value is changed by `count` times `org_time_stamp_rounding_minutes` rather than `count` minutes.

*Also there's inconsistent spelling of `timestamp` and `time_stamp` in the options.*

### Steps to reproduce

1. `set org_time_stamp_rounding_minutes = 5`
2. put cursor on minutes part of datetime
3. do `15<C-A>`

### Expected behavior

`15<C-A>` increases 15 minutes regardless of the value of `org_time_stamp_rounding_minutes`.

### Minimal init.lua

Reproducible on default settings.

### OS / Distro

Arch

### Neovim version/commit

NVIM v0.7.2

jgollenz:
I am not sure if I understood your issue. The behavior you describe is the one I would expect. You are running `<C-a>` fifteen times, each one increasing the time by 5 minutes. 

The name `org_time_stamp_rounding_minutes` comes from Emacs, I think `org_timestamp_up` is exclusive to this port. So it should probably be `org_time_stamp_[up|down]`.

Ram-Z:
Yeah, I understand the current behaviour is explained by `#<C-A>` simply applying `<C-A>` `#` times. I'd argue that that is unexpected though. When you give a count, you probably know exactly how many minutes you want to increment/decrement by, you should not be needing to do mental acrobatics to figure out how many multiples of `org_time_stamp_rounding_minutes` you want.

Your comment prompted me to look up what emacs actually does (I've not used it before, so usually don't compare). And emacs behaves very differently from the current nvim-orgmode. 

From [org-time-stamp-rounding-minutes](http://doc.endlessparentheses.com/Var/org-time-stamp-rounding-minutes.html):
> Number of minutes to round time stamps to.

This applies to the timestamp, not the step size, f.ex `[2022-08-18 Thu 08:2|2]` `<C-A>` -> `[2022-08-18 Thu 08:2|5] `<C-A>` -> `[2022-08-18 Thu 08:3|0] `, i.e. the first `<C-A>` rounds it to the nearest `org_time_stamp_rounding_minutes` then starts incrementing by `org_time_stamp_rounding_minutes`.

Furthermore:
> When this is larger than 1, you can still force an exact time stamp [...] by using a prefix arg to `S-up/down' to specify the exact number of minutes to shift.

[...]

PR #650, opened 2024-01-14
feature: enhance calendar widget with time picker

seflue:
When creating a date using the calendar widget, a time can now set directly from the widget.

- t jumps to the time picker
- h/l switches between hour, "big" minute and "small" minute selection
- j/k decreases/increases the time

The step size for "big" and "small" minutes can be configured. Default is 15 and 5 minutes.
When adjusting hours, the minutes are rounded or not (this behavior can be disabled in the config, if too confusing - I find it quite useful).

See also #607.


lua/orgmode/config/defaults.lua
```diff
@@ -35,6 +35,8 @@ local DefaultConfig = {
   org_highlight_latex_and_related = nil,
   org_custom_exports = {},
   org_indent_mode = 'indent',
+  org_time_picker_min_big = 15,
```
kristijanhusak, line 38: Let's remove these options and rely only on `org_time_stamp_rounding_minutes`.  Holding j/k updates the time fast enough, so we can stick with only one step.

Regarding the rounding, I would suggest going with additional mappings instead of having it as an option. There is a similar feature to be implemented for regular time up/down (https://github.com/nvim-orgmode/orgmode/issues/396) and Emacs uses a mapping with different prefix.

We can either go with uppercase `J/K` or maybe `<leader>j/k`.

lua/orgmode/objects/calendar.lua
```diff
@@ -149,69 +161,177 @@ function Calendar.render()
   else
     table.insert(content, ' [i] - enter date')
   end
+  if Calendar.has_time() then
```
kristijanhusak, line 164: Would be great to have a mapping that exists "time" mode and goes back to selecting a date.

lua/orgmode/objects/date.lua
kristijanhusak, line 1: Why were these changes done? We should be able to use `clone` for situations where we need to override something specific.

kristijanhusak (commented):
Thanks for the PR! This looks good, but I would like to slightly change things around. My first comment sums up the biggest changes.

We could also refactor the `render` function and split it into chunks that can be reused, like `render_days`, `render_time`, `render_help`. That way we can easily rerender only a section of it if necessary.

lua/orgmode/config/defaults.lua
seflue: I actually really would like to keep it, because I like it in the daily usage. My argument is precision. I often schedule meetings to a quarter of an hour, holding  the 5 min key is not the same. If different preferences exist, I rather would make it optional via configuration than passing on it completely.

lua/orgmode/objects/calendar.lua
seflue: I implemented it in the way, that you can exit the time with Esc and with another Esc exit the calendar completely and this feels quite naturally for me. Adding another mapping explicit for the time sounds cumbersome to me. Maybe Tab to switch between calendar and time but actually I currently don't see so much value to change what already works quite good.

lua/orgmode/objects/date.lua
seflue: It's actually quite some time ago, when I implemented this, but I remember, that date behaved inconsistent regarding time and these changes fix the problems I had.

lua/orgmode/config/defaults.lua
kristijanhusak: Ok, lets just create a separate `calendar` config option and nest all these settings inside. I want to separate non Orgmode configuration.

lua/orgmode/objects/date.lua
kristijanhusak: See if you can figure out what was the issue and we can investigate. I'd rather not change this and solve the issue differently.

lua/orgmode/objects/calendar.lua
```diff
@@ -184,14 +201,34 @@ function Calendar:render()
...
+  vim.api.nvim_buf_add_highlight(self.buf, namespace, 'Comment', #content - 3, 0, -1)
   vim.api.nvim_buf_add_highlight(self.buf, namespace, 'Comment', #content - 2, 0, -1)
   vim.api.nvim_buf_add_highlight(self.buf, namespace, 'Comment', #content - 1, 0, -1)
 
+  -- highlight the cell of the current day
+  self:highlight_day(content, Date.today(), 'OrgCalendarToday')
+  -- highlight selected day
+  self:highlight_day(content, self.date, 'OrgCalendarSelected')
+
+  -- TODO this new loop (after Kristjans refactoring) is currently not fully understood.
```
seflue, line 231: @kristijanhusak I am struggeling a bit to fully understand, what your new version of the rendering adjustment does differently than before and how it behaves differently to my `highlight_day` method. I would appreciate your help to consolidate your loop and `on_render_day` method with my `highlight_day` method.

lua/orgmode/objects/calendar.lua
```diff
@@ -184,14 +201,34 @@ function Calendar:render()
     table.insert(content, ' [i] - enter date')
   end
 
+  -- FIXME this line is currently not shown because of a bug.
```
seflue, line 204: There is still a bug, that the following lines are not rendered. Currently don't understand why.

lua/orgmode/objects/calendar.lua
kristijanhusak: @seflue it's somewhat similar. You can just put your logic inside the `on_render_day` in a same way the today's date is highlighted. For each day `on_render_day` is triggered with a date and some positioning info. That should allow you to highlight any of the days however you want.

lua/orgmode/objects/calendar.lua
seflue: This is now fixed.

lua/orgmode/objects/calendar.lua
seflue: Understood. I integrated my code with yours.

lua/orgmode/objects/calendar.lua
seflue: Because I rebased my code onto your refactoring, the window now also closes with `q`. I mapped the switch back to day selection now to `d`.

lua/orgmode/objects/date.lua
seflue: Rebased my branch onto the adjusted date fix.

lua/orgmode/config/defaults.lua
seflue: I wrapped all calendar related settings into a `calendar` table as you suggested.

lua/orgmode/config/defaults.lua
```diff
@@ -12,7 +13,12 @@ local DefaultConfig = {
   org_agenda_span = 'week', -- day/week/month/year/number of days
   org_agenda_start_on_weekday = 1,
   org_agenda_start_day = nil, -- start from today + this modifier
-  calendar_week_start_day = 1,
+  calendar = {
+    round_min_with_hours = true,
+    min_big_step = 15,
+    min_small_step = 5,
+    start_from_sunday = false,
```
seflue, line 20: I changed this to a boolean and gave it a clearer name.

seflue:
Somehow, I got a regression with the date mapping tests, need to investigate that later.

lua/orgmode/objects/calendar.lua
```diff
@@ -150,10 +150,9 @@ function Calendar:render()
   vim.api.nvim_set_option_value('modifiable', true, { buf = self.buf })
 
   local cal_rows = { {}, {}, {}, {}, {}, {} } -- the calendar rows
-  local start_from_sunday = config.calendar_week_start_day == 0
```
seflue, line 139: @kristijanhusak It seems, that this changes to the config introduced a regression to how the date mapping tests work. But I currently don't see why ...

DOCS.md
```diff
@@ -429,14 +430,6 @@ If `true`, generate ID with the Org ID module and append it to the headline as p
 *default value*: `{ [':tangle'] = 'no', [':noweb']  = no }`<br />
 Default header args for extracting source code. See [Extract source code (tangle)](#extract-source-code-tangle) for more details.
 
-#### **calendar_week_start_day**
```
kristijanhusak, line 432: This option should not be removed. It exists in the Emacs by the same name.

DOCS.md
```diff
@@ -677,6 +670,37 @@ Additional files to search from agenda search prompt.<br />
...
+_default_: `15`<br />
+The step size for changing the minutes while the cursor is on the first digit.
+
+#### **calendar.min_small_step**
+
+_type_: `number`<br />
+_default_: `5`<br />
+The step size for changing the minutes while the cursor is on the second digit.
+
+#### **calendar.start_from_sunday**
```
kristijanhusak, line 695: As mentioned in above comment, remove this and use the top level option.

kristijanhusak (changes requested):
@seflue lets keep the calendar week start variable and adapt the rest according to it. That might fix the issue with the tests.

kristijanhusak (changes requested):
> Are we ready to merge, finally?

We're almost there, but a few more things needs to be addressed, sorry. I'll write them down, and put a video below showing all of them in the same order

1. Add an underline to a currently selected day only when switching to a time picker. Currently, it highlights on every movement, but it's not consistent if user is not using bound movement keys (I'm using `w` and `b` in the video)
2.  Triggering any other mappings while in the time picker mode moves the cursor but still manipulates the date. We should probably go out of the time picker mode once any of those are pressed
3. Shortcut shows `[t] Enter time` option when we enter the time mode until we change the time, then it switches.

Video: 

https://github.com/nvim-orgmode/orgmode/assets/1782860/3e5be8eb-f669-4d37-8bd0-485e6c47fe8a



I'm aware you spent a lot of time on this implementation, but what do you think about switching it up a bit to be more time centric when in time mode? Let me give an example:

Currently with this PR, we show date and time like this:

```
             June 2024
 Mon  Tue  Wed  Thu  Fri  Sat  Sun
                          01   02 
 03   04   05   06   07   08   09 
 10   11   12   13   14   15   16 
 17   18   19   20   21   22   23 
 24   25   26   27   28   29   30 
  
               16:30              

 [<] - prev month  [>] - next month
 [.] - today   [Enter] - select day
 [i] - enter date
 [t] - enter time
```

What do you think about switching the whole UI when you go to a time picker mode? Mappings would work the similar, but you wouldn't have to worry about rendering the calendar view and all of the other things. Basically, we could switch the calendar to have 2 views, and code them separately inside the calendar.

Time view would be something like this (wrote it manually just to get an idea):

```
    Date: 2024-05-28

           16:30

 [Enter] - select time
 [T] - clear time
 [Q] - Go back
```

This view should not be able to quit the calendar, so it would be more of a sub-view.
It's not flexible as the current solution, but it should be more maintainable and (somewhat) easier to code.

lua/orgmode/objects/calendar.lua
```diff
@@ -381,4 +574,49 @@ function Calendar:read_date()
...
+  self.select_state = SelState.HOUR
+  vim.fn.cursor({ 9, 16 })
+end
+
+function Calendar:set_sel_day()
+  self.select_state = SelState.DAY
+  self:jump_day()
+end
+
+function Calendar:set_sel_min10()
```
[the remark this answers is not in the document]
seflue: Done.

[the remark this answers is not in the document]
seflue: Done.

seflue:
One last argument. While using nvim orgmode day to day I am discovering a lot of much rougher edges I would actually like to spend time on to fix them. But this long-running branch is already dragging a bit on my motivation, because it feels like a never-ending story.

kristijanhusak (approved):
Ok, sorry for bothering you so much. I'll address some of the smaller things. 
Thanks!

PR #651, opened 2024-01-14
feature: support org_toggle_timestamp_type

seflue:
Allows to make active timestamp inactive and vice versa with one keybinding.

Implements feature request #652.

It is part of Emacs Orgmode, but not mapped.

See also:
- https://doc.endlessparentheses.com/Fun/org-toggle-timestamp-type.html
- https://emacs.stackexchange.com/questions/38062/configure-key-to-toggle-between-active-and-inactive-timestamps/38064#38064

### DIFF
--- a/DOCS.md
+++ b/DOCS.md
@@ -1023,6 +1023,9 @@ Decrease date under cursor by 1 or "count" day(s) (Example count: `5<S-UP>`).
 #### **org_change_date**
 *mapped to*: `cid`<br />
 Change date under cursor. Opens calendar to select new date
+#### **org_toggle_timestamp_type**
+_mapped to_: `<prefix>d!`<br />
+Switches the timestamp under the cursor between inactive and active.
 #### **org_priority**
 *mapped to*: `<Leader>o,`<br />
 Choose the priority of a headline item.
--- a/DOCS.md
+++ b/DOCS.md
@@ -1024,7 +1024,7 @@ Decrease date under cursor by 1 or "count" day(s) (Example count: `5<S-UP>`).
 *mapped to*: `cid`<br />
 Change date under cursor. Opens calendar to select new date
 #### **org_toggle_timestamp_type**
-_mapped to_: `<prefix>d!`<br />
+*mapped to*: `<Leader>od!`<br />
 Switches the timestamp under the cursor between inactive and active.
 #### **org_priority**
 *mapped to*: `<Leader>o,`<br />
--- a/lua/orgmode/config/mappings/init.lua
+++ b/lua/orgmode/config/mappings/init.lua
@@ -353,7 +353,7 @@ return {
     ),
     org_toggle_timestamp_type = m.action(
       'org_mappings.org_toggle_timestamp_type',
-      { opts = { desc = 'org toggle timestamp type' } }
+      { opts = { desc = 'org toggle timestamp type', help_desc = 'Toggle timestamp active/inactive type' } }
     ),
   },
   edit_src = {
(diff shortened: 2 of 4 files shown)


kristijanhusak:
Similar thing can be achieved if you put a cursor on edge `<`/`>` (or `[`/`]` on inactive) and do `ctrl + a` (or `ctrl + x`)

```
<2024-01-15 Mon 13:19>
^ cursor here, do ctrl + a
```

would that work or you think this is still more helpful?

seflue:
> Similar thing can be achieved if you put a cursor on edge `<`/`>` (or `[`/`]` on inactive) and do `ctrl + a` (or `ctrl + x`)
> 
> ```
> <2024-01-15 Mon 13:19>
> ^ cursor here, do ctrl + a
> ```
> 
> would that work or you think this is still more helpful?

Is this actually documented somewhere? And what is the relation to Emacs Orgmode? Because my PR relates to a feature implemented in Emacs Orgmode (and because it doesn't need any precise positioning of the cursor) I actually find it more helpful. It has a clear mapping, a function and is googlable. If you agree to add it, I would also add some documentation.

kristijanhusak:
Yeah, in this section https://orgmode.org/manual/Creating-Timestamps.html#index-S_002dUP-4

>On the beginning or enclosing bracket of a timestamp, change its type

But since this is a built-in option of Orgmode, I'm ok with adding it.  You can add documentation.
I'm just trying to figure out if we could maybe use a different mapping, maybe something like `<prefix>d!`. What do you think?'

seflue:
> Yeah, in this section https://orgmode.org/manual/Creating-Timestamps.html#index-S_002dUP-4
> 
> >On the beginning or enclosing bracket of a timestamp, change its type
> 
> But since this is a built-in option of Orgmode, I'm ok with adding it.  You can add documentation.
> I'm just trying to figure out if we could maybe use a different mapping, maybe something like `<prefix>d!`. What do you think?'

Sure, let's change it. I'm agnostic with the mapping, and your suggestion makes sense.

seflue:
> Yeah, in this section https://orgmode.org/manual/Creating-Timestamps.html#index-S_002dUP-4
> 
> > On the beginning or enclosing bracket of a timestamp, change its type
> 
> But since this is a built-in option of Orgmode, I'm ok with adding it. You can add documentation. I'm just trying to figure out if we could maybe use a different mapping, maybe something like `<prefix>d!`. What do you think?'

Done.

kristijanhusak:
Looks good, just add some docs around it.

seflue:
> Looks good, just add some docs around it.

It took me quite some time to get back to this PR, but I just add a short documentation.

DOCS.md
```diff
@@ -1023,6 +1023,9 @@ Decrease date under cursor by 1 or "count" day(s) (Example count: `5<S-UP>`).
 #### **org_change_date**
 *mapped to*: `cid`<br />
 Change date under cursor. Opens calendar to select new date
+#### **org_toggle_timestamp_type**
+_mapped to_: `<prefix>d!`<br />
```
kristijanhusak, line 1027: Default prefix is `<Leader>o`, so lets use that in docs:

```suggestion
*mapped to*: `<Leader>od!`<br />
```

lua/orgmode/config/mappings/init.lua
```diff
@@ -351,6 +351,10 @@ return {
       'org_mappings.org_babel_tangle',
       { opts = { desc = 'org tangle', help_desc = 'Tangle current file' } }
     ),
+    org_toggle_timestamp_type = m.action(
+      'org_mappings.org_toggle_timestamp_type',
+      { opts = { desc = 'org toggle timestamp type' } }
```
kristijanhusak, line 356: This will add the mapping to the help window
```suggestion
      { opts = { desc = 'org toggle timestamp type', help_desc = 'Toggle timestamp active/inactive type' } }
```

kristijanhusak (changes requested):
Looks good, just minor tweaks to the documentation and to add it to help window.

kristijanhusak (approved):

PR #654, opened 2024-01-24
feat: Add org id support and org store link

kristijanhusak:
This PR addresses several things:
* Adds `org-store-link` 
* Adds "org id" module (https://orgmode.org/manual/Handling-Links.html#FOOT28)
* Expose api for setting headline properties and ID on the headline manually

> [!IMPORTANT]
> **Breaking change**: Mapping for `org-insert-link` is changed from `<prefix>il` (`<leader>oil`) to `<prefix>li` (`<leader>oli`).


Maltimore:
Could you comment on why the default mapping was changed away from `<prefix>il`?
I couldn't figure out what the original orgmode uses, but doesn't look to me like they use something similar to `<prefix>li`.
Obviously, the new mapping is not in line with all the other keymappings that insert something, so I'm sure you had a very good reason to change this!

kristijanhusak:
I changed it to be in line with store link mapping, so to store link you do `<prefix>ls` as "link store" and then you do `<prefix>li` as "link insert". Now when I think about it, I could've probably used `<prefix>sl` as "store link" and leave the old mapping for insert, but I'm not sure why I went the other way.

Issue #422, opened 2022-10-06
Status of org-indent-mode functionality?


IllustratedMan-code:
### Does this feature exist in Emacs orgmode core?

Yes

### Orgmode link

https://orgmode.org/manual/Org-Indent-Mode.html

IllustratedMan-code:
This was talked about before in another issue, but neovim did not support the required functionality. What needs to occur to get this to work? I'm happy to make issues in the main neovim repository to get this to happen.

jgollenz:
Are the virtual spaces just used for non-headline text or is the second space for each headline level also virtual?

```
* foo
  content            <-- two virtual spaces

  * bar              <-- one virtual space?
    more content     <-- four virtual spaces
```

IllustratedMan-code:
without org-indent-mode
```
* foo
content

** bar
more content
```
with org-indent-mode
```
* foo
  content

 ** bar
    more content
```

IllustratedMan-code:
I believe all indents are virtual.

jgollenz:
Ah, it's prepending two virtual spaces to every line, even nested headings.

This 
```
* one
foo
** two
bar
*** three
baz
```
becomes

![image](https://user-images.githubusercontent.com/12050098/194509256-d16817b9-ec2d-4606-8d83-e65049ae541b.png)

IllustratedMan-code:
`org-indent-mode` is also configurable with `org-indent-indentation-per-level` which is my case is set to 2. So the number of virtual spaces prepended should be configurable.

IllustratedMan-code:
Looks like https://github.com/neovim/neovim/pull/9496 needs to be merged before this functionality will be possible.

jgollenz:
@kristijanhusak can you please add the `neovim-dependency` label?

danilshvalov:
See #561

treatybreaker:
The following stills remains to be done:

- [ ] Toggling indent mode on or off in a buffer
- [ ] Support `#+STARTUP: indent` & `#+STARTUP: noindent`

I'll be opening PRs to resolve these as time permits. Probably no later than Sunday `2024-27-01`.

PR #658, opened 2024-01-28
feat: make VirtualIndent dynamically attach/detach based on `vim.b.org_indent_mode`

treatybreaker:
Following on from #627, this adds the capability to toggle virtual indentation within an orgmode buffer. This allows me to check one thing off my list [here](https://github.com/nvim-orgmode/orgmode/issues/422#issuecomment-1910991160).

See the short gif below:
dynamic-org-indent-mode

Bullet list of changes:
- Refactored `VirtualIndent` to allow multiple instances to easily track attachment to different buffers
- Added capability to `VirtualIndent` to enable/disable based on `vim.b.org_indent_mode` using a timer set to check the variable every 50ms
- Added tests for new functionality
- Updated docs to reflect this new functionality (`org_startup_indented` is where I placed them)

Do I need to add a user command like `OrgToggleIndentMode` or a keybind for this? Or is using the buffer variable, `vim.b.org_indent_mode` good enough to expose this functionality?

### DIFF
--- a/lua/orgmode/ui/virtual_indent.lua
+++ b/lua/orgmode/ui/virtual_indent.lua
@@ -5,6 +5,7 @@ local tree_utils = require('orgmode.utils.treesitter')
 ---@field private _attached boolean Whether or not VirtualIndent is attached for its buffer
 ---@field private _bufnrs {integer: boolean} Buffers with VirtualIndent attached
 ---@field private _timer uv_timer_t Timer used for tracking `org_indent_mode`
+---@field private _watcher_running boolean Whether or not VirtualIndent is reacting to `vim.borg_indent_mode`
 local VirtualIndent = {
   _ns_id = vim.api.nvim_create_namespace('orgmode.ui.indent'),
   _bufnrs = {},
@@ -29,6 +30,7 @@ function VirtualIndent:new(bufnr)
   new._bufnr = bufnr
   new._attached = false
   VirtualIndent._bufnrs[new._bufnr] = new
+  new._watcher_running = false
   new._timer = vim.uv.new_timer()
   return new
 end
@@ -92,27 +94,31 @@ function VirtualIndent:set_indent(start_line, end_line, ignore_ts)
   end
 end
 
---- Begins a timer to check `vim.b.org_indent_mode` and correctly attach or detatch VirtualIndent as
---- necessary
+--- Begins a timer to check `vim.b.org_indent_mode` if `vim.b.org_indent_mode` is not already being
+--- monitored
 function VirtualIndent:start_watch_org_indent()
-  self._timer:start(
-    50,
-    50,
-    vim.schedule_wrap(function()
-      local success, indent_mode_enabled = pcall(vim.api.nvim_buf_get_var, self._bufnr, 'org_indent_mode')
-      if success and indent_mode_enabled then
-        if not self._attached then
-          self:attach()
+  if not self._watcher_running then
+    self._watcher_running = true
+    self._timer:start(
+      50,
+      50,
+      vim.schedule_wrap(function()
+        local success, indent_mode_enabled = pcall(vim.api.nvim_buf_get_var, self._bufnr, 'org_indent_mode')
+        if success and indent_mode_enabled then
+          if not self._attached then
+            self:attach()
+          end
+        elseif self._attached then
+          self:detach()
         end
-      elseif self._attached then
-        self:detach()
-      end
-    end)
-  )
+      end)
+    )
+  end
 end
 
 --- Stops the current VirtualIndent instance from reacting to changes in `vim.b.org_indent_mode`
 function VirtualIndent:stop_watch_org_indent()
+  self._watcher_running = false
   self._timer:stop()
 end
 
--- a/lua/orgmode/ui/virtual_indent.lua
+++ b/lua/orgmode/ui/virtual_indent.lua
@@ -1,14 +1,15 @@
 local tree_utils = require('orgmode.utils.treesitter')
+local dict_watcher = require('orgmode.utils.dict_watcher')
 ---@class OrgVirtualIndent
 ---@field private _ns_id number extmarks namespace id
 ---@field private _bufnr integer Buffer VirtualIndent is attached to
 ---@field private _attached boolean Whether or not VirtualIndent is attached for its buffer
----@field private _bufnrs {integer: boolean} Buffers with VirtualIndent attached
----@field private _timer uv_timer_t Timer used for tracking `org_indent_mode`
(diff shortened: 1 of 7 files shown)


treatybreaker:
Found a bug with scheduling and the current code. When setting a schedule the `org_indent_mode` variable stills shows as enabled, but the virtual indents are removed.

treatybreaker:
[Solved this](https://github.com/nvim-orgmode/orgmode/pull/658#issuecomment-1913528949)... dumb mistake.

I'm going to leave this as a draft for now and drive this PR's branch when I get up later to see if anything else crops up. There might be other minor screwups, I am rather tired and it *does* in fact happen to be 3:23am which I am sure is negatively impacting my code 💀 . 

If this is looking good on my side by EOD, or sometime on Monday then I'll mark it "Ready for Review".

treatybreaker:
Thus far it's been ok for me, opening this up for review 🙂

lua/orgmode/ui/virtual_indent.lua
```diff
@@ -72,22 +92,60 @@ function VirtualIndent:set_indent(bufnr, start_line, end_line, ignore_ts)
...
----@param bufnr? number buffer id
-function VirtualIndent:attach(bufnr)
-  bufnr = bufnr or 0
-  self:set_indent(0, 0, vim.api.nvim_buf_line_count(bufnr) - 1, true)
+--- Begins a timer to check `vim.b.org_indent_mode` if `vim.b.org_indent_mode` is not already being
+--- monitored
+function VirtualIndent:start_watch_org_indent()
+  if not self._watcher_running then
+    self._watcher_running = true
+    self._timer:start(
```
kristijanhusak, line 100: Instead of timers, let's rely on Vim's `dictwatcheradd` and `dictwatcherdel` functions.
Those can be used only from vimscript, but there is a workaround.

Create a `dict_watcher.lua` file in the `utils` folder:

```lua
local watchers = {}
local M = {}

---@param change_dict { old?: any, new: any }
---@param key string
---@param dict table
function M.dict_changed(change_dict, key, dict)
  if watchers[key] then
    watchers[key](change_dict, key, dict)
  end
end

---@param key string
---@param callback fun(change_dict: { old?: any, new: any }, key: string, dict: table)
function M.watch_buffer_variable(key, callback)
  vim.cmd(([[
    call dictwatcheradd(b:, '%s', 'OrgmodeWatchDictChanges')
  ]]):format(key))
  watchers[key] = callback
end

---@param key string
function M.unwatch_buffer_variable(key)
  vim.cmd(([[
    call dictwatcherdel(b:, '%s', 'OrgmodeWatchDictChanges')
  ]]):format(key))
  watchers[key] = nil
end

return M
```

Add this to plugin/org.vim
```vim
function OrgmodeWatchDictChanges(dict, key, change_dict) abort
  return luaeval('require("orgmode.utils.dict_watcher").dict_changed(_A[1], _A[2], _A[3])', [a:change_dict, a:key, a:dict])
endfunction
```

Watch/unwatch for changes like this:

```lua
require('orgmode.utils.dict_watcher').watch_buffer_variable('org_indent_mode', function(change_dict)
  print('old', change_dict.old)
  print('new', change_dict.new)
end)
```

Unwatch

```lua
require('orgmode.utils.dict_watcher').unwatch_buffer_variable('org_indent_mode')
```

kristijanhusak (changes requested):
I didn't give this a test yet, just checked the code.
Lets do 2 things here:

1. Rebase this from `nightly` branch and set target to `nightly` branch.
2. Add the dict watcher I mentioned in below comment

lua/orgmode/ui/virtual_indent.lua
treatybreaker: So the only issue I see with dictwatcher is that there's genuine desire to deprecate it upstream:

https://github.com/neovim/neovim/issues/21469

I'll do it if you're ok with the above in mind. Just figured you should know.

kristijanhusak: Add the link to the above issue in the `dict_watcher` file as comment to have it as a reference, and lets proceed with it. I don't think it will be deprecated that easily since it's used by few popular GUIs.

treatybreaker: Yep, sounds good.

treatybreaker: Done.

treatybreaker:
> I didn't give this a test yet, just checked the code. Lets do 2 things here:
> 
>     1. Rebase this from `nightly` branch and set target to `nightly` branch.
> 
>     2. Add the dict watcher I mentioned in below comment

1.) Done
2.) Done

On \#2 though, there's a slight catch. `dictwatcheradd` doesn't return *any* context of which buffer the variable was changed in. This means we have to loop all of the VirtualIndent instances to determine which one had its variable changed.

Still *way* more efficient than my timer approach. The efficiency hit from having to do the loop is negligible imo, just feels nasty. Shame I couldn't find a way to extract the buffer in which the variable was updated in.

kristijanhusak:
> On #2 though, there's a slight catch. dictwatcheradd doesn't return any context of which buffer the variable was changed in. This means we have to loop all of the VirtualIndent instances to determine which one had its variable changed.

We could just store the buffer number as a buffer variable, and read that from the 3rd param of the watcher callback.

1. In the `ftplugin/org.lua` add something like `vim.b.org_bufnr = vim.fn.bufnr()`.
2. In the `watch_buffer_variable` callback, grab the 3rd param which is the whole `vim.b` value for the buffer. Take the buffer number from `org_bufnr`.

treatybreaker:
@kristijanhusak Done as part of the latest force push, pretty smart idea you had there. I definitely wouldn't have thought of that.

As an aside, sorry for the back and forth. I'd like to respond and nail these problems within the hour of your responses, but I'm getting tied up in things outside of github. I'm hoping it's not too much a bother. 😅

kristijanhusak (approved):
Ok looks good now, thanks!

Sorry for delaying it a bit, I was busy working on improvements for markup highlighter. 

> As an aside, sorry for the back and forth. I'd like to respond and nail these problems within the hour of your responses, but I'm getting tied up in things outside of github. I'm hoping it's not too much a bother

Don't worry about it. There's no need to rush. This is FOSS, no one is expecting you to be available :)

treatybreaker:
> Don't worry about it. There's no need to rush. This is FOSS, no one is expecting you to be available :)

Good point. 

> Sorry for delaying it a bit, I was busy working on improvements for markup highlighter.

No worry about the delays, nobody is paying you to work on this. I just run straight from the branch until it gets merged, a literal one line change in my Neovim config. 😄

PR #659, opened 2024-01-28
feat: add `org-indent-mode-turns-on-hiding-stars` equivalent

treatybreaker:
Follow on from #627.

Bullet points:
- Added new config flag: `org_indent_mode_turns_on_hiding_stars` which is enabled by default
- Added docs reflecting new configuration flag

Crosses another one off [my list](https://github.com/nvim-orgmode/orgmode/issues/422#issuecomment-1910991160) of indent mode features.

### DIFF
--- a/DOCS.md
+++ b/DOCS.md
@@ -301,8 +301,8 @@ Possible values:
 *type*: `boolean`<br />
 *default value*: `true`<br />
 Possible values:
-* `true` - Enable [`org_hide_leading_stars`](#org_hide_leading_stars) by default when [`org_startup_indented`](#org_startup_indented) is enabled.
-* `false` - Do not modify the value in [`org_hide_leading_stars`](#org_hide_leading_stars) by default when [`org_startup_indented`](#org_startup_indented) is enabled.
+* `true` - Enable [`org_hide_leading_stars`](#org_hide_leading_stars) by default when [`org_indent_mode`](#org_startup_indented) is enabled for buffer (`vim.b.org_indent_mode = true`).
+* `false` - Do not modify the value in [`org_hide_leading_stars`](#org_hide_leading_stars) by default when [`.org_indent_mode`](#org_startup_indented) is enabled for buffer (`vim.b.org_indent_mode = true`).
 
 #### **org_src_window_setup**
 *type*: `string|function`<br />
--- a/DOCS.md
+++ b/DOCS.md
@@ -296,6 +296,14 @@ Possible values:
 * `true` - Disable [`org_adapt_indentation`](#org_adapt_indentation) by default when [`org_startup_indented`](#org_startup_indented) is enabled.
 * `false` - Do not disable [`org_adapt_indentation`](#org_adapt_indentation) by default when [`org_startup_indented`](#org_startup_indented) is enabled.
 
+#### **org_indent_mode_turns_on_hiding_stars**
+
+*type*: `boolean`<br />
+*default value*: `true`<br />
+Possible values:
+* `true` - Enable [`org_hide_leading_stars`](#org_hide_leading_stars) by default when [`org_startup_indented`](#org_startup_indented) is enabled.
+* `false` - Do not modify the value in [`org_hide_leading_stars`](#org_hide_leading_stars) by default when [`org_startup_indented`](#org_startup_indented) is enabled.
+
 #### **org_src_window_setup**
 *type*: `string|function`<br />
 *default value*: "top 16new"<br />
--- a/lua/orgmode/colors/highlighter/stars.lua
+++ b/lua/orgmode/colors/highlighter/stars.lua
@@ -14,8 +14,20 @@ function OrgStars:new(opts)
   return data
 end
 
+function OrgStars:_is_enabled(bufnr)
+  if config.org_hide_leading_stars then
+    return true
+  end
+
+  if vim.b[bufnr].org_indent_mode and config.org_indent_mode_turns_on_hiding_stars then
+    return true
+  end
+
+  return false
+end
+
 function OrgStars:on_line(bufnr, line)
-  if not config.org_hide_leading_stars then
+  if not self:_is_enabled(bufnr) then
     return
   end
 
--- a/lua/orgmode/colors/highlights.lua
+++ b/lua/orgmode/colors/highlights.lua
@@ -102,13 +102,11 @@ function M.define_org_headline_colors(faces)
   for _, face in pairs(faces) do
     table.insert(contains, face)
   end
-  if config.org_hide_leading_stars then
-    if not ts_highlights_enabled then
-      vim.cmd([[syn match OrgHideLeadingStars /^\*\{2,\}/me=e-1 contained]])
-    end
-    vim.cmd([[hi default OrgHideLeadingStars ctermfg=0 guifg=bg]])
-    table.insert(contains, 'OrgHideLeadingStars')
+  if not ts_highlights_enabled then
+    vim.cmd([[syn match OrgHideLeadingStars /^\*\{2,\}/me=e-1 contained]])
   end
+  vim.cmd([[hi default OrgHideLeadingStars ctermfg=0 guifg=bg]])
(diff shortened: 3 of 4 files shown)


lua/orgmode/config/init.lua
```diff
@@ -42,6 +42,9 @@ function Config:extend(opts)
   self.opts = vim.tbl_deep_extend('force', self.opts, opts)
   if self.org_startup_indented then
     self.org_adapt_indentation = not self.org_indent_mode_turns_off_org_adapt_indentation
+    if self.org_indent_mode_turns_on_hiding_stars then
+      self.org_hide_leading_stars = true
```
kristijanhusak, line 46: I'm not 100% sure how Emacs does it, but we might need to set this to true/false depending on the `org_indent_mode` variable.

kristijanhusak: Yes, I confirmed that's the case. It hides the stars only if `org_indent_mode` is enabled.

Let's add the check here  https://github.com/nvim-orgmode/orgmode/blob/7317cc97c45521d61104362adb30bcd299d1060c/lua/orgmode/colors/highlighter/stars.lua?plain=1#L18. Check for `org_indent_mode_turns_on_hiding_stars` and `vim.b[bufnr]org_indent_mode`. Depending on those proceed with hiding the stars.

treatybreaker: You beat me to the punch (I had a whole wall of text confirming the case lol). There is an issue though with just doing the check and calling it a day.

`org_indent_mode_turns_on_hiding_stars` in Emacs can be set on a buffer basis. Meaning that isn't *just* a global variable like we have right now. 

There's no way to toggle `org_indent_mode_turns_on_hiding_stars` in this PR for a single buffer (or at all for that matter) like you can in Emacs (granted I'm abusing Emacs `eval-expression` to do that).

With the above in mind, do you want me to migrate that to a `vim.b` variable similar to`vim.b.org_indent_mode` and make it toggleable?

kristijanhusak: > With the above in mind, do you want me to migrate that to a vim.b variable similar tovim.b.org_indent_mode and make it toggleable or is it not worth it?

I don't think it's worth it. Let's just rely on the `vim.b.org_indent_mode` and the configuration for now, and we can add the improvement later if necessary.

treatybreaker: > I don't think it's worth it. Let's just rely on the vim.b.org_indent_mode and the configuration for now, and we can add the improvement later if necessary.

Ok sounds good. I can already see a small bug coming though so I'll get a PR out later this week to fix it.

Since `org_hide_leading_stars` is a "global" (meaning defined in `config`) variable in terms of the Orgmode plugin, reacting to `vim.b.org_indent_mode` will change the leading stars setting for other org buffers as well because they all look at `config.org_hide_leading_stars`.

Something to keep in mind.

treatybreaker: Done as part of the latest force push.

lua/orgmode/config/init.lua
```diff
@@ -42,6 +42,9 @@ function Config:extend(opts)
   self.opts = vim.tbl_deep_extend('force', self.opts, opts)
   if self.org_startup_indented then
     self.org_adapt_indentation = not self.org_indent_mode_turns_off_org_adapt_indentation
+    if self.org_indent_mode_turns_on_hiding_stars then
```
kristijanhusak, line 45: Let's remove this and use all the checks only in the `stars` highlighter.

lua/orgmode/colors/highlighter/stars.lua
```diff
@@ -15,7 +15,7 @@ function OrgStars:new(opts)
 end
 
 function OrgStars:on_line(bufnr, line)
-  if not config.org_hide_leading_stars then
+  if not config.org_hide_leading_stars or not (vim.b.org_indent_mode and config.org_hide_leading_stars) then
```
kristijanhusak, line 18: Lets use `org_indent_mode` + `org_indent_mode_turns_on_hiding_stars` here. Since it's a more complex check, you can extract it to a private method `_is_enabled`:

```lua
function OrgStars:_is_enabled(bufnr)
  if config.org_hide_leading_stars then
    return true
  end

  if vim.b[bufnr].org_indent_mode and config.org_indent_mode_turns_on_hiding_stars then
    return true
  end

  return false
end
```

DOCS.md
```diff
@@ -296,6 +296,14 @@ Possible values:
 * `true` - Disable [`org_adapt_indentation`](#org_adapt_indentation) by default when [`org_startup_indented`](#org_startup_indented) is enabled.
 * `false` - Do not disable [`org_adapt_indentation`](#org_adapt_indentation) by default when [`org_startup_indented`](#org_startup_indented) is enabled.
 
+#### **org_indent_mode_turns_on_hiding_stars**
+
+*type*: `boolean`<br />
+*default value*: `true`<br />
+Possible values:
+* `true` - Enable [`org_hide_leading_stars`](#org_hide_leading_stars) by default when [`org_startup_indented`](#org_startup_indented) is enabled.
+* `false` - Do not modify the value in [`org_hide_leading_stars`](#org_hide_leading_stars) by default when [`org_startup_indented`](#org_startup_indented) is enabled.
```
kristijanhusak, line 305: Let's update docs to refer to `org_indent_mode` instead of `org_startup_indented`

```suggestion
* `true` - Enable [`org_hide_leading_stars`](#org_hide_leading_stars) by default when [`org_indent_mode`](#org_startup_indented) is enabled for buffer (`vim.b.org_indent_mode = true`).
* `false` - Do not modify the value in [`org_hide_leading_stars`](#org_hide_leading_stars) by default when [`.org_indent_mode`](#org_startup_indented) is enabled for buffer (`vim.b.org_indent_mode = true`).
```

kristijanhusak (changes requested):
Lets just move things around a bit so it works only for buffers with org_indent_mode.

lua/orgmode/colors/highlighter/stars.lua
treatybreaker: Done as part of latest force push.

lua/orgmode/config/init.lua
treatybreaker: Done as part of latest force push.

treatybreaker:
As an aside, do you mind if I PR a quick `make format` to nightly? Getting assaulted by the linter on every pre-push is driving me ever so slightly insane.

kristijanhusak (changes requested):
I formatted the code on nightly, you can rebase.

Regarding the PR, it works ok, but the hl group for hiding the stars is not applied.

We also need these changes:

```diff
diff --git a/lua/orgmode/colors/highlights.lua b/lua/orgmode/colors/highlights.lua
index dcb1a7a..618095e 100644
--- a/lua/orgmode/colors/highlights.lua
+++ b/lua/orgmode/colors/highlights.lua
@@ -102,13 +102,11 @@ function M.define_org_headline_colors(faces)
   for _, face in pairs(faces) do
     table.insert(contains, face)
   end
-  if config.org_hide_leading_stars then
-    if not ts_highlights_enabled then
-      vim.cmd([[syn match OrgHideLeadingStars /^\*\{2,\}/me=e-1 contained]])
-    end
-    vim.cmd([[hi default OrgHideLeadingStars ctermfg=0 guifg=bg]])
-    table.insert(contains, 'OrgHideLeadingStars')
+  if not ts_highlights_enabled then
+    vim.cmd([[syn match OrgHideLeadingStars /^\*\{2,\}/me=e-1 contained]])
   end
+  vim.cmd([[hi default OrgHideLeadingStars ctermfg=0 guifg=bg]])
+  table.insert(contains, 'OrgHideLeadingStars')
   for i, color in ipairs(headline_colors) do
     local j = i
     while j < 40 do
```

treatybreaker:
@kristijanhusak I applied your patch and rebased on top of nightly for the formatting

kristijanhusak (approved):
Looks good, thanks!

PR #669, opened 2024-02-17
feat(indent): Always check for b:org_indent_mode if available

kristijanhusak:
Always check for `vim.b.org_indent_mode` before applying indentation to the buffer.
If it's `true`, treat all indentations in a buffer as if the configuration has `org_startup_indented` enabled.

@PriceHiller can you review and test this to see if it works ok?

### DIFF
--- a/lua/orgmode/config/init.lua
+++ b/lua/orgmode/config/init.lua
@@ -394,6 +394,7 @@ function Config:ts_highlights_enabled()
       return false
     end
 
+    ---@diagnostic disable-next-line: param-type-mismatch
     if type(hl_module.disable) == 'table' and vim.tbl_contains(hl_module.disable, 'org') then
       return false
     end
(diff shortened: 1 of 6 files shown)


PriceHiller:
> @PriceHiller can you review and test this to see if it works ok?

Will check here in a bit when I get home

lua/orgmode/config/init.lua
```diff
@@ -413,33 +414,25 @@ function Config:respect_blank_before_new_entry(content, option, prepend_content)
...
 
----@param amount number
----@return string
-function Config:get_indent(amount)
-  if self.org_adapt_indentation then
-    return string.rep(' ', amount)
+---Check if buffer should apply indentation
+---@param bufnr? number
+function Config:should_indent(bufnr)
+  if bufnr > -1 and vim.b[bufnr].org_indent_mode then
```
PriceHiller, line 420: Since `bufnr` is optional according to the annotation on line 418, ideally we'd also check if `bufnr` exists:

```suggestion
  if bufnr and bufnr > -1 and vim.b[bufnr].org_indent_mode then
```

This will keep callers who don't provide a `bufnr` from having their head explode when `nil` is compared `-1`. 

Alternatively, you could make providing a `bufnr` non-optional. It seems all other callers of this function currently *do* provide a `bufnr` -- seems it may be better to require it.

lua/orgmode/files/headline.lua
```diff
@@ -113,8 +113,8 @@ function Headline:demote(amount, recursive, dryRun)
       if line:sub(1, 1) == '*' then
         lines[i] = string.rep('*', amount) .. line
       else
-        if config.org_adapt_indentation then
-          lines[i] = config:apply_indent(line, amount)
+        if config:should_indent(self.file:bufnr()) then
+          lines[i] = self:_apply_indent(line, amount)
         else
           line, _ = line:gsub('^%s+', '')
           local indent_amount = indent.indentexpr(start_line + i)
```
PriceHiller, line 120: Add:
```suggestion
          local indent_amount = indent.indentexpr(start_line + i, self.file:bufnr())
```
to take advantage of the new `bufnr` arg on the `indentexpr` function.

PriceHiller (changes requested):
Something I'll do is take a look at the `virtual_indent.lua` file and update it to take advantage of some of the changes here once this is merged.

I'll have to pull in the `files` object to take advantage of the headline methods used in this PR. That should unify these a bit more.

lua/orgmode/files/headline.lua
kristijanhusak: Added it, thanks.

lua/orgmode/config/init.lua
kristijanhusak: The idea was to have it mandatory. Updated the annotation.

PriceHiller (approved):
That's all that stuck out to me.

Again I'll make a PR later this weekend or beginning of this week to incorporate some of the headline methods into the `virtual_indent` implementation.

PR #681, opened 2024-02-24
refactor(virtual-indent): do not set virtual indents on empty lines

PriceHiller:
This PR has two elements:
- Removes setting indents on blank lines
- Updated the tests to check for blank lines virtual columns, as well as moving `content_virtcols` above both virtual indent tests so both use the same content.

This improves compatibility with plugins like [Headlines.nvim](https://github.com/lukas-reineke/headlines.nvim) and others that leverage extmarks where virtual indents may apply.

Performance impact for pulling each individual line's content is basically none from my testing -- even spamming thousands of lines I saw no real slowdown related to this change.

#### Before with `Headlines.nvim` on latest nightly:
image
Notice the indent before the bottom of the headlines causing a visual error.

#### After with `Headlines.nvim` on latest nightly:
image
And now those indents are gone.

This is really the behavior that should be expected anyhow -- we really *shouldn't* be setting virtual indentation on blank lines as there isn't anything to indent in the first place.


PriceHiller:
Nevermind, this is not so *great*. Seemed good in the moment. I'll rework this a bit specifically for compat with `Headlines.nvim` or figure out an alternative solution.

This is annoying though, doesn't seem to be tied in with the extmarks priority as far as I can tell.

PR #683, opened 2024-02-25
refactor(agenda): use extmarks to set tags in right column

PriceHiller:
Before this change:
agenda-tags-extmark-before

Notice that the tags end up off screen and do not get shifted to align to the right side of the screen.

After this change:
agenda-tags-extmark-after

Notice that the tags are correctly aligned when the window is resized. Tags are not off screen at all.

### DIFF
--- a/lua/orgmode/agenda/init.lua
+++ b/lua/orgmode/agenda/init.lua
@@ -17,7 +17,9 @@ local Promise = require('orgmode.utils.promise')
 ---@field views table[]
 ---@field filters OrgAgendaFilter
 ---@field files OrgFiles
-local Agenda = {}
+local Agenda = {
+  _ns_id = vim.api.nvim_create_namespace('orgmode.ui.agenda'),
+}
 
 ---@param opts? table
 function Agenda:new(opts)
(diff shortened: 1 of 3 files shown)


PriceHiller:
Looks like the latest nightly causes the CI to fail. Not an issue with this PR.

Running `make test` with the latest nightly (`NVIM v0.10.0-dev-2445+g2e1f5055a`) causes the tests to fail with or without these changes.

PriceHiller:
The test error should be fixed by the PR I opened at #684

PriceHiller:
Rebased on top of #684 since it got merged.

kristijanhusak:
Tags should re-align once you redraw the agenda with `r`. That is not happening so it's a bug on that side.
I'm not sure how I feel about this. I'd prefer to leave it the original Orgmode way and just fix the re-align after redrawing.

This is how it looks in Emacs:
emacs

This is Neovim with wrapping disabled:
neovim-nowrap

And Neovim with wrapping enabled:
neovim-wrap

PriceHiller:
I didn't realize how orgmode on Emacs handles it before opening this PR -- so extmarks is kinda non-standard to the Emacs implementation.

> I pushed a fix for the realigning after render. Let me know how it looks now.

It aligns *much* better now, but still doesn't dynamically reflow etc. like extmarks and alignment gets really out of sorts when the window width is small. It seems to match up with Emacs though.

Would you be open to supporting a nonstandard extmark approach as well hidden behind a config flag? I personally prefer the alignment of extmarks and the fact they take priority over long titles as well as working well on a single line without word wrapping.

kristijanhusak:
I guess we could add it. Add an `agenda` key under [ui in config](https://github.com/nvim-orgmode/orgmode/blob/master/lua/orgmode/config/defaults.lua#L192), and add the flag there, something like `use_virtual_text_for_tags = false`.

PriceHiller:
> I guess we could add it. Add an agenda key under [ui in config](https://github.com/nvim-orgmode/orgmode/blob/master/lua/orgmode/config/defaults.lua#L192), and add the flag there, something like use_virtual_text_for_tags = false.

Cool, then I'll close this PR and make a new one with the changes from master when I get a sec.

PR #684, opened 2024-02-25
fix(mappings): use explicit bufnr for meta return in `apply_text_edits`

PriceHiller:
Copied from the commit message:

> Since https://github.com/neovim/neovim/commit/2e1f5055acdef650c27efc4afdf8606037ec021b, Neovim now explicitly requires the bufnr for `vim.lsp.util.apply_text_edits` to be fully resolved — meaning it cannot be 0 or null.

This just ensures nightly users don't run into errors when using `meta return` and that the tests for nightly pass as expected.


kristijanhusak (approved):
Nice, thanks!

PR #689, opened 2024-02-27
Use the REPEAT_TO_STATE property

massix:
Hello,

I honestly do not know if this is part of Emacs because I started using org-mode directly in Neovim thanks to this amazing plugin.

One thing that really bothered me with the current version of the plugin was that the `nvim-cmp` was proposing a `REPEAT_TO_STATE` property completion, but it wasn't used, and I know this is something I use a lot (lot of recurring meetings at work, like daily and stuff).

So some time ago I made this fix and now I adapted it to the recent modifications you've done. Let me know if there's something I can do!

### DIFF
--- a/lua/orgmode/org/mappings.lua
+++ b/lua/orgmode/org/mappings.lua
@@ -1022,7 +1022,7 @@ function OrgMappings:_change_todo_state(direction, use_fast_access)
     elseif direction == 'prev' then
       next_state = todo_state:get_prev()
     elseif direction == 'reset' then
-      next_state = todo_state:get_todo()
+      next_state = headline:get_property('REPEAT_TO_STATE') or todo_state:get_todo()
     end
   end
 
--- a/lua/orgmode/org/mappings.lua
+++ b/lua/orgmode/org/mappings.lua
@@ -1022,7 +1022,20 @@ function OrgMappings:_change_todo_state(direction, use_fast_access)
     elseif direction == 'prev' then
       next_state = todo_state:get_prev()
     elseif direction == 'reset' then
-      next_state = headline:get_property('REPEAT_TO_STATE') or todo_state:get_todo()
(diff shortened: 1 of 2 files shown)


lua/orgmode/org/mappings.lua
```diff
@@ -1022,7 +1022,20 @@ function OrgMappings:_change_todo_state(direction, use_fast_access)
     elseif direction == 'prev' then
       next_state = todo_state:get_prev()
     elseif direction == 'reset' then
-      next_state = todo_state:get_todo()
+      local repeat_to_state = headline:get_property('REPEAT_TO_STATE')
```
kristijanhusak, line 1025: Let's create a method on `TodoState` called `get_reset_todo`, and pass the headline inside.
Then we can figure out all the logic in there for this:

```lua
next_state = todo_state:get_reset_todo(headline)
```


Inside, we should do two things:
1. Check if headline has property as you do here, if yes, goto step 4
2. if no property on the headline, check if config has `org_todo_repeat_to_state` ( add this with default `nil`, also a documentation update would be nice)
3. If no config option, use the one that we use now
4. Check if value is valid by looking into `todo_keywords.keys[todo_value]`

kristijanhusak (changes requested):
> I honestly do not know if this is part of Emacs because I started using org-mode directly in Neovim thanks to this amazing plugin.

It actually does, you can read about it here 
https://orgmode.org/manual/Repeated-tasks.html

Generally this is good, but I would like to expand it to be also configurable from main configuration as Emacs also supports it. The comment below should help you out.

lua/orgmode/org/mappings.lua
massix: It should be ok now, let me know if there's something more I can do!

kristijanhusak (commented):
Looks good, thanks!

kristijanhusak (approved):
Looks good, thanks!

PR #690, opened 2024-02-28
Faster statusline

massix:
Hello,
With the last updates I've noticed that the `get_statusline()` function was slowing down my Neovim instance a lot and I think the problem is related with the way the clock gets the currently clocked in headline, which is basically parsing all files and retrieving the first clocked in headline.

Now, on my PC the slow down was barely noticeable, but on my Android phone it became a real issue (despite it being a good smartphone) so I did my best trying to patch it.

The basic idea is that now, with the modifications here, we pay a little upfront cost (it is noticeable on my phone but it is only at startup, I could not test it on my PC yet, will do later this week) but afterwards we store the information of the clocked headline in the instance of the clock itself, making it faster to access the information to display the statusline.

I've been testing this code for a few hours and it is quite stable and reliable, the only problem I see is if someone deletes the LOGBOOK entry manually, the clocked headline won't be reset.. probably we could run an async job every now and then to check if the clock is still valid?

Also, with this modification it is easier for hackers to retrieve the current clocked headline and [customise the statusline](https://github.com/massix/dot-termux/blob/20bb90720353d98cb32fd500753bfc709e9b1919/nvim/lua/plugins/lualine.lua#L71) message, we don't even need a debouncer anymore since we're not modifying or touching the files.

Please let me know if you have better ideas!

### DIFF
--- a/lua/orgmode/init.lua
+++ b/lua/orgmode/init.lua
@@ -65,6 +65,7 @@ function Org:init()
   self.clock = require('orgmode.clock'):new({
     files = self.files,
   })
+  self.clock:init()
   self.completion = require('orgmode.org.autocompletion'):new({ files = self.files })
   self.statusline_debounced = require('orgmode.utils').debounce('statusline', function()
     return self.clock:get_statusline()
--- a/lua/orgmode/init.lua
+++ b/lua/orgmode/init.lua
@@ -65,7 +65,6 @@ function Org:init()
   self.clock = require('orgmode.clock'):new({
     files = self.files,
   })
-  self.clock:init()
   self.completion = require('orgmode.org.autocompletion'):new({ files = self.files })
   self.statusline_debounced = require('orgmode.utils').debounce('statusline', function()
     return self.clock:get_statusline()
--- a/lua/orgmode/clock/init.lua
+++ b/lua/orgmode/clock/init.lua
@@ -17,20 +17,27 @@ function Clock:new(opts)
   return data
 end
 
+-- When first loading, check if there are active clocks
+function Clock:init()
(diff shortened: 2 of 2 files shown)


lua/orgmode/init.lua
```diff
@@ -65,6 +65,7 @@ function Org:init()
   self.clock = require('orgmode.clock'):new({
     files = self.files,
   })
+  self.clock:init()
```
kristijanhusak, line 68: You can move this call inside the `new()` function and remove it from here.

kristijanhusak (approved):
Thanks for the PR!
I wanted to address this issue at some point but totally forgot, since I'm not actively using clocking.

This is a good solution for now. It has some drawbacks that you mentioned, but some of them existed even before the refactor, so I don't think we need to worry about them at this point. We can have that as an improvement.

Just move the `init` function per my suggestion in the comment and it's good to go!

lua/orgmode/init.lua
massix: Done! Thanks for the review

massix:
> It has some drawbacks that you mentioned, but some of them existed even before the refactor

I will most probably do another PR next week to update the clock in the background using an async worker, so to solve the issue I mentioned in my first comment!

kristijanhusak (approved):

kristijanhusak:
Thanks for the PR!

> I will most probably do another PR next week to update the clock in the background using an async worker, so to solve the issue I mentioned in my first comment!

Instead of async worker, let's go with listening on `BufWritePost` on org files and check and re-set the clocked headline. That still leaves the issue if someone edits the file outside Neovim, but it should solve all other cases.

PR #698, opened 2024-03-08
fix: block content alignment

PriceHiller:
This PR has two commits as part of it.

The first commit ensures when block contents have a single line at the correct indentation level and the rest of the contents are indented more than that line, the block contents won't have their indentation changed.

The second commit makes the block content indentation ignore blank lines. Meaning it won't modify indentation if a blank line is "underindented" relative to the header as it was previously.

I wrote tests as part of each commit to exercise each path. Please see them to see specific examples. Each commit has its own tests associated, so it may be easier to view them that way.

Let me know if there's anything else I can do :).


kristijanhusak (approved):
Looks good, thanks!

PR #699, opened 2024-03-10
fix(highlights): make linked hls set `hi-default`

PriceHiller:
This PR allows the linked highlights to not override/respect already set highlights.

This ensures any colorschemes or custom highlights set *before* Orgmode is loaded do not have their highlights invalidated when Orgmode loads or is reloaded. By setting `default = true` for the linked highlights, lazy loading plays nicer with Orgmode.


kristijanhusak (commented):
Thanks!

kristijanhusak (approved):

PR #710, opened 2024-04-04
feat: rocks.nvim/luarocks support

mrcjkb:
Depends on https://github.com/nvim-orgmode/tree-sitter-org/pull/1

### DIFF
--- a/.github/workflows/luarocks.yml
+++ b/.github/workflows/luarocks.yml
@@ -39,4 +39,4 @@ jobs:
         with:
           version: ${{ env.LUAROCKS_VERSION }}
           dependencies: |
-            tree-sitter-orgmode == ${{ env.LUAROCKS_VERSION }}
+            tree-sitter-orgmode ~> 1
--- a/.github/workflows/luarocks.yml
+++ b/.github/workflows/luarocks.yml
@@ -1,0 +1,42 @@
+name: Push to Luarocks
+
+on:
+  push:
+    tags:
+      - '*'
+  release:
+    types:
+      - created
+  pull_request: # Runs test install without uploading
+  workflow_dispatch: # Allows to trigger manually
+
+jobs:
+  luarocks-upload:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0 # Required to count the commits
+      - name: Get Version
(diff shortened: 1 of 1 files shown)


.github/workflows/luarocks.yml
```diff
@@ -0,0 +1,29 @@
...
+      - name: Get Version
+        run: echo "LUAROCKS_VERSION=$(git describe --abbrev=0 --tags)" >> $GITHUB_ENV
+      - name: LuaRocks Upload
+        uses: nvim-neorocks/luarocks-tag-release@v5
+        env:
+          LUAROCKS_API_KEY: ${{ secrets.LUAROCKS_API_KEY }}
+        with:
+          version: ${{ env.LUAROCKS_VERSION }}
+          dependencies: |
+            tree-sitter-orgmode == ${{ env.LUAROCKS_VERSION }}
```
mrcjkb, line 29: This assumes that there will always be a `tree-sitter-orgmode` release with exactly the same tag (published to luarocks before this one).
An alternative option would be to say 

```suggestion
            tree-sitter-orgmode <= ${{ env.LUAROCKS_VERSION }}
```

But I'm not sure if luarocks will also upgrade the dependency when upgrading the package.

Probably the best solution would be to say

```suggestion
            tree-sitter-orgmode >= ${{  env.MIN_PARSER_VERSION }}, <= ${{ env.LUAROCKS_VERSION }}
```

... where `MIN_PARSER_VERSION` could be hard coded (in which case it would need to be maintained manually) or generated somehow.

kristijanhusak: Is there a way to somehow grab the latest version? Or just use some wildcard that will automatically install the latest major version? Something like `1.*` ?
I don't plan to introduce breaking changes to the parser so using this should be safe.

mrcjkb: Yes, you can do `tree-sitter-org ~> 1`, which translates to `>= 1, < 2`

kristijanhusak: Ok we can hardcode it to that for now.

kristijanhusak (approved):
Thanks!

kristijanhusak:
@mrcjkb do you have an idea why this started to fail https://github.com/nvim-orgmode/orgmode/actions/runs/10071869912/job/27842740421?pr=780#step:8:463 ?

mrcjkb:
Yep. I'll PR a fix later today :smile:

PR #716, opened 2024-04-10
fix(virtual indent): do not return virtual indentation for headlines

PriceHiller:
Copied from the body of the commit message:

> This ensures when the treesitter parser has errors, the manual indent
sizing doesn't return virtual indentation for headlines.

### Before the Fix
messed-up-vindent-headlines

### After the Fix
fixed-vindent-headlines

### Misc

What's annoying is that the tests didn't catch this -- need a test to check raw insertion and typing I guess. That would've caught this for sure. The current tests loads the file, where the virtual indent works as expected.

Something to put on my TODO when I have more time.

Again, if anything crops up with Virtual Indent, ping me. I'll gladly look at it.

### DIFF
--- a/lua/orgmode/ui/virtual_indent.lua
+++ b/lua/orgmode/ui/virtual_indent.lua
@@ -53,9 +53,12 @@ function VirtualIndent:_get_indent_size(line, tree_has_errors)
   if tree_has_errors then
     local linenr = line
     while linenr > 0 do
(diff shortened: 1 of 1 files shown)


kristijanhusak (approved):
I noticed things jumping around from time to time but I wasn't able to reproduce it consistently.

This is probably the culprit. Thanks!

PR #719, opened 2024-04-12
Fix usage of 0 for buffer handle instead of `OrgFile:bufnr()` where possible

chipsenkbeil:
Gave a quick scan of both `OrgFile` and `OrgHeadline` for additional situations of `nvim_buf_set_text` and `nvim_buf_set_lines` where 0 was used instead of the `OrgFile` buffer number.

I left some TODO comments for you to look through as I don't know how you'd want to handle these. I'm also totally fine with you writing this yourself. 😄 This was just easier for me to do to show areas where there may be gotchas if someone is using the API expecting it to update in once place, but it is instead updating it somewhere else.

### DIFF
--- a/lua/orgmode/files/file.lua
+++ b/lua/orgmode/files/file.lua
@@ -432,7 +432,7 @@ function OrgFile:set_node_text(node, text, front_trim)
     end_row = last_line
 
     local end_row_text = vim.api.nvim_buf_get_lines(bufnr, end_row, end_row + 1, true)[1]
-    end_col = string.len(end_row_text) - 2
+    end_col = string.len(end_row_text)
   end
   local ok = pcall(vim.api.nvim_buf_set_text, bufnr, start_row, start_col, end_row, end_col, replacement)
   return ok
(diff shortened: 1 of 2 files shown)


chipsenkbeil:
@kristijanhusak I've clearly done something wrong starting with the commit, [Replace call to vim.fn.line() and vim.fn.col() with appropriate neovim APIs that support bufnr](https://github.com/nvim-orgmode/orgmode/pull/719/commits/8a76bc64886158f147fe9943ea731f89bd3f53c9). Making that adjustment caused it to work for me, but maybe I have something off with the numbers. Was trying to follow when you had an zero-index and one-index, but there's a mix and it gets confusing.

Anyway, I'll leave this for you to see how to go about adjusting these, and use this as an example of where some issues might occur. I ended up looking through your usage of `vim.fn` for API calls that are for the current buffer, and tried to replace them with neovim API calls that support specifying the buffer. I only looked within `OrgFile` and `OrgHeadline`, though.

kristijanhusak:
I fixed these in 5875037fa9c7, but the PR was very helpful to find all the occurrences. Thanks!

Issue #530, opened 2023-03-18
`org-log-repeat`


xiugaze:
### Does this feature exist in Emacs orgmode core?

Yes

### Orgmode link

https://doc.endlessparentheses.com/Var/org-log-repeat.html

### Additional context

This would be a very nice to have for building out calendars or class schedules. I would like to be able to disable this globally, or at the very least get rid of the `State "DONE" from TODO` bullet that gets added on a repeat.

kristijanhusak:
Added in https://github.com/nvim-orgmode/orgmode/pull/738.

PR #721, opened 2024-04-21
feat: org_log_repeat

vaioure:
Separate logging settings for repeating TODOs

Implements #530 

Unfortunately I couldn't figure out how to redraw the agenda view without setting redo in:

```lua
function Agenda:change_todo_state()
  return self:_remote_edit({
    action = 'org_mappings.todo_next_state',
    redo = true,
  })
end
```


kristijanhusak:
I did some refactoring and added support for this in https://github.com/nvim-orgmode/orgmode/pull/738, so closing.

Why do you need to redraw the whole agenda?

vaioure:
It's not strictly necessary, but I figured the entry should disappear from the agenda without having to manually refresh it since it's no longer accurate.

Anyways, thanks for implementing this.

PR #723, opened 2024-05-06
Update promise.wait to throw an error when timeout reached or interrupted

chipsenkbeil:
Per the comment I made in our main org roam issue, I discovered that `Promise.wait` does not throw an error when the timeout is reached. Instead, it returns `nil`.

This can be a problem if you're returning `OrgPromise<nil>`, as it misleads the promise as succeeding when in reality it may have not completed yet.

This updates `Promise.wait` to check the return values from `vim.wait` to determine if a value should be returned or if an error should be thrown.

### DIFF
--- a/lua/orgmode/utils/promise.lua
+++ b/lua/orgmode/utils/promise.lua
@@ -289,15 +289,17 @@ function Promise.wait(self, timeout)
     return error(value)
   end
 
-  if not success and code == -1 then
-    return error('promise timeout of ' .. tostring(timeout) .. 'ms reached')
-  elseif not success and code == -2 then
-    return error('promise interrupted')
-  elseif not success then
-    return error('promise failed with unknown reason')
+  if success then
+    return value
   end
 
-  return value
+  if code == -1 then
+    return error('promise timeout of ' .. tostring(timeout) .. 'ms reached')
+  elseif code == -2 then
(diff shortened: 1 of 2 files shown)


lua/orgmode/utils/promise.lua
```diff
@@ -288,6 +289,14 @@ function Promise.wait(self, timeout)
...
 
+  if not success and code == -1 then
+    return error('promise timeout of ' .. tostring(timeout) .. 'ms reached')
+  elseif not success and code == -2 then
+    return error('promise interrupted')
+  elseif not success then
+    return error('promise failed with unknown reason')
+  end
+
   return value
```
kristijanhusak, line 300: Lets reverse the success check to make the conditionals more readable:

```suggestion
  if success then
    return value
  end

  if code == -1 then
    return error('promise timeout of ' .. tostring(timeout) .. 'ms reached')
  elseif code == -2 then
    return error('promise interrupted')
  end

  return error('promise failed with unknown reason')
```

kristijanhusak (changes requested):
Looks good, just some minor refactoring around the checks.

chipsenkbeil:
> Looks good, just some minor refactoring around the checks.
> 

Good to go :)

kristijanhusak:
Thanks!

PR #728, opened 2024-05-17
update doc to use correct highlight name

ktsujister:
According to [this file](https://github.com/nvim-orgmode/orgmode/blob/8ec0bcc6f6476d246159f738081256c97a7a9b2c/lua/orgmode/colors/highlighter/stars.lua#L36), highlight for leading stars should be `@org.leading_stars` instead of `@org.leading.stars`.


kristijanhusak (commented):
Thanks!

kristijanhusak (approved):
Thanks!

PR #731, opened 2024-05-21
feat(api): Allow to insert links through API

seflue:
To give plugins like Telescope-orgmode the ability to insert links, the API needs an extension.

The refactoring moves some methods from org.mappings into modules, where they should belong anyway, so they can be used from the new api method.

The new API method api.insert_link is mostly existing code moved from org.mappings.insert_link, which now just calls api.insert_link.

### DIFF
--- a/lua/orgmode/api/init.lua
+++ b/lua/orgmode/api/init.lua
@@ -1,6 +1,8 @@
 ---@diagnostic disable: invisible
 local OrgFile = require('orgmode.api.file')
 local OrgHeadline = require('orgmode.api.headline')
+local Hyperlinks = require('orgmode.org.hyperlinks')
+local Link = require('orgmode.org.hyperlinks.link')
 local orgmode = require('orgmode')
 
 ---@class OrgApiRefileOpts
@@ -8,6 +10,10 @@ local orgmode = require('orgmode')
 ---@field destination OrgApiFile | OrgApiHeadline
 
 ---@class OrgApi
+---@field load fun(name?: string|string[]): OrgApiFile|OrgApiFile[]
+---@field current fun(): OrgApiFile
+---@field refile fun(opts: OrgApiRefileOpts)
+---@field insert_link fun(link_location: string): boolean
 local OrgApi = {}
 
 ---@param name? string|string[] specific file names to return (absolute path). If ommitted, returns all loaded files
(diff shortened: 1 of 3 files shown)


seflue:
@kristijanhusak Please see my PR in [Telescope-orgmode](https://github.com/lyz-code/telescope-orgmode.nvim/pull/16) as an example, how this change can be used. I actually need this change for that feature.

lua/orgmode/api/init.lua
```diff
@@ -99,4 +105,54 @@ function OrgApi.refile(opts)
   return true
 end
 
+--- Insert a link to a given location at the current cursor position
+--- @param link_location string
+--- @return boolean
+function OrgApi.insert_link(link_location)
```
kristijanhusak, line 111: Let's abstract all this logic in `Hyperlinks` and then require that one here. I'm trying not to depend on `orgmode.api` in the core, but vice versa.

lua/orgmode/api/init.lua
```diff
@@ -1,13 +1,19 @@
...
+local Hyperlinks = require('orgmode.org.hyperlinks')
+local Link = require('orgmode.org.hyperlinks.link')
 local orgmode = require('orgmode')
 
 ---@class OrgApiRefileOpts
 ---@field source OrgApiHeadline
 ---@field destination OrgApiFile | OrgApiHeadline
 
 ---@class OrgApi
+---@field load fun(name?: string|string[]): OrgApiFile|OrgApiFile[]
```
kristijanhusak, line 13: These types are not necessary. Methods should appear in the types.

kristijanhusak (changes requested):
Lets just abstract the logic in different place and use it on both places accordingly.

lua/orgmode/api/init.lua
seflue: The reason was to make it easy to annotate the API module in a different plugin to help development. See [here](https://github.com/lyz-code/telescope-orgmode.nvim/pull/16/commits/c1492a655685f097886c0dc3edc3603dedd7d2cd#r1608568226)
Is there a way to tell the LSP where to find them without copying?

lua/orgmode/api/init.lua
seflue: I have to admit, that I just looked at other functions, which had similar calling schemes, but I agree, that it would be cleaner to move them into `Hyperlinks`.

lua/orgmode/api/init.lua
seflue: I removed the type hints here, but I will keep them for now in Telescope-orgmode.

lua/orgmode/api/init.lua
seflue: I moved the code as you suggested.

kristijanhusak (approved):
Looks good, thanks!

Issue #645, opened 2024-01-04
Support for org-refile-use-outline-path


chipsenkbeil:
### Does this feature exist in Emacs orgmode core?

Yes

### Orgmode link

* Manual mention: https://orgmode.org/org.html#index-org_002drefile_002duse_002doutline_002dpath
* Additional explanation: https://emacs.stackexchange.com/questions/36505/org-mode-refile-multiple-locations-with-same-name-as-target

While the manual mentions the variable, I can't actually find manual documentation on its values. There are multiple website discussions such as stackexchange that provide examples of the options, though.

### Additional context

Suppose I want to have an org file named `work.org` that looks like this:

```org
* Project 1
** Tasks
* Project 2
** Tasks 
```

I can only refile today using a destination like `work.org/Tasks`, which will place  under the first heading that matches. I'd like to have a singular file where I can file tasks into different headings without either

a. Making them top-level and distinct such as `Project 1 Tasks` and `Project 2 Tasks`
b. Breaking out into separate files such as `project_1.org/Tasks` and `project_2.org/Tasks` when refiling

chipsenkbeil:
Seems related to https://github.com/nvim-orgmode/orgmode/discussions/513.

seflue:
Interesting issue, which exposes a feature gap between nvim-orgmode and Emacs Orgmode. I would be curious to implement it, but I would prefer to first refactor the refiling code a bit more seriously, because this feature touches a couple of points, which currently treat headline names like a unique identifier within one org file.

This includes also some user interface decisions, which we need to discuss.

chipsenkbeil:
@seflue what can I do to help here? Is there a tracker for the refile refactoring?

I've got experience writing neovim plugins such as https://github.com/chipsenkbeil/distant.nvim, so maybe I can pitch in a bit.

seflue:
There is currently no tracker regarding this refactoring, but because I spend some time in the refiling code the last days while fixing some issue, I came to the conclusion, that it would make future changes easier to refactor it a little bit. Currently the code is sometimes a little bit ambiguous to read, also a bit entangled, which makes it challenging to understand. But because I already invested some time to grasp it, it would make sense to use this knowledge, to improve it.

[...]

PR #734, opened 2024-05-25
Add tests for archiving subtree

andreaswachowski:
Given
```
* foo
** bar
*** baz
```

and the cursor on the foo-line, when archiving, the archive context is added to the last headline:

```
* foo
** bar
*** baz
    :PROPERTIES:
    :ARCHIVE_TIME: ...
    ...
    :END:
```

Emacs however adds the archive context to the foo-headline (which seems to be more appropriate):

```
* foo
  :PROPERTIES:
  :ARCHIVE_TIME: ...
  ...
  :END:
** bar
*** baz
```

Hence this PR, which aligns nvim-orgmode's behaviour with Emacs.

BTW, I second @seflue 's suggestion from https://github.com/nvim-orgmode/orgmode/pull/641#issue-2061135549 and https://github.com/nvim-orgmode/orgmode/issues/645#issuecomment-1881724227 to extract Refile and Archive from Capture (probably analogous to, `org-capture.el`, `org-refile.el`, and `org-archive.el` in https://github.com/emacs-mirror/emacs/tree/master/lisp/org), to prevent it from becoming unwieldy further down the line. I am interested in fleshing out the archiving a bit more (e.g. adding ARCHIVE_OLPATH and making the context configurable (see [org-archive-save-context-info](https://github.com/emacs-mirror/emacs/blob/master/lisp/org/org-archive.el#L119))), but I could work on the refactoring first. It would be helpful if you could provide an indication as to whether that would be fine (or not). Either way, thanks a lot for your work!

### DIFF
--- a/tests/plenary/ui/mappings/archive_spec.lua
+++ b/tests/plenary/ui/mappings/archive_spec.lua
@@ -32,4 +32,31 @@ describe('Archive', function()
       '  :END:',
     }, vim.api.nvim_buf_get_lines(0, 0, -1, false))
   end)
+
+  it('should set properties on top-level headline when refiling subtree', function()
+    local file = helpers.create_agenda_file({
+      '* foobar',
+      '** baz',
+      '* foo',
+    })
+
+    vim.cmd([[exe "norm ,o$"]])
+    -- Pause to finish the archiving
+    vim.wait(50)
+    assert.are.same({
+      '* foo',
+    }, vim.api.nvim_buf_get_lines(0, 0, -1, false))
+
+    vim.cmd(('edit %s'):format(file.filename .. '_archive'))
+    assert.are.same({
+      '* foobar',
+      '  :PROPERTIES:',
+      '  :ARCHIVE_TIME: ' .. Date.now():to_string(),
+      '  :ARCHIVE_FILE: ' .. file.filename,
+      '  :ARCHIVE_CATEGORY: ' .. file:get_category(),
+      '  :ARCHIVE_TODO: ',
+      '  :END:',
+      '** baz',
+    }, vim.api.nvim_buf_get_lines(0, 0, -1, false))
+  end)
 end)


andreaswachowski:
Oh, I almost forgot: Evaluating `Date.now():to_string()` again in the test introduces a minuscule probability of making the test flaky. Since the string is only accurate to the minute, I have neglected that. I am not too familiar with Lua yet, I suppose it could be solved by providing a custom assertion to match regexes, or (if that exists in the Lua universe) by freezing the time (like with Ruby's timecop. Anyway just FYI in case that needs to be changed.

seflue:
> Oh, I almost forgot: Evaluating `Date.now():to_string()` again in the test introduces a minuscule probability of making the test flaky. Since the string is only accurate to the minute, I have neglected that. I am not too familiar with Lua yet, I suppose it could be solved by providing a custom assertion to match regexes, or (if that exists in the Lua universe) by freezing the time (like with Ruby's timecop. Anyway just FYI in case that needs to be changed.

I already fixed the flaky behavior of Date in PR #650 , but because [I couldn't convince](https://github.com/nvim-orgmode/orgmode/pull/650#discussion_r1455258736) @kristijanhusak to accept the changes and hadn't had the time to collect detailed arguments after working on this code months ago, the PR got stale.

I have it on my todo list though, to rebase and adapt the PR to current master. I could also try to separate my fixes to Date into a separate PR and I would appreciate, if you can confirm, that these fixes make your tests reliable. A stable Date class is crucial for a lot of features regarding time, so any help to stabilize it is heavily welcome.

andreaswachowski:
> > Oh, I almost forgot: Evaluating `Date.now():to_string()` again in the test introduces a minuscule probability of making the test flaky. Since the string is only accurate to the minute, I have neglected that. I am not too familiar with Lua yet, I suppose it could be solved by providing a custom assertion to match regexes, or (if that exists in the Lua universe) by freezing the time (like with Ruby's timecop. Anyway just FYI in case that needs to be changed.
> 
> I already fixed the flaky behavior of Date in PR #650 , but because [I couldn't convince](https://github.com/nvim-orgmode/orgmode/pull/650#discussion_r1455258736) @kristijanhusak to accept the changes and hadn't had the time to collect detailed arguments after working on this code months ago, the PR got stale.
> 
> I have it on my todo list though, to rebase and adapt the PR to current master. I could also try to separate my fixes to Date into a separate PR and I would appreciate, if you can confirm, that these fixes make your tests reliable. A stable Date class is crucial for a lot of features regarding time, so any help to stabilize it is heavily welcome.

I have only briefly looked at your PR, but I believe I meant a different, almost theoretical issue. `Date.now()` calls `os.time()`, which returns the current time in seconds precision. `Capture:refile_file_headline_to_archive(headline)` calls `Date.now()` and a very tiny bit later (nano- to microseconds probably), `Date.now()` is called in the test file.

If (a big if) the first call is made at, say, 12:45:59.999999 (where to_string will truncate this to 12:45), and the second call in the test is made at 12:46:00.0000000, the test will fail. So usually we would want to ensure that `Date.now()` (or `os.time()`) returns exactly the same value (e.g. by mocking a return value). Alternatively, since this is not actually the focus of the test, one could ignore the time altogether by using a regex.

Or is there another issue in the Date class? In any case, I haven't seen flakiness yet, but theoretically it is possible.

seflue:
> I have only briefly looked at your PR, but I believe I meant a different, almost theoretical issue. `Date.now()` calls `os.time()`, which returns the current time in seconds precision. `Capture:refile_file_headline_to_archive(headline)` calls `Date.now()` and a very tiny bit later (nano- to microseconds probably), `Date.now()` is called in the test file.
> 
> If (a big if) the first call is made at, say, 12:45:59.999999 (where to_string will truncate this to 12:45), and the second call in the test is made at 12:46:00.0000000, the test will fail. So usually we would want to ensure that `Date.now()` (or `os.time()`) returns exactly the same value (e.g. by mocking a return value). Alternatively, since this is not actually the focus of the test, one could ignore the time altogether by using a regex.
> 
> Or is there another issue in the Date class? In any case, I haven't seen flakiness yet, but theoretically it is possible.

Ah, thanks for adding a bit more context to your issue. Now I understand, that you actually meant flaky test in the sense of "with same input they sometimes succeed, sometimes fail".

The issue I had, was different (and I still cannot get the details together without spending some time in code archeology, because I forgot most of it) was in `Date:from_time_table`. Because it doesn't consistently takes all the state from the argument if present there, it dropped information, when the argument contained minutes and hours.

There are probably some tests missing, to expose the bug, but I stumbled over it, while working on my timepicker. I am currently trying to rebase #650 onto the current master, but because of a bigger refactoring, it takes quite some effort to fix all the merge conflicts.

kristijanhusak:
@andreaswachowski I pushed a change to master that should fix this issue. Can you give it a test?

andreaswachowski:
@kristijanhusak Thank you, nice fix and certainly cleaner than my approach, which shows my unfamiliarity with the code base. It works now, but you don't need to take my word for it: I rebased this PR and only kept the tests, which demonstrate the correctness of the fix (if you want to keep those.)

kristijanhusak:
@andreaswachowski test looks ok, but lets just move it to [archive_spec.lua](https://github.com/nvim-orgmode/orgmode/blob/2bc00728720753e645ca922e2b0ecc1bf018ac3a/tests/plenary/ui/mappings/archive_spec.lua?plain=1#L4) and use a shortcut to trigger the archive like to other test does. Thanks!

andreaswachowski:
> @andreaswachowski test looks ok, but lets just move it to [archive_spec.lua](https://github.com/nvim-orgmode/orgmode/blob/2bc00728720753e645ca922e2b0ecc1bf018ac3a/tests/plenary/ui/mappings/archive_spec.lua?plain=1#L4) and use a shortcut to trigger the archive like to other test does. Thanks!

I had not even noticed archive_spec.lua 🙈 Of course, I will have a look!

andreaswachowski:
@kristijanhusak Alright, so one of my tests was redundant as it was already included in `archive_spec.lua`. I've added one test that covers archiving a subtree.

Now there's one thing in the current commit I'd appreciate feedback on (on top of anything else you might have): You will notice that I basically copied the first test, including the check for whether the archive file is created. To make this work, I did not see another way as to specialise the glob with `0.org_archive` and `1.org_archive`, respectively, because the hashed path from `vim.fn.tempname` is exactly the same in both tests.

The problem is a) this implies knowledge about the inner workings of `vim.fn.tempname` and b) if the test execution order were to be randomized one day, and the tests are executed in different order, they would fail.

I'd thus prefer to just remove thopse two assertions (those with the glob) from the second test, since the first test covers that already (and then revert to `**/*.org_archive` in the first test). 

What do you think?

kristijanhusak:
Yeah you can leave the glob assertion only in the first test.

andreaswachowski:
Thanks! Done.

kristijanhusak (approved):
Looks good, thanks!

PR #742, opened 2024-06-02
fix(links): insert id link properly

seflue:
When inserting a stored id link, it is now properly split into link and description.

For this to work with the current implementation we need actually to insert id and headline title both into the link input. It get's split and correctly formatted afterwards.

### DIFF
--- a/lua/orgmode/org/hyperlinks/init.lua
+++ b/lua/orgmode/org/hyperlinks/init.lua
@@ -181,10 +181,10 @@ function Hyperlinks.get_link_to_headline(headline, path)
     id = headline:id_get_or_create()
   end
 
-  if not config.org_id_link_to_org_use_id or not id then
-    return ('file:%s::*%s'):format(path, title)
+  if config.org_id_link_to_org_use_id and id then
+    return ('id:%s::*%s'):format(id, title)
   end
-  return ('id:%s  %s'):format(id, title)
+  return ('file:%s::*%s'):format(path, title)
 end
 
 ---@param headline OrgHeadline
@@ -216,10 +216,9 @@ end
 function Hyperlinks.insert_link(link_location)
   local selected_link = Link:new(link_location)
   local desc = selected_link.url:get_target_value()
+
   if selected_link.url:is_id() then
-    local id_link = ('id:%s'):format(selected_link.url:get_id())
-    desc = link_location:gsub('^' .. vim.pesc(id_link) .. '%s+', '')
-    link_location = id_link
+    link_location = ('id:%s'):format(selected_link.url:get_id())
   end
 
   local link_description = vim.trim(vim.fn.OrgmodeInput('Description: ', desc or ''))


lua/orgmode/org/hyperlinks/init.lua
```diff
@@ -181,10 +181,10 @@ function Hyperlinks.get_link_to_headline(headline, path)
     id = headline:id_get_or_create()
   end
 
-  if not config.org_id_link_to_org_use_id or not id then
-    return ('file:%s::*%s'):format(path, title)
+  if config.org_id_link_to_org_use_id and id then
```
seflue, line 184: I just swapped the logic for easier readability (no double negation).

lua/orgmode/org/hyperlinks/init.lua
```diff
@@ -181,10 +181,10 @@ function Hyperlinks.get_link_to_headline(headline, path)
...
+    return ('id:%s::*%s'):format(id, title)
```
seflue, line 185: Using the same url syntax as in file links enables the Url class to parse the parts correctly without adding logic.

lua/orgmode/org/hyperlinks/init.lua
```diff
@@ -216,9 +216,10 @@ end
 function Hyperlinks.insert_link(link_location)
   local selected_link = Link:new(link_location)
```
seflue, line 217: The full url including the description a.k.a. headline title is inserted by the user via autocompletion. Link construction parses it and we can get the parts needed parts easily from it to assemble the actual link.

lua/orgmode/org/hyperlinks/init.lua
```diff
@@ -216,9 +216,10 @@ end
...
   local desc = selected_link.url:get_target_value()
+
   if selected_link.url:is_id() then
     local id_link = ('id:%s'):format(selected_link.url:get_id())
-    desc = link_location:gsub('^' .. vim.pesc(id_link) .. '%s+', '')
+    desc = selected_link.url:get_target_value()
```
seflue, line 222: And now as I review it myself, I see, that this is redundant ... I will remove it. :sweat_smile:

seflue:
@kristijanhusak I need a reviewer to merge. :sweat_smile:

kristijanhusak (approved):

kristijanhusak:
@chipsenkbeil can you double check if this can cause some issues with your org-roam? I think it shouldn't but I want to be on the safe side.

chipsenkbeil:
@kristijanhusak nothing stands out as an issue to me. @seflue is using the org roam plugin - now a collaborator on the project - so I'm assuming this stems from an issue encountered while using org roam.

seflue:
I actually only got aware about the functionality due to my exploration into orgroam. Because I still have a large amount of notes not in orgroam, I use it to connect the "old world" and the "new world" in a filepath-agnostic manner.
And immediately discovered, that "insert_link" is generating broken links. Because org-roam generates it's links by itself, the both plugins do not interfere.
I have actually a hard time to imagine, how the old code was supposed to work correctly. With this fix it works smoothly - as I said, I use both plugins together on a daily basis.

PR #744, opened 2024-06-04
fix(calendar): fix bug with empty date

seflue:
- simplify state by replacing month with date and month method
- handle rendering correctly when changing month
- work around a bug in date (will be fixed later)


seflue:
@kristijanhusak @chipsenkbeil I discovered a bug in `Calendar` while working with Org-Roam I may have introduced myself, because it exposes a bug in date. I didn't had the time to actually take a look at the Date class, but I managed to implement a tiny workaround in `Calendar` and even add some tests.

I would appreciate if we could review and merge this quickly, because the bug is currently on master. 

Now that I have tests (and actually an idea, how to test the widget) I will continue to improve the calendar with further PRs. I might also fix the problem in `Date`.

chipsenkbeil:
@seflue what is the actual bug you're solving?

seflue:
It is an edge case with the end of month, which leads to unexpected nil date, which breaks the widget. It is in the date class, but easily to work around.

seflue:
The fix of the actual cause is a bit more complicated - I will make a second PR soon.

seflue:
In #748 I fixed the underlying problem, which is explained in #747. We can merge #748 before this one, then I will also remove the workaround in `Calendar`.

kristijanhusak:
I don't remember why I had both `month` and `date`, but could we just ditch the `month` and use only `date`? @chipsenkbeil would that cause some issues for you?

chipsenkbeil:
Gave a scan over where date is used (only in one file, IIRC) within org roam. I only access `month` as a field from `OrgDate`, not `OrgCalendar`. So I think this is fine.

seflue:
> I don't remember why I had both `month` and `date`, but could we just ditch the `month` and use only `date`? @chipsenkbeil would that cause some issues for you?

That's what I actually did. The month member function is now deriving the month from the date state. Actually I had a hard time thinking over this, when I wrote my first implementation of the time picker - and already tried to get rid of month as an independent state, which was not easy with the original implementation. But with the refactoring of calendar it got reintroduce - meanwhile I had mostly forgotten, what I initially did. When I discovered and investigated the bugs, the deja vus came on after another ... it should now be much more stable.
Actually the ideas of improving the class are stacking up, so expect more PRs here.

seflue:
I rebased the code onto master and removed the quickfix. Now the essence of this PR is the removal of the `month` field in `Calendar` and adding some test for it.
@kristijanhusak Ready for merge from my side.

lua/orgmode/objects/calendar.lua
```diff
@@ -156,25 +158,31 @@ function Calendar:render()
     weekday_row = { 'Sun', 'Mon', 'Tue', 'Wed', 'Thu', 'Fri', 'Sat' }
   end
 
+  local date = self.date or Date.today()
```
kristijanhusak, line 161: Can it happen that we don't have a `self.date` ?

lua/orgmode/objects/calendar.lua
```diff
@@ -46,13 +44,17 @@ function Calendar.new(data)
   this.on_day = data.on_day
   if data.date then
     this.date = data.date
-    this.month = this.date:set({ day = 1 })
   else
-    this.month = Date.today():start_of('month')
+    this.date = Date.today():start_of('month')
```
kristijanhusak, line 48: Why default to start of month?

tests/plenary/object/calendar_spec.lua
```diff
@@ -0,0 +1,87 @@
+local helpers = require('tests.plenary.helpers')
+local Calendar = require('orgmode.objects.calendar')
+local Date = require('orgmode.objects.date')
+
+local close_alle_buffers = function()
```
kristijanhusak, line 5:
```suggestion
local close_all_buffers = function()
```

kristijanhusak (changes requested):
Looks good, just some minor changes.

Also please rebase from master. I found a bug with the calendar not being able to go to previous/next year because the month number overflows, and `days_of` was not able to find the correct number of days.

lua/orgmode/objects/calendar.lua
seflue: This is actually a form of defensive programming and came in as an attempt to fix the bug I initially observed. Removed.

lua/orgmode/objects/calendar.lua
seflue: Also part of the initial quickfix, now obsolete. Removed.

tests/plenary/object/calendar_spec.lua
seflue: Cannot hide my german roots. :smile:

lua/orgmode/objects/calendar.lua
```diff
@@ -46,13 +44,17 @@ function Calendar.new(data)
...
-    this.month = this.date:set({ day = 1 })
   else
-    this.month = Date.today():start_of('month')
+    this.date = Date.today()
   end
   return this
 end
 
+---@return OrgDate
+function Calendar:month()
```
kristijanhusak, line 54: I don't see this being used anywhere. Why we need it?

tests/plenary/object/calendar_spec.lua
```diff
@@ -0,0 +1,87 @@
...
+describe('Calendar', function()
+  after_each(close_all_buffers)
+  it('should open a calendar at plain headline', function()
+    helpers.create_agenda_file({
+      '#+TITLE: Test',
+      '',
+      '* Some Headline',
+    })
+    vim.fn.cursor(3, 1)
+    feed(open_calendar)
```
kristijanhusak, line 26: Let's use the real mapping here via `vim.cmd` like it's done for some of the [other tests](https://github.com/nvim-orgmode/orgmode/blob/c9bf6d8e926c5d6ad7103ee8e7b1e38500b7bfc5/tests/plenary/ui/mappings/headline_spec.lua?plain=1#L265)
```suggestion
    vim.cmd('norm ,oid')
```
That way you can also remove the 3 variables above the describe.

kristijanhusak (commented):
Works great! Just few minor things and it's good to go.

lua/orgmode/objects/calendar.lua
seflue: I first use it to differentiate beween the date state and the information about the current month during the refactoring. But from the latter nothing is left anymore, so it got stale, but I didn't recognize it. Removed.

tests/plenary/object/calendar_spec.lua
seflue: Exchanged it. I still need the feed method to actually simulate keystrokes within the calendar. This didn't work with `vim.cmd`.

seflue:
@kristijanhusak I don't know, why the tests fail, it is unrelated to the code and it doesn't fail, when I run the tests locally. But unfortunately I don't have permissions to rerun the tests.

kristijanhusak (approved):
I can reproduce the test fail on master branch with nightly, so it's unrelated to this. Everything looks good now, merging. Thanks!

PR #746, opened 2024-06-04
feat(archive): set outline path when archiving

andreaswachowski:
This adds the outline path (`ARCHIVE_OLPATH`, i.e. the list of ancestors of the archived headline) to the archive context. In Emacs, see

* org-archive-save-context-info: https://github.com/emacs-mirror/emacs/blob/7f415f7d5e19ea3903500426a0724e49271ac699/lisp/org/org-archive.el#L119
* org-get-outline-path: https://github.com/emacs-mirror/emacs/blob/7f415f7d5e19ea3903500426a0724e49271ac699/lisp/org/org.el#L7464

This PR includes a fix to set the `ARCHIVE_TODO` attribute as well.

I checked with Emacs first and then specified the test to match what Emacs does. Looking forward to any feedback you might have!

### DIFF
--- a/lua/orgmode/capture/init.lua
+++ b/lua/orgmode/capture/init.lua
@@ -275,6 +275,7 @@ function Capture:refile_file_headline_to_archive(headline)
   end
 
   local destination_file = self.files:get(archive_location)
+  local todo_state = headline:get_todo()
 
   local target_line = self:_refile_from_org_file({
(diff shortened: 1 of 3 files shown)


seflue:
@kristijanhusak do you have any concern with this PR? It sounds quite useful.

PR #748, opened 2024-06-05
Fix date arithmetic issues with month

### LINKED ISSUE #747: Date arithmetic does calculate correctly +/- 1 month
### Describe the bug

If I add a month to the 31st of a long month (January, March, Mai ...), the Date overshoots one month. This was exposed by using `:end_of('month')` and `:add({month = 1})` and `:subtract({month = 1})` in Calendar.

### Steps to reproduce

See tests in attached PR.

### Expected behavior

Adding a month to 31. of a month should give me the last day of the next month. This also means correct handling of leap years.

### Emacs functionality

_No response_

### Minimal init.lua

Empty on purpose

_No response_

### OS / Distro

Linux

### Neovim version/commit

0.10

### Additional context

_No response_

seflue:
Fixes #747.


seflue:
@kristijanhusak I fixed also all the type warnings, for which I'll create a second PR when this one is merged.

lua/orgmode/objects/date.lua
```diff
@@ -517,17 +518,45 @@ end
...
+    return 30
+  end
+end
+
+function Date._days_of_february(year)
+  return Date._is_leap_year(year) and 29 or 28
+end
+
+function Date._is_leap_year(year)
+  return year % 400 == 0 or not (year % 100 == 0) and year % 4 == 0
```
kristijanhusak, line 557: Let's add some parenthesis around these so it's easier to read:
 
```suggestion
  return year % 400 == 0 or (year % 100 ~= 0 and year % 4 == 0)
```

lua/orgmode/objects/date.lua
```diff
@@ -481,7 +481,8 @@ function Date:end_of(span)
   end
 
   if span == 'month' then
-    return self:add({ month = 1 }):start_of('month'):adjust('-1d'):end_of('day')
+    local date = os.date('*t', self.timestamp)
+    return self:set({ day = Date._days_of_month(date) }):end_of('day')
```
kristijanhusak, line 485: Use `self` when it is available in the method. We can also move these new methods to be on an instance.
```suggestion
    return self:set({ day = self:_days_of_month(date) }):end_of('day')
```

kristijanhusak (changes requested):
This is a good catch, I wasn't aware of the issue.
Lets just change few things around and it's good to go.

lua/orgmode/objects/date.lua
```diff
@@ -517,17 +518,45 @@ end
...
   opts = opts or {}
+  ---@type table
   local date = os.date('*t', self.timestamp)
   for opt, val in pairs(opts) do
     if opt == 'week' then
       opt = 'day'
       val = val * 7
     end
     date[opt] = date[opt] + val
+    if opt == 'month' then
```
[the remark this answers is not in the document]
seflue: Done. I like the `math.min` version actually a lot better than mine. :+1:

lua/orgmode/objects/date.lua
seflue: Done.

lua/orgmode/objects/date.lua
seflue: I actually tried it with self, but date here is the table from `os.date`, not `OrgDate`. It get's clearer, when I open my PR, where I cleaned up the type warnings.
What I thought about to make it a local function instead. We can then add a public wrapper for OrgDate. But I would prefer to do this with the next PR.

kristijanhusak (approved):
Looks good, thanks!

PR #751, opened 2024-06-10
refactor(date): improve type hints

seflue:
- introduce types for internal date tables
- introduce some helper to encapsulate interaction with os api

### DIFF
--- a/lua/orgmode/objects/date.lua
+++ b/lua/orgmode/objects/date.lua
@@ -1,6 +1,6 @@
 -- TODO
 -- Support diary format and format without short date name
----@type table<string, OrgDateSpan>
+---@type OrgDateSpan
 local spans = { d = 'day', m = 'month', y = 'year', h = 'hour', w = 'week', M = 'min' }
 local config = require('orgmode.config')
 local utils = require('orgmode.utils')
@@ -11,25 +11,32 @@ local time_format = '%H:%M'
 
 ---@alias OrgDateSpan 'minute' | 'hour' | 'day' | 'week' | 'month' | 'year'
 
----@class OrgDate
----@field type string
----@field active boolean
----@field date_only boolean
----@field range OrgRange
+---@alias OsTime number
+
+---@class OsDate
 ---@field day number
+---@field wday string
 ---@field month number
 ---@field year number
 ---@field hour number
 ---@field min number
+---
+---@class OrgDateOpts : OsDate
+---@field active boolean
+---@field date_only boolean
+---@field type string
+---@field range OrgRange
 ---@field timestamp number
 ---@field timestamp_end number
 ---@field is_dst boolean
 ---@field is_date_range_start boolean
 ---@field is_date_range_end boolean
----@field related_date_range OrgDate
----@field dayname string
+---@field related_date_range OrgDate?
+---@field dayname string?
 ---@field adjustments string[]
 ---@field private is_today_date boolean?
+
+---@class OrgDate : OrgDateOpts
 local Date = {
   ---@type fun(this: OrgDate, other: OrgDate): boolean
   __eq = function(this, other)
@@ -56,7 +63,7 @@ local Date = {
 ---@param source table
 ---@param target? table
 ---@param include_sec? boolean
----@return table
+---@return OrgDateOpts
 local function set_date_opts(source, target, include_sec)
   target = target or {}
   for _, field in ipairs({ 'year', 'month', 'day' }) do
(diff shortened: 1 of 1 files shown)


seflue:
@kristijanhusak As promised here the cleaned up type hints in `Date`.
The failing test is unrelated, but when I had a look at it, it seems to be a flaky candidate. I don't think it can be made 100% deterministic, but at least if we cache the current time right after the command (before `vim.wait` we can reduce the probability to fail:

https://github.com/nvim-orgmode/orgmode/blob/ac9a6e0eac2d298a27ab4f59511e60897cb61a08/tests/plenary/ui/mappings/todo_spec.lua#L343-L357

lua/orgmode/objects/date.lua
```diff
@@ -11,25 +11,31 @@ local time_format = '%H:%M'
 
 ---@alias OrgDateSpan 'minute' | 'hour' | 'day' | 'week' | 'month' | 'year'
 
----@class OrgDate
----@field type string
+---@class OsTime
```
kristijanhusak, line 14: This one does not have any fields. Since `os.time()` always returns a timestamp, we can probably just create an alias:

```suggestion
---@alias OsTime number
```

lua/orgmode/objects/date.lua
```diff
@@ -11,25 +11,31 @@ local time_format = '%H:%M'
...
+
+---@class OsDate
+---@field day number?
```
kristijanhusak, line 17: I think all of the fields should be required, at least `os.date('*t', timestamp)` always returns all of these.
Unless you did it because of some existing logic that is not handling it, which we can discuss.

lua/orgmode/objects/date.lua
```diff
@@ -53,10 +59,13 @@ local Date = {
   end,
 }
 
+---@alias DateTbl { year: number, month: number, day: number, hour: number, min: number }
```
kristijanhusak, line 62: Let's prefix all of the types with `Org` just to avoid potential clash with other plugins.

lua/orgmode/objects/date.lua
```diff
@@ -194,8 +229,8 @@ end
 ---@param data? table
 ---@return OrgDate
 local function today(data)
-  local date = os.date('*t', os.time()) --[[@as osdate]]
-  local opts = vim.tbl_deep_extend('force', date, data or {})
+  local date = os_date(os.time())
```
kristijanhusak, line 232: Should we use `os_time` here?

```suggestion
  local date = os_date(os_time())
```

lua/orgmode/objects/date.lua
```diff
@@ -490,12 +525,16 @@ end
...
+  ---@type table
+  ---@diagnostic disable-next-line: assign-type-mismatch
   local date = os.date('*t', self.timestamp)
   return utils.convert_to_isoweekday(date.wday)
 end
 
 ---@return number
 function Date:get_weekday()
+  ---@type table
+  ---@diagnostic disable-next-line: assign-type-mismatch
```
kristijanhusak, line 537: We can add `wday` to `OsDate` type and use `os_date` here.

lua/orgmode/objects/date.lua
```diff
@@ -858,8 +898,9 @@ function Date:get_repeater()
   return nil
 end
 
+---@return OrgDate
 function Date:set_todays_date()
-  local time = os.date('*t', os.time())
+  local time = os_date(os.time())
```
kristijanhusak, line 903: Same as other comment, should we use `os_time()` here?

kristijanhusak (changes requested):
Generally looks ok. I just skimmed through it. See if we need to address some of these comments and then I'll check it thoroughly.

lua/orgmode/objects/date.lua
seflue: Actually I made the field optional to fix the warnings when assigning the result of the `tonumber` functions in `parse_datetime`. The other option would be to provide a default, if tonumber returns nil.

https://github.com/nvim-orgmode/orgmode/blob/ac9a6e0eac2d298a27ab4f59511e60897cb61a08/lua/orgmode/objects/date.lua#L150-L174

I mean, you have a point - because OrgDate transitively inherits from OsDate (as it is currently implemented) it inherits also the optionals, which we might not want to. Should I fill in some sane defaults in the parsing function and if so, which one should we choose? We could just take the current date or zero. For the year, I think the current year makes more sense, same might be the case for month and even day. What do you think?

kristijanhusak: You can default it to `0`, because at this point the value should already be validated and we should have all the numbers.

lua/orgmode/objects/date.lua
seflue: No, we want to get the current time. The argument of `os.time` is optional. We use `os_time` to parse a date and get rid of the type warnings. Because we feed the result of `os.time` in our warning-swallowing `os_date` this line is fine as it is.

lua/orgmode/objects/date.lua
seflue: Done.

lua/orgmode/objects/date.lua
seflue: Actually in the end I didn't used these aliases anymore. Removed.

lua/orgmode/objects/date.lua
seflue: Done.

lua/orgmode/objects/date.lua
seflue: Done.

lua/orgmode/objects/date.lua
seflue: Done.

kristijanhusak (approved):
Ok this looks good now. I'll do some minor changes after merging, but I don't want to bother you with those nits. Thanks!

PR #753, opened 2024-06-10
Fix subtree move (fixes #752)

### LINKED ISSUE #752: Cursor position after "Move subtree" doesn't allow for convenient repetition
### Describe the bug

While playing around with [Hydra.nvim](https://github.com/anuvyklack/hydra.nvim) to create a convenient way to reorder headlines and subtrees, I discovered, that the cursor position after moving the subtree up or down doesn't allow for convenient repetition. If the current tree has no child items, everything works fine, but as soon as it has child items, the behavior of placing the cursor at the last line after the movement (which just happens automatically, if we don't take additional care) ends in "Cannot move past superior level" with the next attempt to move the subtree, even the move would be valid.

I propose to set the cursor to the heading which get's moved around as soon as the command ends.

### Steps to reproduce

Example org file:

``` org
* Heading
** Subheading1
some content
** Subheading2
** Subheading3
*** Child heading
more content
```

Try to move `Subheading3` up twice.

### Expected behavior

Command should be repeatable without any quirk.

### Emacs functionality

I don't remember any issues with this in Emacs

### Minimal init.lua

Nothing of relevance here.

_No response_

### OS / Distro

Linux

### Neovim version/commit

0.10

### Additional context

I wile provide a PR with a fix.

seflue:
Allows to move subtrees around smoothly (e.g. using [Hydra.nvim](https://github.com/anuvyklack/hydra.nvim), see #752). It also fixes the foldings after changing the subtree and preserves the folding state in the current subtree when moving it up or down.

### DIFF
--- a/lua/orgmode/org/mappings.lua
+++ b/lua/orgmode/org/mappings.lua
@@ -762,7 +762,10 @@ function OrgMappings:move_subtree_up()
     return utils.echo_warning('Cannot move past superior level.')
   end
   local range = item:get_range()
-  vim.cmd(string.format(':%d,%dmove %d', range.start_line, range.end_line, prev_headline:get_range().start_line - 1))
+  local target_line = prev_headline:get_range().start_line - 1
+  vim.cmd(string.format(':%d,%dmove %d', range.start_line, range.end_line, target_line))
+  local pos = vim.fn.getcurpos()
+  vim.fn.cursor(prev_headline:get_range().start_line, pos[2])
 end
 
 function OrgMappings:move_subtree_down()
@@ -772,7 +775,10 @@ function OrgMappings:move_subtree_down()
     return utils.echo_warning('Cannot move past superior level.')
   end
   local range = item:get_range()
-  vim.cmd(string.format(':%d,%dmove %d', range.start_line, range.end_line, next_headline:get_range().end_line))
+  local target_line = next_headline:get_range().end_line
+  vim.cmd(string.format(':%d,%dmove %d', range.start_line, range.end_line, target_line))
+  local pos = vim.fn.getcurpos()
+  vim.fn.cursor(target_line + range.start_line - range.end_line, pos[2])
 end
 
 function OrgMappings:show_help(type)
--- a/lua/orgmode/org/mappings.lua
+++ b/lua/orgmode/org/mappings.lua
@@ -763,9 +763,13 @@ function OrgMappings:move_subtree_up()
   end
   local range = item:get_range()
   local target_line = prev_headline:get_range().start_line - 1
+  local foldclosed = vim.fn.foldclosed('.')
   vim.cmd(string.format(':%d,%dmove %d', range.start_line, range.end_line, target_line))
   local pos = vim.fn.getcurpos()
-  vim.fn.cursor(prev_headline:get_range().start_line, pos[2])
+  vim.fn.cursor(target_line + 1, pos[2])
+  if foldclosed > -1 and vim.fn.foldclosed('.') == -1 then
+    vim.cmd([[norm!zc]])
+  end
 end
 
 function OrgMappings:move_subtree_down()
@@ -776,9 +780,13 @@ function OrgMappings:move_subtree_down()
   end
   local range = item:get_range()
   local target_line = next_headline:get_range().end_line
+  local foldclosed = vim.fn.foldclosed('.')
   vim.cmd(string.format(':%d,%dmove %d', range.start_line, range.end_line, target_line))
   local pos = vim.fn.getcurpos()
   vim.fn.cursor(target_line + range.start_line - range.end_line, pos[2])
+  if foldclosed > -1 and vim.fn.foldclosed('.') == -1 then
+    vim.cmd([[norm!zc]])
+  end
 end
 
 function OrgMappings:show_help(type)
--- a/tests/plenary/ui/mappings/headline_spec.lua
+++ b/tests/plenary/ui/mappings/headline_spec.lua
@@ -301,7 +301,7 @@ describe('Heading mappings', function()
     }, vim.api.nvim_buf_get_lines(0, 0, 2, false))
   end)
 
-  it('should move subtree up (org_move_subtree_up)', function()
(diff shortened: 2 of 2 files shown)


kristijanhusak:
I think there is a simpler and shorter solution to this. We can just jump to the target line after moving.
This patch works ok for me:
```diff
diff --git a/lua/orgmode/org/mappings.lua b/lua/orgmode/org/mappings.lua
index 41d7c9f..7272f52 100644
--- a/lua/orgmode/org/mappings.lua
+++ b/lua/orgmode/org/mappings.lua
@@ -763,6 +763,7 @@ function OrgMappings:move_subtree_up()
   end
   local range = item:get_range()
   vim.cmd(string.format(':%d,%dmove %d', range.start_line, range.end_line, prev_headline:get_range().start_line - 1))
+  vim.fn.cursor(prev_headline:get_range().start_line, 1)
 end
 
 function OrgMappings:move_subtree_down()
@@ -773,6 +774,7 @@ function OrgMappings:move_subtree_down()
   end
   local range = item:get_range()
   vim.cmd(string.format(':%d,%dmove %d', range.start_line, range.end_line, next_headline:get_range().end_line))
+  vim.fn.cursor(next_headline:get_range().end_line - 2, 1)
 end
 
 function OrgMappings:show_help(type)

```

I'd prefer not to touch the folds with `zx` because that will recalculate everything and close other stuff that user might not want to be closed.

seflue:
Would it be ok for you, if I add a configuration variable, which is false by default, but if set to true, will fix the foldings? Because I find myself pressing `zx` all the time and it would make my workflow much more convenient, if I Orgmode would do that for me.

kristijanhusak:
We can fire an event similar to what we do for promote/demote (example:  https://github.com/nvim-orgmode/orgmode/blob/e91b94cf29b24e62558218343b464d2a1ef44c03/lua/orgmode/org/mappings.lua?plain=1#L502), and then you can listen for that in your config to do whatever you want.
You can call the events `HeadlineMovedUp` and `HeadlineMovedDown`

seflue:
> HeadlineDemoted

Two questions:

First, should we distinguish between up and down? Is there a difference or is just "HeadlineMoved" sufficient? My feeling is, that one event is enough, because it does not make much of a difference from perspective of a user. At least I currently cannot imagine a use case, where it makes a difference.

Second, is there some example code, how the events can be used from the config? A fast search over the help file didn't get any results.

For now I will just remove the fold fixes and would like to get the PR merged as it is and would open another to add the events.

seflue:
> I think there is a simpler and shorter solution to this. We can just jump to the target line after moving. This patch works ok for me:
> 
> ```diff
> diff --git a/lua/orgmode/org/mappings.lua b/lua/orgmode/org/mappings.lua
> index 41d7c9f..7272f52 100644
> --- a/lua/orgmode/org/mappings.lua
> +++ b/lua/orgmode/org/mappings.lua
> @@ -763,6 +763,7 @@ function OrgMappings:move_subtree_up()
>    end
>    local range = item:get_range()
>    vim.cmd(string.format(':%d,%dmove %d', range.start_line, range.end_line, prev_headline:get_range().start_line - 1))
> +  vim.fn.cursor(prev_headline:get_range().start_line, 1)
>  end
>  
>  function OrgMappings:move_subtree_down()
> @@ -773,6 +774,7 @@ function OrgMappings:move_subtree_down()
>    end
>    local range = item:get_range()
>    vim.cmd(string.format(':%d,%dmove %d', range.start_line, range.end_line, next_headline:get_range().end_line))
> +  vim.fn.cursor(next_headline:get_range().end_line - 2, 1)
>  end
>  
>  function OrgMappings:show_help(type)
> ```
> 
> I'd prefer not to touch the folds with `zx` because that will recalculate everything and close other stuff that user might not want to be closed.



> ```diff
> vim.fn.cursor(prev_headline:get_range().start_line, 1)
> ```

Your patch does not work for me, at least not for moving the headline down. I backed everything with tests, with my code they succeed, with yours they fail. I simplified the code anyway though. Hope that works for you.

kristijanhusak:
> First, should we distinguish between up and down? Is there a difference or is just "HeadlineMoved" sufficient?

You're right, one event is enough. Just provide an argument to the event which way it went.

> Second, is there some example code, how the events can be used from the config? A fast search over the help file didn't get any results.

There isn't since I wanted to keep it internal, and expose it only through the api. For now, you can do something like this (Example with TodoChanged event):

```lua
  local eventManager = require('orgmode.events')
  local eventTypes = require('orgmode.events.types')

  eventManager.listen(eventTypes.TodoChanged, function()
    print('todo changed')
  end)
```

> Your patch does not work for me, at least not for moving the headline down

I'll take another look. I tested with the example from the issue you reported.

seflue:
> > First, should we distinguish between up and down? Is there a difference or is just "HeadlineMoved" sufficient?
> 
> You're right, one event is enough. Just provide an argument to the event which way it went.
> 
> > Second, is there some example code, how the events can be used from the config? A fast search over the help file didn't get any results.
> 
> There isn't since I wanted to keep it internal, and expose it only through the api. For now, you can do something like this (Example with TodoChanged event):
> 
> ```lua
>   local eventManager = require('orgmode.events')
>   local eventTypes = require('orgmode.events.types')
> 
>   eventManager.listen(eventTypes.TodoChanged, function()
>     print('todo changed')
>   end)
> ```

Thanks for the insight, will try that out - but as I said, in an upcomping PR.

> > Your patch does not work for me, at least not for moving the headline down
> 
> I'll take another look. I tested with the example from the issue you reported.

I mean, it's quite simple math: If you move a subtree of arbitrary length in the file down, then this arbitrary length has to be somehow included in the calculation of the new cursor position, because moving the current subtree down does basically mean, pulling the next subtree (and the whole subsequent lines) this arbitrary length up. So to calculate the new cursor position _after_ the move from a line number _before_ the move has to subtract the number of lines of our moving subtree.
Determining the cursor position after moving the subtree up is much simpler though, because the target line has the same line number before and after the move.

kristijanhusak (approved):
Ok yeah I didn't test all the cases. Looks good, thanks!

Issue #727, opened 2024-05-17
Error (freezes nvim) when opening markdown with code block name in different case (e.g. "Python", "JSON", "R"): no such language: R


milanglacier:
### Describe the bug

This is a macOS only bug, I cannot reproduce it in linux. 

When opening a markdown file that contains code block with same name with the treesitter parser but have different case (e.g. `python` vs `Python`, `r` vs `R`, `Json` vs `json`), you will get the following error:

```
stack traceback:
        [C]: in function '_ts_parse_query'
        ...m/0.10.0/share/nvim/runtime/lua/vim/treesitter/query.lua:252: in function 'fn'
        ...ovim/0.10.0/share/nvim/runtime/lua/vim/func/_memoize.lua:58: in function 'fn'
        ...ovim/0.10.0/share/nvim/runtime/lua/vim/func/_memoize.lua:58: in function 'get'
        ...0/share/nvim/runtime/lua/vim/treesitter/languagetree.lua:128: in function 'new'
        ...0/share/nvim/runtime/lua/vim/treesitter/languagetree.lua:511: in function 'add_child'
        ...0/share/nvim/runtime/lua/vim/treesitter/languagetree.lua:395: in function '_add_injections'
        ...0/share/nvim/runtime/lua/vim/treesitter/languagetree.lua:447: in function 'parse'
        ....0/share/nvim/runtime/lua/vim/treesitter/highlighter.lua:397: in function <....0/share/nvim/runtime/lua/vim/treesitter/highlighter.lua:392>
Error in decoration provider treesitter/highlighter.win:
Error executing lua: ...m/0.10.0/share/nvim/runtime/lua/vim/treesitter/query.lua:252: no such language: R
```

This error is actually the same error as in nvim-treesitter: https://github.com/nvim-treesitter/nvim-treesitter/issues/6642

### Steps to reproduce

1. open the following example org mode file, say : hello.org

2. Open it in neovim, get billions of errors so that you can't do anything with neovim.

https://pastebin.com/sSLFxXBZ

### Expected behavior

1. nvim should open this file normally.

### Minimal init.lua

```lua

local plugins = {
	orgmode = "https://github.com/nvim-orgmode/orgmode",
	ts = "https://github.com/nvim-treesitter/nvim-treesitter",
}

for name, url in pairs(plugins) do
	local install_path = "/tmp/nvim/site/" .. name
	if vim.fn.isdirectory(install_path) == 0 then
[...]

PR #757, opened 2024-06-14
fix: fix file type detection in `org-edit-special` on macOS.

milanglacier:
There exists the same issue as in https://github.com/nvim-orgmode/orgmode/issues/727 for `org edit special`.

This commit uses the same logic as the one to handle the case in highlight injection to solve the problem for `edit-special`.


kristijanhusak (approved):
Thanks!

PR #758, opened 2024-06-14
fix: schedule write action for edit-special.

milanglacier:
The action of writing to the original buffer need to be schedule wrapped. otherwise, on the first time writing the org buffer, it will issue an `"E315: ml_get: invalid lnum: 87"` error. I guess this might be some race conditions? Anyway schedule this call makes this error disappear.

### DIFF
--- a/lua/orgmode/objects/edit_special/types/src.lua
+++ b/lua/orgmode/objects/edit_special/types/src.lua
@@ -129,7 +129,9 @@ function EditSpecialSrc:write(ctx)
   local new_content = vim.api.nvim_buf_get_lines(ctx.bufnr, 0, -1, false)
   new_content = self:_update_content('add', ctx.start_extmark_pos[1], new_content)
 
-  vim.api.nvim_buf_set_lines(ctx.org_bufnr, content_start, content_end, false, new_content)
+  vim.schedule(function()
+    vim.api.nvim_buf_set_lines(ctx.org_bufnr, content_start, content_end, false, new_content)
+  end)
 
   self.file:reload()
 


milanglacier:
I see these tests are failed. However, if works fine on my ends. I think this might because `vim.schedule` makes the write action async and the tests are assumed it is synchronized?

kristijanhusak:
Can you give me the steps to reproduce the issue? I can't reproduce it so can't really help.

milanglacier:
This is a problem that's are really hard to reproduce.

The same config can yield the "E315: ml_get: invalid lnum: 1" on macos, but not on linux. And I cannot create a minimal config either. May need to spend more time on this to find what's triggering the issue.

Another interesting thing is that: `:h E315` shows that"

```
							*E315*  >
  ml_get: invalid lnum: {number}

This is an internal Vim error.  Please try to find out how it can be
reproduced, and submit a |bug-report|.
```

So this might point to a C source code error.

xav-ie:
I was encountering this error on task creation until I set `org_agenda_files` and `org_default_notes_file`. I think it was because this plugin was trying to create refile.org in a location it could not. I was also using noice.nvim and, for some reason, it just spawns an empty confirm window when I tried to make a new agenda task. Confirming/closing the empty window showed this error, broke noice to no longer work for the rest of the session, and failed to create the task.

It was not until I set these two variables the error went away.

Just thought I should comment here if someone else finds this error, too!

kristijanhusak:
Closing this for now. If the issue happens again, reopen and we'll take a look.

Issue #20, opened 2021-06-30
Improve highlight group links for done state


akinsho:
**Describe the bug**

Currently the done state takes its highlighting from `diffAdd` with some fallbacks around the `Diff` highlight group. I guess this is because these are usually green across most colorschemes. The issue with this is that it isn't dependably a high enough contrast of green. In my case the green is quite muted so that it's easier to make out the added text, but this means that the green as a text highlight rather than background is hard to read.

<img width="396" alt="Screen Shot 2021-06-30 at 10 55 43" src="https://user-images.githubusercontent.com/22454918/123949399-6d7c8480-d99a-11eb-882c-422edeaad08f.png">

**Expected behavior**
A more dependably green highlight might be `String` or tbh if the intent is that the colour should be green then you can alternatively use `guifg={Green|LightGreen}` etc which I believe also works for cterm. That way the colour will be green regardless of the users colorscheme.

I guess this could apply to in progress states as well. It might be that your intention is not to aim for a particular colour at all in which case an alternative or additional solution could be to surface the highlight groups to be overriden. Generally I'm fine with how most colours appear so am not looking to go down a highlight altering rabbit hole, so this is less about surfacing a bunch of highlights and more about a more predictable default.

kristijanhusak:
Yeah, you're right. I wanted to reuse as much colors as possible from the colorscheme with some assumptions that can be wrong sometimes (obviously). I didn't want to use these strict colors because they can be really high in contrast for some colorschemes (Red on dark colorscheme looks really bad for example).

I think there are 2 solutions to this:

1. Introduce a configuration option that would allow defining highlights group that will be used for all of these types
2. Make builtin hl groups from these parsed colors and just `default link` them to the builtin ones. Then user can just do `hi default link OrgTODO Error` or whatever they want to link it to.
3. Implement [orgmode's todo keyword faces](https://orgmode.org/manual/Faces-for-TODO-keywords.html) and just allow users to configure all of these by themselves.

Temporary workaround (which you are probably already aware of) is to override these manually for now:
```vim
hi OrgTODO guifg=Red
hi OrgDONE guifg=Green
hi OrgAgendaDeadline guifg=Red
hi OrgAgendaScheduled guifg=Green
hi OrgAgendaScheduledPast guifg=Orange
```

akinsho:
[...]

PR #768, opened 2024-06-30
Extend API to get links to headline and org file

seflue:
In Telescope-orgmode I want to support id-based links when `org_id_link_to_org_use_id` is set. I already published the code to a [draft PR #20](https://github.com/nvim-orgmode/telescope-orgmode.nvim/pull/20).


lua/orgmode/api/init.lua
```diff
@@ -101,7 +100,67 @@ function OrgApi.refile(opts)
...
+--- file:<filepath>::*title
+---
+--- The result is meant to be used as link_location for OrgApi.insert_link.
+--- @param headline OrgApiHeadline
+--- @return string
+function OrgApi.get_link_to_headline(headline)
+  local filename = headline.file.filename
+  local bufnr = vim.fn.bufnr(filename)
+
+  if bufnr == -1 then
```
kristijanhusak, line 118: Why do we need to do this if there is no buffer?

seflue: Jep, actually this is the crucial part (and took me some time to figure out, reading through the agenda code). 

The file we want to target a link to, is very likely not open. If we use `org_id_link_to_org_use_id = true`, there might not be an ID yet. But creating this id needs an edit of the file, which is done by the `id_get_or_create` methods under the hood. And this needs to open the file automatically in the background and close it again, if it is not already open.

Let me clarify the motivation behind this API extension:
When I use telescope-orgmode to create links, I can access the whole index of files build up by orgmode on startup. This allows for very convenient creation of links and cross-references using fuzzy finding. It is not very convenient, if I would have to open the files first, before being able to set the link.
The manual workflow for ID links works actually that way, I have to open the target file, press `org-store-link` (which also creates the id), navigate to my file, where the link originates and press `org-insert-link`. My telescope plugin makes this workflow much smoother, because I can just filter for the target headline without open the file, press enter and, boom, I have the link and the ID.

seflue: I think, we should keep it as it is for know. I expect to discover more requirements while I am implementing more features and playing around with we have. If it doesn't bother you, I would adapt when need and have better knowledge.

kristijanhusak: I just figured out I didn't phrase my question correctly.

What I wanted to ask is:
Would it work for you to move these to api/file and api/headline? All of the file or headline specific api methods live in those file, so I'd prefer to keep it consistent.

If it doesn't work for you, can you clarify why?

seflue: All three methods are used for link creation and are meant to be used together. I might need to understand better, what your primary criterion is for where to put a method and also what the main purpose of the overall API is.
I don't actually have a big problem of making these methods member functions of `OrgApiFile` and `OrgApiHeadline`.

kristijanhusak: >  I might need to understand better, what your primary criterion is for where to put a method

If a function is accepting an instance of something, I always prefer to have that function on the instance itself. Of course that's not always possible, but when it is, I prefer to take that path.
For example, if we would follow this way for example for `get_property`, we would have something like this:

```lua
orgmode.api.get_property(headline, 'id')
```

But this is much cleaner IMO:

```lua
headline:get_property('id')
```

seflue: Yes, was already convienced, when I realized, that the argument of the original functions is actually the self argument in `orgmode.api.headline` and `orgmode.api.file`. . :smile: 
Now that I refactored the code I am even more convienced, because it cleaned also things up on the client side.

What do you think, are we ready to merge?

lua/orgmode/org/hyperlinks/init.lua
```diff
@@ -174,16 +174,32 @@ end
...
 
-  if config.org_id_link_to_org_use_id and id then
-    return ('id:%s::*%s'):format(id, title)
+  path = path or utils.current_file_path()
+  return ('file:%s::*%s'):format(path, title)
+end
+
+---@param file OrgFile
+---@param path? string
+function Hyperlinks.get_link_to_file(file, path)
```
kristijanhusak, line 192: Is a file path with the title valid format for a file link? 

From the examples here https://orgmode.org/guide/Hyperlinks.html I don't see any of those.

I think for link to file it is enough to either be an id, or just `file:/path/to/file.org` if there is no id.

seflue: It is not, and it's also not treated such. I am piggy-backing here on the "title" format to sneak a description into insert_link to get it prefilled.
This is something, which actually needs to get cleaned up a bit more properly:
`get_link` should actually return even better an api equivalent of `Url`. And `insert_link` should take this. But this would need some more decent refactoring effort which touches a lot of places (think also of `store_link`) so I would prefer to defer this to a separate PR.

Right now I am quite happy, how Telescope-orgmode works although the API is not perfect yet.

kristijanhusak: Ok, we can leave it for the time being, but later once we do some refactoring, I'd prefer to remove it.
It can also cause side effects in cases where `#+TITLE` or the file name is the same as some other headline value. In that case, the link would jump to the headline.

kristijanhusak (approved):

PR #773, opened 2024-07-15
bug(core) Ignore invisible headings

SlayerOfTheBad:
Previously, `next-visible-heading`, moved to the next heading, regardless of its visibility.


kristijanhusak (approved):
Looks good, thanks!

PR #790, opened 2024-08-04
[WIP][RFC] Hyperlink Refactor

SlayerOfTheBad:
I noticed that link abbreviations weren't supported, and wanted to make a PR to do that.
After looking at the code, I quickly started understanding why they weren't supported.
So I set to work refactoring the entire Hyperlink section.

This is a draft because
1. Unittests still need to be written
2. Other unit tests still need to be modified
3. Type annotations still need to be put in
4. Documentation for new features needs to be written

However, I'm going on vacation soon, so I thought I'd present it so this can be reviewed whilst I'm away.
Also, if this is not something that's desired, I don't have to bother writing all the Unit tests.

I've tried to keep the Api stable, but I do know that at least `org-roam.nvim` does some calls to `link.url` which doesn't exist anymore.
I am willing to make a PR for `org-roam.nvim` to resolve those issues.

As for what this pull request changes feature-wise:
1. Improved autocompletion for links on `org-insert-link`
2. Support for [Link Abbreviations](https://orgmode.org/org.pdf#Link%20Abbreviations)
3. Support for [Custom Hyperlink Types](https://orgmode.org/org.pdf#Adding%20Hyperlink%20Types)

In addition to being much more maintainable and easier to extend in future.

PR #793, opened 2024-08-13
[WIP][RFC] Hyperlink Refactor - Registry Version

SlayerOfTheBad:
This pull request achieves most of the same goals as https://github.com/nvim-orgmode/orgmode/pull/790, but is structured in a similar way to `orgmode.org.autocompletion`.

This PR superseeds https://github.com/nvim-orgmode/orgmode/pull/790.

TODO: unit tests

### DIFF
--- a/lua/orgmode/init.lua
+++ b/lua/orgmode/init.lua
@@ -20,6 +20,7 @@ local auto_instance_keys = {
 ---@field capture OrgCapture
 ---@field clock OrgClock
 ---@field completion OrgCompletion
+---@field links OrgLinkHandlerRegistry
 ---@field org_mappings OrgMappings
 ---@field notifications OrgNotifications
 local Org = {}
@@ -66,6 +67,9 @@ function Org:init()
     files = self.files,
   })
   self.completion = require('orgmode.org.autocompletion'):new({ files = self.files })
+  self.links = require('orgmode.org.links'):new({
+    files = self.files,
+  })
   self.statusline_debounced = require('orgmode.utils').debounce('statusline', function()
     return self.clock:get_statusline()
   end, 300)
--- a/lua/orgmode/init.lua
+++ b/lua/orgmode/init.lua
@@ -20,7 +20,7 @@ local auto_instance_keys = {
 ---@field capture OrgCapture
 ---@field clock OrgClock
 ---@field completion OrgCompletion
----@field links OrgLinkHandlerRegistry
+---@field links OrgLinks
 ---@field org_mappings OrgMappings
 ---@field notifications OrgNotifications
 local Org = {}
--- a/lua/orgmode/org/links/handlers/_internal/custom_id.lua
+++ b/lua/orgmode/org/links/handlers/_internal/custom_id.lua
@@ -30,7 +30,7 @@ function CustomId:follow()
     return utils.echo_warning(('Could not find custom ID "%s".'):format(self.custom_id))
   end
 
-  self.goto_oneof(headlines)
+  utils.goto_oneof(headlines)
 end
 
 function CustomId:insert_description()
@@ -44,7 +44,7 @@ function CustomId:complete(lead, context)
   local completions = {}
   for _, headline in pairs(headlines) do
     local id = headline:get_property('CUSTOM_ID')
-    table.insert(completions, self:new(id):__tostring())
+    table.insert(completions, tostring(self:new(id)))
   end
 
   return completions
--- a/lua/orgmode/org/links/handlers/_internal/custom_id.lua
+++ b/lua/orgmode/org/links/handlers/_internal/custom_id.lua
@@ -1,0 +1,53 @@
+local utils = require('orgmode.utils')
+local Org = require('orgmode')
+local Internal = require('orgmode.org.hyperlinks.builtin.internal')
+
+---@class OrgLinkHandlerCustomId:OrgLinkHandlerInternal
+local CustomId = Internal:new()
+
+function CustomId:new(custom_id)
+  ---@class OrgLinkHandlerCustomId
+  local this = Internal:new()
+  this.custom_id = custom_id
+  setmetatable(this, self)
+  self.__index = self
+  return this
+end
+
+---@param input string
+function CustomId.parse(input)
+  return CustomId:new(input)
+end
+
+function CustomId:__tostring()
+  return string.format('#%s', self.custom_id)
+end
+
+function CustomId:follow()
+  local headlines = Org.files:get_current_file():find_headlines_with_property_matching('custom_id', self.custom_id)
+
+  if #headlines == 0 then
+    return utils.echo_warning(('Could not find custom ID "%s".'):format(self.custom_id))
+  end
+
+  self.goto_oneof(headlines)
+end
+
+function CustomId:insert_description()
+  return self.custom_id
+end
+
+function CustomId:complete(lead, context)
+  local file = self.get_file_from_context(context)
+  local headlines = file:find_headlines_with_property_matching('CUSTOM_ID', lead)
+
+  local completions = {}
+  for _, headline in pairs(headlines) do
+    local id = headline:get_property('CUSTOM_ID')
+    table.insert(completions, self:new(id):__tostring())
+  end
+
+  return completions
+end
+
+return CustomId
--- a/lua/orgmode/org/links/handlers/_internal/headline.lua
+++ b/lua/orgmode/org/links/handlers/_internal/headline.lua
@@ -1,0 +1,68 @@
+local utils = require('orgmode.utils')
+local Org = require('orgmode')
+local Internal = require('orgmode.org.hyperlinks.builtin.internal')
+local Id = require('orgmode.org.hyperlinks.builtin.id')
+
+---@class OrgLinkHandlerHeadline:OrgLinkHandlerInternal
+local Headline = Internal:new()
+
+function Headline:new(headline)
+  ---@class OrgLinkHandlerHeadline
+  local this = Internal:new()
+  this.headline = headline
+  setmetatable(this, self)
+  self.__index = self
+  return this
+end
+
+---@param input string
+function Headline.parse(input)
+  return Headline:new(input)
+end
+
+function Headline:__tostring()
+  return string.format('*%s', self.headline)
+end
+
+function Headline:follow()
+  local headlines = Org.files:get_current_file():find_headlines_by_title(self.headline)
+
+  if #headlines == 0 then
+    return utils.echo_warning(('Could not find headline "%s".'):format(self.headline))
+  end
+
+  self.goto_oneof(headlines)
+end
+
+function Headline:resolve()
+  local headlines = Org.files:get_current_file():find_headlines_by_title(self.headline)
+
+  if #headlines == 0 then
+    return self
+  end
+
+  local id = headlines[1]:get_property('id')
+  if not id then
+    return self
+  end
+
+  return Id:new(id):resolve()
+end
+
+function Headline:insert_description()
+  return self.headline
+end
+
+function Headline:complete(lead, context)
+  local file = self.get_file_from_context(context)
+  local headlines = file:find_headlines_by_title(lead)
+
+  local completions = {}
+  for _, headline in pairs(headlines) do
+    table.insert(completions, Headline:new(headline:get_title()):__tostring())
+  end
+
+  return completions
+end
+
+return Headline
--- a/lua/orgmode/org/links/handlers/_internal/headline.lua
+++ b/lua/orgmode/org/links/handlers/_internal/headline.lua
@@ -31,22 +31,7 @@ function Headline:follow()
     return utils.echo_warning(('Could not find headline "%s".'):format(self.headline))
   end
 
-  self.goto_oneof(headlines)
-end
-
-function Headline:resolve()
-  local headlines = Org.files:get_current_file():find_headlines_by_title(self.headline)
-
-  if #headlines == 0 then
-    return self
-  end
-
-  local id = headlines[1]:get_property('id')
-  if not id then
-    return self
-  end
-
-  return Id:new(id):resolve()
+  utils.goto_oneof(headlines)
 end
 
 function Headline:insert_description()
@@ -59,7 +44,7 @@ function Headline:complete(lead, context)
 
   local completions = {}
   for _, headline in pairs(headlines) do
-    table.insert(completions, Headline:new(headline:get_title()):__tostring())
+    table.insert(completions, tostring(Headline:new(headline:get_title())))
   end
 
   return completions
--- a/lua/orgmode/org/links/handlers/file.lua
+++ b/lua/orgmode/org/links/handlers/file.lua
@@ -4,41 +4,23 @@ local Id = require('orgmode.org.links.handlers.id')
 local Internal = require('orgmode.org.links.handlers.internal')
 
 ---@class OrgLinkHandlerFile:OrgLinkHandler
----@field new fun(self: OrgLinkHandlerFile, path: string, target: OrgLinkHandlerInternal | nil, prefix: boolean | nil): OrgLinkHandlerFile
+---@field new fun(self: OrgLinkHandlerFile, path: string, target: OrgLinkHandlerInternal | nil, prefix: boolean | nil, files: OrgFile[]): OrgLinkHandlerFile
 ---@field parse fun(link: string, prefix: boolean | nil): OrgLinkHandlerFile | nil
 ---@field path string
 ---@field skip_prefix boolean
 ---@field target OrgLinkHandlerInternal | nil
 local File = Link:new('file')
 
-function File:new(path, target, skip_prefix)
+function File:new(path, target, skip_prefix, files)
   ---@class OrgLinkHandlerFile
   local this = Link:new()
   this.skip_prefix = skip_prefix or false
   this.path = path
   this.target = target
+  this.files = files
   setmetatable(this, self)
   self.__index = self
   return this
-end
-
-function File.parse(input, skip_prefix)
-  if input == nil or #input == 0 then
-    return nil
-  end
-  local deliniator_start, deliniator_stop = input:find('::')
-
-  ---@type OrgLinkHandlerInternal | nil
-  local target = nil
-  local path = input
-
-  if deliniator_start then
-    ---@class OrgLinkHandlerInternal | nil
-    target = Internal.parse(input:sub(deliniator_stop + 1), true)
-    path = input:sub(0, deliniator_start - 1)
-  end
-
-  return File:new(path, target, skip_prefix)
 end
 
 -- TODO make protocol prefix optional. Based on what?
@@ -65,9 +47,8 @@ function File:follow()
   end
 end
 
-local function autocompletions_filenames(lead)
-  local Org = require('orgmode')
-  local filenames = Org.files:filenames()
+function File:_autocompletions_filenames(lead)
+  local filenames = self.files:filenames()
 
   local matches = {}
   for _, f in ipairs(filenames) do
@@ -80,24 +61,6 @@ local function autocompletions_filenames(lead)
 
   print(vim.inspect(matches))
   return matches
-end
-
-function File:resolve()
-  local Org = require('orgmode')
-  local path = fs.get_real_path(self.path)
-  if not path then
-    return self
-  end
-  local file = Org.files:get(path)
-  if not file then
-    return self
-  end
-  local id = file:get_property('id')
-  if not id then
-    return self
-  end
-
-  return Id:new(id, self.target):resolve()
 end
 
 function File:insert_description()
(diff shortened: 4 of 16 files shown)


lua/orgmode/init.lua
```diff
@@ -20,6 +20,7 @@ local auto_instance_keys = {
 ---@field capture OrgCapture
 ---@field clock OrgClock
 ---@field completion OrgCompletion
+---@field links OrgLinkHandlerRegistry
```
kristijanhusak, line 23: Let's name this just `OrgLinks`
```suggestion
---@field links OrgLinks
```

lua/orgmode/org/links/handlers/_internal/custom_id.lua
```diff
@@ -0,0 +1,53 @@
...
+function CustomId.parse(input)
+  return CustomId:new(input)
+end
+
+function CustomId:__tostring()
+  return string.format('#%s', self.custom_id)
+end
+
+function CustomId:follow()
+  local headlines = Org.files:get_current_file():find_headlines_with_property_matching('custom_id', self.custom_id)
```
kristijanhusak, line 27: We should pass down `files` from the top through constructor instead of requiring `orgmode` directly.

lua/orgmode/org/links/handlers/_internal/custom_id.lua
```diff
@@ -0,0 +1,53 @@
...
+
+  if #headlines == 0 then
+    return utils.echo_warning(('Could not find custom ID "%s".'):format(self.custom_id))
+  end
+
+  self.goto_oneof(headlines)
```
kristijanhusak, line 33: If we are using `self` to call something, call it with a colon notation
```suggestion
  self:goto_oneof(headlines)
```

lua/orgmode/org/links/handlers/_internal/custom_id.lua
```diff
@@ -0,0 +1,53 @@
...
+end
+
+function CustomId:complete(lead, context)
+  local file = self.get_file_from_context(context)
+  local headlines = file:find_headlines_with_property_matching('CUSTOM_ID', lead)
+
+  local completions = {}
+  for _, headline in pairs(headlines) do
+    local id = headline:get_property('CUSTOM_ID')
+    table.insert(completions, self:new(id):__tostring())
```
kristijanhusak, line 47: Since we are using the metamethod to set up a string format of an entry, we should use `tostring(self:new(id))`. Other option is to have a custom `to_string` method and call that here directly. I'm ok with any of the options.

lua/orgmode/org/links/handlers/_internal/headline.lua
```diff
@@ -0,0 +1,68 @@
...
+end
+
+function Headline:resolve()
+  local headlines = Org.files:get_current_file():find_headlines_by_title(self.headline)
+
+  if #headlines == 0 then
+    return self
+  end
+
+  local id = headlines[1]:get_property('id')
```
kristijanhusak, line 44: Why are we getting the id property here? Ids are handled through a separate handler.

Also, we need to consider that there are multiple headlines here.

lua/orgmode/org/links/handlers/file.lua
```diff
@@ -0,0 +1,152 @@
...
+function File:follow()
+  vim.cmd('edit ' .. fs.get_real_path(self.path))
+
+  if self.target then
+    self.target:follow()
+  end
+end
+
+local function autocompletions_filenames(lead)
+  local Org = require('orgmode')
```
kristijanhusak, line 69: I see this is a very WIP, but just a note that we should not require orgmode directly anywhere. We need to pass down dependencies from top.

lua/orgmode/org/links/handlers/internal.lua
```diff
@@ -0,0 +1,114 @@
...
+  local this = Link:new(protocol)
+  setmetatable(this, self)
+  self.__index = self
+  return this
+end
+
+---@param target string | nil
+---@param disallow_file boolean?
+---@return OrgLinkHandlerInternal | OrgLinkHandlerFile | nil
+function Internal.parse(target, disallow_file)
```
kristijanhusak, line 19: This whole class is doing a bit too much, and also has the checks that we are trying to avoid with the sources.
Each source should do this check and know how to parse the value from the target. Some example code:

```lua
local sources = { Headline, CustomId, LineNumber, File, Plain }

for _, source in ipairs(sources) do
  local parsed = source:parse(target)
  if parsed then
  	return parsed
  end
end
```

This should happen in the top level class, not here.

lua/orgmode/org/links/init.lua
```diff
@@ -0,0 +1,130 @@
...
+    handlers = {},
+    handlers_by_name = {},
+    stored_links = {},
+  }, self)
+  self.__index = self
+  this:setup_builtin_handlers()
+end
+
+function OrgLinkHandlerRegistry:setup_builtin_handlers()
+  self:add_handler(require('orgmode.org.links.handlers.file'))
```
kristijanhusak, line 27: We should instantiate the handlers here and pass down the dependencies.

lua/orgmode/org/links/init.lua
```diff
@@ -0,0 +1,130 @@
...
+---@param handler OrgLinkHandler
+function OrgLinkHandlerRegistry:add_handler(handler)
+  if self.handlers_by_name[handler.protocol] then
+    error('Completion handler ' .. handler.protocol .. ' already exists')
+  end
+
+  -- Ensure that all handlers have defaults for the methods we rely on.
+  for k, v in pairs(Link) do
+    handler[k] = handler[k] or v
+  end
```
kristijanhusak, line 43: We don't need this for now since we are not exposing adding custom handlers yet.

lua/orgmode/org/links/init.lua
```diff
@@ -0,0 +1,130 @@
...
+end
+
+local function filter_by_prefix(lead)
+  return function(needle)
+    if lead == '' or needle:find('^' .. lead) then
+      return true
+    end
+    return false
+  end
+end
```
kristijanhusak, line 77: This part can be just inlined where needed since it's a simple check
```suggestion
local function filter_by_prefix(lead)
  return function(needle)
    return lead == '' or needle:find('^'..lead)
  end
end
```

kristijanhusak (changes requested):
Thanks for the PR!

There's progress from the previous PR, but we still have few things to handle.

Since things are fairly complicated, we can chunk these changes a bit more for now. For example, we can leave autocompleting the links for a separate PR, and for now just build the sources and all the structure around it, and maybe have a `follow` on the top `links` class. From there we can later add autocompletion.

lua/orgmode/org/links/handlers/_internal/headline.lua
SlayerOfTheBad: I think this was meant to implement the behaviour determined by `org-id-link-to-use-id` as described in https://orgmode.org/org.pdf#Handling%20Links.
Upon further reading however, this is resolved on storing a link, rather than on inserting it. I thought it happened at both. Removed

lua/orgmode/org/links/init.lua
SlayerOfTheBad: Applied

lua/orgmode/org/links/init.lua
SlayerOfTheBad: It's just as valid for protecting against internal handler's issues as it is for protecting custom handlers

lua/orgmode/org/links/handlers/_internal/custom_id.lua
SlayerOfTheBad: Used `tostring`, just forgot it existed tbh.

lua/orgmode/org/links/handlers/_internal/custom_id.lua
SlayerOfTheBad: Moved to `utils`, as it seems more applicable than as a method of `Internal`

SlayerOfTheBad:
Thanks for the review @kristijanhusak!

I've been processing a lot of the comments, but I'm left with a couple of questions:

> maybe have a `follow` on the top `links` class.
Assuming you mean `OrgLinks`, How do you imagine this working?
Should the function signature be `fun(link: OrgLinkHandler)`, and just call `link:follow()`?
`OrgLinks` doesn't really have anything it can `follow`, it's just a parser for a generic link.

And I'm currently of half a mind to split up each handler into a `LinkFactory` and a `Link`, so like:
```lua
local Id = Link:new('id')

function Id:new(id, target, files)
  -- Returns instance of Id
end

function Id:follow()
  -- Follows this instance
end

function Id:__tostring()
  -- Converts to link
end

local IdFactory = {}

function IdFactory:init(files)
  self.files = files
end

function IdFactory:new(id, target)
  return Id:new(id, target, self.files)
end

---@param input string
function IdFactory:parse(input)
  -- Parse id and target
  return self:new(id, target)
end

---@param lead string
function IdFactory:complete(lead)
  -- Find and return the completions
end

return IdFactory
```

This has the benefit of separating the link's functionality from the functionality to parse the link,
and ensures the files are always injected properly. Is this a structure that's desired?
If yes, should these two classes be split into separate files, or are they so interlinked they should live in the same file?
And how does this impact the structure for custom handlers?

(The `complete` function can be omitted from the factories for this PR as you suggested, but it's included here to illustrate the responsibilities of the Factory VS the responsibilities of the Link itself.)

lua/orgmode/org/links/handlers/internal.lua
SlayerOfTheBad: I split this up into a separate handler because these are a bit of a special case, on account of not having a `<protocol>:` prefix.

It seemed redundant to me to repeat the `<protocol>:` prefix handling in each of the `LinkHandler`s individually, so I put this logic in `OrgLinks`, but these five cases are special in the fact that they don't have prefixes.

These are also special in that these are the only allowed targets for `id:` and `file:` links after the `::`, so it made handling of that easier.

`file:` is an even more special case because its `<protocol>:` prefix is optional, so if we say each 'internal' link should be able to parse itself in its entirety, either `File` needs special treatment to parse both the prefixed case and the non-prefixed case, or all `LinkHandler`s need to parse their own prefix.

It's a weird set of special cases, and I'm not sure there's a perfect solution, but I'm interested in your thoughts.

kristijanhusak:
I'll attempt to answer both of your comments with an example. This is what I have in mind:

Top level `OrgLinks` would look something like this:

```lua
local OrgLinks = {}

function OrgLinks:new(opts)
  local this = setmetatable({
    files = opts.files,
    sources = {},
    sources_by_name = {},
  }, OrgLinks)
  this:setup_builtin_sources()
end

function OrgLinks:setup_builtin_sources()
  self:add_source(require('orgmode.org.links.sources.headline'):new({ files = self.files }))
  self:add_source(require('orgmode.org.links.sources.custom_id'):new({ files = self.files }))
  --- etc
end

function OrgLinks:add_source(source)
  if self.sources_by_name[source:get_name()] then
    error('Link source ' .. source:get_name() .. ' already exists')
  end
  self.sources_by_name[source:get_name()] = source
  table.insert(self.sources, source)
end

---@param link string
function OrgLinks:follow(link)
  -- For example, link = 'file:~/Documents/notes.org::headline'

  for _, source in ipairs(self.sources) do
    -- I named this `can_handle`, but we can think of a better name, maybe `source:matches` or `source:handles`
    if source:can_handle(link) then
      return source:follow(link)
    end

    -- Alternative is to atempt to follow immediately and return boolean if it was successful or not
    if source:follow(link) then
      return true
    end
  end
end

return OrgLinks
```

And then the headline link example would look like this:

```lua
local OrgLinkHeadline = {}

function OrgLinkHeadline:new(opts)
  local this = setmetatable({
    files = opts.files,
  }, OrgLinkHeadline)
end

---@param link string
function OrgLinkHeadline:can_handle(link)
  -- check if link have any of these formats
  -- * file:~/Documents/notes.org::*headline
  -- * *headline

  if 'matches_above_checks' then
    return true
  end

  return false
end

function OrgLinkHeadline:follow(link)
  local link_parts = self:parse_link_parts(link)
  -- if se use the altnerative approach to follow immediately in the for loop, we can just check if parsing parts was successful or not
  local headlines = self.files:find_headlines_by_title(link_parts.headline)
  return utils.goto_oenof(headlines)
end


return OrgLinkHeadline
```

And the custom id source would look like this:

```lua
local OrgLinkCustomId = {}

function OrgLinkCustomId:new(opts)
  local this = setmetatable({
    files = opts.files,
  }, OrgLinkCustomId)
end

---@param link string
function OrgLinkCustomId:can_handle(link)
  -- check if link have any of these formats
  -- * file:~/Documents/notes.org::#custom-id
  -- * #custom-id

  if 'matches_above_checks' then
    return true
  end

  return false
end

function OrgLinkCustomId:follow(link)
  local link_parts = self:parse_link_parts(link)
  -- if se use the altnerative approach to follow immediately in the for loop, we can just check if parsing parts was successful or not
  local headlines = self.files:find_headlines_with_property_matching('CUSTOM_ID', link_parts.custom_id)
  return utils.goto_oenof(headlines)
end

return OrgCustomId
```


We should not use factories since it complicates things additionally.

> It seemed redundant to me to repeat the <protocol>: prefix handling in each of the LinkHandlers individually, so I put this logic in OrgLinks, but these five cases are special in the fact that they don't have prefixes.

We should extract the parsing of it to a util or something and reuse it where necessary. In the top 2 examples we do need it to check if it's a `file` protocol. Note that `/path/to/file::*headline` is also a valid link for a headline (without a protocol), so there's more to it.

Hope this helps clearing up some of the misunderstandings.

SlayerOfTheBad:
Thanks for the response @kristijanhusak.

I'm willing to concede on parsing the protocol prefix within their own handler.
It's sensible enough, and the logic for it can indeed be fit into a reused utility function.

I do have a concern about the example you provided however.
It seems to suggest that links will be passed around as strings, and parsed any time it's accessed.
Is this actually what you would suggest, or is this just for terseness of the example?

kristijanhusak:
> It seems to suggest that links will be passed around as strings, and parsed any time it's accessed.
Is this actually what you would suggest, or is this just for terseness of the example?

It's just an example. It can be an instance of something, like what currently is `OrgLink`.

kristijanhusak:
I'm not really liking the idea of modifying the handlers in that way.
I think this is mostly to simplify custom sources that we will add later, but we can address those differently.

> 1. Eliminates the `Factory` components

I don't think we need this anyway. 

> 2. Allows custom link handles to be defined as a single table

I believe it will be only a single table. We just need to inject the `files` somehow. I gave some options at the end.

> 3. Only requires a link handler's `parse` function to return an object that contains its own data. The functions are attached automatically.

I'm not sure what we would share besides `files`, since each source would need to implement their own `parse` and `follow` (and later autocompletion). Even if there's something, we can just expose a helper class that users can use to extend their own.

> 4. Gives the objects returned by `parse` access to the `LinkHandler`'s internal fields (such as `self.files` in this case).

For giving access to the `files` for custom sources, we can do 2 things:

1. Not provide anything, and let user manually reference `require('orgmode').files`
2. Assign the `files` manually when adding the source. Thankfully, it's just a table so we can assign things
```lua
local OrgLinks = {
    files = {},
    handlers = {},
    handlers_by_name = {},
}

function OrgLinks:setup_builtin_handlers()
    -- Note that I removed `files` here.
    self:add_handler(require('orgmode.org.links.handlers.file'):new())
end

---@param handler OrgLinkHandler
function OrgLinks:add_handler(handler)
    if self.handlers_by_name[handler.protocol] then
        error('Completion handler ' .. handler.protocol .. ' already exists')
    end

    --- assign files to the handler so they can use it
    handler.files = self.files

    self.handlers_by_name[handler.protocol] = handler
    table.insert(self.handlers, handler)
end
```

Let me know if I misunderstood what you are trying to solve here.

SlayerOfTheBad:
The problem I'm trying to solve is a certain combination of requirements, some self-imposed, some imposed in the review. I will try to enumerate them here.

For examples I will use a handler for `id:`

1. Have `OrgLinkHandler:parse(input: string)` return an object (`OrgLink`) that
  - Contains all the fields it needs (`id`, `files`)
  - Contains all the actions that can be executed on that link (`follow`, `__tostring`, `insert_description`)
2. Have fields that are the same between different `OrgLink` instances of the same type be automatically inserted (`files`)
  - Would normally do this through `require('orgmode').files`, as you've suggested for custom handlers, but you've also expressed a desire for these dependencies to be passed down from the top
  - This is usually done through a factory pattern, but you've rejected this for complexity reasons.
3. Treat user-defined hyperlinks in the same way as internal hyperlinks.
  - Reduces complexity
4. Not require the user to manually add the functions to the object returned from their `OrgLinkHandler:parse`.
  - Not nice if you want to just define your custom hyperlink as a table.


I don't think just adding `files` to the handlers is a great solution, as different types of links have different requirements. For example: The way I currently envision `OrgLinkId`, is having a field `target: OrgLink`, which stores the `OrgLink` associated with the part after `::`. Thus
`id:some-file::*some-title`
would be parsed to
```lua
{
  id: 'some-file',
  files: OrgFiles,
  target: {
    headline: 'some-title',
    files: OrgFiles,
  },
}
```

To be able to parse it to something like that, `OrgLinkHandlerId` needs a reference to `OrgLinks`, so it can use that to parse `*some-title`, but `OrgLinkHandlerHeadline` does not need this.


Let me know if anything is unclear, or if you think I missed or misunderstood anything.

kristijanhusak:
I agree with first 3 points. 

Regarding the 4th point:

> 4. Not require the user to manually add the functions to the object returned from their OrgLinkHandler:parse.
> * Not nice if you want to just define your custom hyperlink as a table.

I think we have misunderstanding on what `parse` should do. In [my example](https://github.com/nvim-orgmode/orgmode/pull/793#issuecomment-2291005358) I didn't actually create a `parse` method, but `can_handle`. Parsing the input should be internal to each handler. With my example, `can_handle` would call an internal `parse` method, and if it's able to parse it, would return `true`. Then `follow` would do same thing and actually execute `follow` with the parsed value. 
There is some double parsing here to avoid creating factories, and that's why I suggested to immediately attempt to follow. I'll put a more detailed example below.

> I don't think just adding files to the handlers is a great solution, as different types of links have different requirements.

That's true, but I think we can pass `files` to all of them and have that as part of the interface they are implementing. If some source does not use it, that's fine. There's no overhead in that.

> For example: The way I currently envision OrgLinkId, is having a field target: OrgLink, which stores the OrgLink associated with the part after ::. Thus
id:some-file::*some-title
> To be able to parse it to something like that, OrgLinkHandlerId needs a reference to OrgLinks, so it can use that to parse *some-title, but OrgLinkHandlerHeadline does not need this.

We should not reference the "parent" class in the handlers. We can have some helpers in another utils file.
So if a specific handler needs a `target`, we can have a util helper `parse_target('url with double colon::*something')`
`id` format does not actually support `target`, since it points to a very specific entry. This is a good example why each handler should have their own parsing. 

Here's a more concrete example of how I envisioned it. Mostly copied from my previous comment, with more details around parsing:

Top level `OrgLinks` would look something like this:

```lua
local OrgLinks = {}

function OrgLinks:new(opts)
  local this = setmetatable({
    files = opts.files,
    sources = {},
    sources_by_name = {},
  }, OrgLinks)
  this:setup_builtin_sources()
end

function OrgLinks:setup_builtin_sources()
  self:add_source(require('orgmode.org.links.sources.id'):new({ files = self.files }))
  self:add_source(require('orgmode.org.links.sources.custom_id'):new({ files = self.files }))
  --- etc
end

function OrgLinks:add_source(source)
  if self.sources_by_name[source:get_name()] then
    error('Link source ' .. source:get_name() .. ' already exists')
  end
  self.sources_by_name[source:get_name()] = source
  table.insert(self.sources, source)
end

---@param link string
function OrgLinks:follow(link)
  for _, source in ipairs(self.sources) do
    if source:follow(link) then
      return true
    end
  end
end

return OrgLinks

```

And then the id handler example would look like this:

```lua
local OrgLinkId = {}

function OrgLinkId:new(opts)
  local this = setmetatable({
    files = opts.files,
  }, OrgLinkId)
end

---@param link string
---@return boolean
function OrgLinkId:follow(link)
  local link_info = self:_parse(link)
  if not link_info then
    return false
  end
  local id = link_info.id
  -- Part below copied from `mappings:open_at_point`
  local files = self.files:find_files_with_property('id', id)
  if #files > 0 then
    if #files > 1 then
      utils.echo_warning(string.format('Multiple files found with id: %s, jumping to first one found', id))
    end
    vim.cmd(('edit %s'):format(files[1].filename))
    return
  end

  local headlines = self.files:find_headlines_with_property('id', id)
  if #headlines == 0 then
    return utils.echo_warning(string.format('No headline found with id: %s', id))
  end
  if #headlines > 1 then
    return utils.echo_warning(string.format('Multiple headlines found with id: %s', id))
  end
  local headline = headlines[1]
end

---Note that this is an internal method and should not be part of interface that handlers need to implement
---@private
---@param link string
---@return { id: string } | nil
function OrgLinkId:_parse(link)
  local id = link:match('^id:(.+)$')
  if not id then
    return nil
  end
  return { id = id }
end


return OrgLinkId

```


It's ok if we need to duplicate some of the code, as long as is scoped within a single handler. Once everything is finished, we can easily extract reusable parts.

Hope this clarifies things a bit.

SlayerOfTheBad:
I am sorry, but the structure you propose is not something I would enjoy writing. I do not have the time, nor the energy, to continue pushing this PR anymore. Feel free to use what I have made here as a basis for your own implementation of this.

kristijanhusak:
I very much appreciate all the time you put into this, and I'm sorry that it became a burden because of my strict guidelines. 
Once I do the implementation, I'll definitely consult you about the custom sources, since you seem eager to start using them.
Thanks!

PR #795, opened 2024-08-15
fix: explicitly pass treesitter language to `get_node()`

troiganto:
Fixes chipsenkbeil/org-roam.nvim#49

Org-roam offers a "node buffer", a sidebar that shows information such as backlinks from other locations to the current headline/file. The node buffer allows previewing the location of such a backlink. This inserts orgmode syntax into a buffer with filetype `org-roam-node-buffer`.

If virtual indent is enabled, it'll attempt to deduce the virtual indent of the preview. This in turn calls `closest_headline_node()`, which in turn calls `get_node_at_cursor`, which in turn calls `vim.treesitter.get_node()`.

If `get_node()` is called without an explicit treesitter language, it'll attempt to deduce that language from the filetype. The filetype `org-roam-node-buffer` isn't associated with the `org` language (and I don't think it *should*), so this call fails.

This fixes the issue by explicitly passing `{ lang = 'org' }` in all instances where the language *may* be deduced. I think this is acceptable because if someone calls
`orgmode.utils.treesitter.get_node()` AFAIK is only ever called on text that is org syntax.

While making this change, I grepped for other treesitter calls with an optional language argument. Where I found them, I passed it explicitly as well. AFAICT, this is only `Stars:on_line()` and `ts_utils.restart_highlighting()`.

### DIFF
--- a/lua/orgmode/colors/highlighter/stars.lua
+++ b/lua/orgmode/colors/highlighter/stars.lua
@@ -22,6 +22,7 @@ function OrgStars:on_line(bufnr, line)
   local node = vim.treesitter.get_node({
     bufnr = bufnr,
     pos = { line, 0 },
+    lang = 'org',
   })
 
   if not node or node:type() ~= 'stars' then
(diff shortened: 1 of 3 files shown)


kristijanhusak (approved):
Looks good!
I'll merge it in and extract it to own helper so we are sure we always get the node with proper language.

Thanks!

PR #797, opened 2024-08-18
Limit _get_date_under_cursor to parsing current line

### LINKED ISSUE #796: ctrl-a/ctrl-x do not reliably modify dates
### Describe the bug

Depending on where a date is in a file, it may or may not update correctly with ctrl-a/ctrl-x (timestamp_up/timestamp_down).

When the operation fails, the number field is observed to be updated in the opposite direction from that intended, and the update is not recognized as an update to a date field.

### Steps to reproduce

Using this test file:

```
* TODO testing <2024-08-17 Sat> 
* stuff <2024-01-17 Wed> 
  <2024-02-17 Sat> 
  <2024-03-17 Sun> 
```

Placing the cursor the day portion of each line and pressing ctrl-a results in the following behavior for each line:
1. Incorrectly changes to <2024-08-16 Sat> and day of week does not change
2. Incorrectly changes to <2024-01-16 Wed> and day of week does not change
3. Correctly changes to <2024-02-18 Sun>
4. Incorrectly changes to <2024-03-16 Sun> and day of week does not change

The decrement action is likely a result of nvim default behavior that interprets the value as a negative number (-17), thus increasing it to -16.

### Expected behavior

ctrl-a and ctrl-x should recognize the date on each line and increment/decrement accordingly.

### Emacs functionality

_No response_

### Minimal init.lua

```lua
vim.opt.rtp:prepend(vim.env.REAL_HOME .. "/.local/share/nvim/lazy/orgmode")
require("orgmode").setup {
                        org_agenda_files = { '~/test-orgfiles/**/*', },
                        org_default_notes_file = '~/test-orgfiles/refile.org',
}
```

_No response_

### OS / Distro

OSX 14.6.1

### Neovim version/commit

NVIM v0.10.1 Build type: Release LuaJIT 2.1.1723675123

### Additional context

Tested using commit adf277c

Starting from `OrgMappings:_adjust_date_part()`, and given that the cursor is on line 1, col 26 when ctrl-a is pressed in the sample file mentioned above, The date object referenced by `date_on_cursor` contains `range.end_col = 29` and `range.start_col=14`. Considering the line of the sample file:

```
1234567890123456789012345678901
* TODO testing <2024-08-17 Sat>
             \-start_col     \-end_col
```

The values are 2 less than they should be to correctly reference this line of text. When increment/decrement work, the start/end_col values reference the locations of the date field delimiters on the line. (I haven't explored why the 4th line doesn't work but it may related to the TODO at the end of `OrgMappings:_get_date_under_cursor()`

I drilled down and found that in `Headline:get_non_plan_dates()` the value of `headline_text` that is used to locate the date does not include the leading `* ` from the buffer. This would explain the off-by-two error in locating the date in the original line.

I haven't sorted out what the functions are doing to set the value of `headline_text` (I'm still learning Lua) but this is getting close to where the problem is. I think the fix is to ensure that the headline text represents the full line of text, but I'm both unclear how to do that, and unclear what other side effects that may have on callers that expect this section of code to be working based on tree sitter objects vs. lines of text. Given my lack of understanding, it's also possible that `OrgMappings:_adjust_date_part()` should be using tree sitter objects where now it appears to be using text lines.

Inspecting the first line of the example file, tree-sitter is not identifying the date on this line as a timestamp. If I move the date to its own line, inspecting the tree includes a timestamp node. Part of the problem?

KnBrckr:
To retrieve a date under the cursor, limit the scope to the current line. Parse all dates on the line, then filter the table to retrieve the member containing the cursor.

Addresses the TODO that would fail to return the correct entry when multiple are found on a given line.

Closes #796

### DIFF
--- a/lua/orgmode/org/mappings.lua
+++ b/lua/orgmode/org/mappings.lua
@@ -1135,21 +1135,14 @@ function OrgMappings:_get_date_under_cursor(col_offset)
   col_offset = col_offset or 0
   local col = vim.fn.col('.') + col_offset
   local line = vim.fn.line('.') or 0
-  local item = self.files:get_closest_headline_or_nil()
-  local dates = {}
-  if item then
-    dates = vim.tbl_filter(function(date)
-      return date.range:is_in_range(line, col)
-    end, item:get_all_dates())
-  else
-    dates = Date.parse_all_from_line(vim.fn.getline('.'), line)
-  end
+  local dates = vim.tbl_filter(function(date)
+    return date.range:is_in_range(line, col)
+  end, Date.parse_all_from_line(vim.fn.getline('.'), line))
 
   if #dates == 0 then
     return nil
   end
 
-  -- TODO: this will result in a bug, when more than one date is in the line
   return dates[1]
 end
 


kristijanhusak:
I fixed the underlying issue in d0baf31181435c2378dc2c19a57dcebcc36d9464 so this PR is not needed any more. Thanks!

PR #803, opened 2024-08-23
feat(links)!: Use refactor links structure

kristijanhusak:
Start using new links structure introduced in https://github.com/nvim-orgmode/orgmode/pull/802.

@chipsenkbeil can you check if using this branch breaks anything in org-roam.nvim? It shouldn't, but it's best to double check.
You will need to introduce one change here: https://github.com/chipsenkbeil/org-roam.nvim/blob/192732a8de167e7770b12db31e737239311b1a43/lua/org-roam/setup/plugin.lua?plain=1#L20

I added a hyperlink class in https://github.com/nvim-orgmode/orgmode/blob/71c57cab31b78d2c6907d2961b6c2666ea9383a1/lua/orgmode/org/links/hyperlink.lua?plain=1#L8, and it works basically the same for your usage. Underlining `url` prop is also a different class, but it has the `get_id` method that you are using.

For this line you can use `Hyperlink.at_cursor` function instead of these 3 lines to get current link.

Besides this I think everything should be working as expected, but please give it a thorough test. 

@seflue @SlayerOfTheBad could you switch to this branch and see if everything related to hyperlinks works as expected? 

Thanks!

### DIFF
--- a/lua/orgmode/api/file.lua
+++ b/lua/orgmode/api/file.lua
@@ -1,6 +1,5 @@
 ---@diagnostic disable: invisible
 local OrgHeadline = require('orgmode.api.headline')
-local Hyperlinks = require('orgmode.org.hyperlinks')
 local org = require('orgmode')
 
 ---@class OrgApiFile
@@ -112,12 +111,12 @@ function OrgFile:get_link()
     -- do remote edit
     return org.files
       :update_file(filename, function(file)
-        return Hyperlinks.get_link_to_file(file)
+        return org.links:get_link_to_file(file)
       end)
       :wait()
   end
 
-  return Hyperlinks.get_link_to_file(self._file)
+  return org.links:get_link_to_file(self._file)
 end
 
 return OrgFile
--- a/lua/orgmode/api/headline.lua
+++ b/lua/orgmode/api/headline.lua
@@ -4,7 +4,6 @@ local PriorityState = require('orgmode.objects.priority_state')
 local Date = require('orgmode.objects.date')
 local Calendar = require('orgmode.objects.calendar')
 local Promise = require('orgmode.utils.promise')
-local Hyperlinks = require('orgmode.org.hyperlinks')
 local org = require('orgmode')
 
 ---@class OrgApiHeadline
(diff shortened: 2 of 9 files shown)


kristijanhusak:
@chipsenkbeil @seflue @SlayerOfTheBad mentioning just in case you didn't receive the notification, since I mentioned you in an edit of a description.

chipsenkbeil:
@kristijanhusak thanks for keeping me in the loop :) I got your first mention, but glad you double-checked.

I can get this change into the plugin in the next day or two. Is this going to mark a new release? Wondering if I should

1. Mark a new release that locks to a new version of the orgmode plugin so I can change the code to use the update
2. Or, update my code to try to use the new method and fall back to the old approach if it doesn't exist, thereby supporting both changes. Would still mean that folks using older versions of my plugin would break if they updated to your latest

Leaning towards the former if you're planning to mark this as a new orgmode release 😄

kristijanhusak:
I put out a release 0.3.5 yesterday that does not have any of these link changes. You can pin your current version to this version.

Once i merge this, i can tag another one, but I'd prefer to iron out any bugs before doing that.

chipsenkbeil:
> Edit: Note that your current code should not break even on this PR since i didn't remove old code, but please double check.

Ah, okay! I thought this was a callout to breaking change. If this is not breaking then even better. 

I think I'd like to wait towards making the change until after the tag, but I can create a branch of my plugin with the change to verify that it works with the new code. I just won't merge it until I can tag a new release on my side that syncs up with yours.

So TLDR, I'll help verify that things work and the change behaves as expected, but will wait to land until after it rolls out in a tagged orgmode release.

seflue:
> @chipsenkbeil @seflue @SlayerOfTheBad mentioning just in case you didn't receive the notification, since I mentioned you in an edit of a description.

@kristijanhusak Thanks for the reminder. I will check, if I need to adjust something in the next days and update you here.

kristijanhusak:
I'm merging this in. We can solve any issues as we go. Org-roam suggests using 0.3.4 version so we should be ok.

chipsenkbeil:
@kristijanhusak all good to merge since I lock versions.

Haven't set aside time to check. Will make a note to try to do so this week.

chipsenkbeil:
@kristijanhusak months later, and while I didn't explicitly check anything I've been on the latest for org-roam.nvim and orgmode without issue, so I think the links are fine.

seflue:
I can agree from my side. I am using telescope-orgmode on a daily basis, especially to create links and I have no issues.

kristijanhusak:
Awesome, thanks!

PR #812, opened 2024-10-02
fix: fix autocomplete for tags search prompt

hasansujon786:
`<leader>oa` + `m` - Match a TAGS/PROP/TODO query prompt autocomplete isn't working on `<tab>`.


kristijanhusak (approved):
Thanks!

PR #814, opened 2024-10-16
feat(org_mappings): add use_fast_access for todo_next_state and todo_prev_state

LongyuXie:
In Emacs, we can use `<S-left>` to change todo state without fast_access.
So i add the option `use_fast_access` for `todo_next_state` and `todo_prev_state`.
The default value is `true`, and it will not affect other parts.
```
vim.keymap.set('n', '<S-left>', function()
    require("orgmode").action("org_mappings.todo_prev_state", false)
end)
```
This keymap will work unless you [configure the the shift keycodes](https://stackoverflow.com/a/42461580).

### DIFF
--- a/lua/orgmode/org/mappings.lua
+++ b/lua/orgmode/org/mappings.lua
@@ -340,12 +340,12 @@ function OrgMappings:set_priority(direction)
   headline:set_priority(new_priority)
 end
 
-function OrgMappings:todo_next_state()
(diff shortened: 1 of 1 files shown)


kristijanhusak (changes requested):
Changes looks incorrect. I think your branch is outdated. Rebase from master and apply the changes again.

Issue #813, opened 2024-10-09
Fix checkbox issues with Nightly


kristijanhusak:
### Describe the bug

With the tree-sitter update https://github.com/neovim/neovim/pull/30589 , few features related to checkbox started to fail.

### Steps to reproduce

Create a checkbox, try to toggle it.

### Expected behavior

-

### Minimal init.lua

```lua

-- Enter your minimal_init.lua here

```

### OS / Distro

Linux

### Neovim version/commit

NVIM v0.11.0-dev-936+gf449a38f6

kristijanhusak:
Minor issue fixed in c6540955acbe.
Other issues coming from https://github.com/neovim/neovim/issues/30792

lyz-code:
@kristijanhusak I think this one has been solved with the latest changes

ynhhoJ:
I'm not sure if this is correct issue for that, but i can't toggle `A` checkbox using `<C-Space>`.

Example:
```md
- [ ] A
- [ ] B
```

If I press `<C-Space>` with cursor on `B`, everything works good.

`:Inspect` result of `A` checkbox:

![Image](https://github.com/user-attachments/assets/c96dd074-1ec4-42e5-af43-bdafabf75f41)

kristijanhusak:
@ynhhoJ does it happen on stable Neovim version, like 0.10.3?

ynhhoJ:
@kristijanhusak , just tested on `0.10.4` and everything works fine, seems that can be a problem in `0.11.0` version.

My `nvim -v`:

```
NVIM v0.11.0-dev-1664+ge71d2c817d
Build type: RelWithDebInfo
LuaJIT 2.1.1736781742
Run "nvim -V1 -v" for more info
```

PR #815, opened 2024-10-18
feat(refile): improve ux by matching files more fuzzily

BreitbandModem:
Hi, I really like this project!

I'm using the org-roam plugin, and it will prefix every file with the current timestamp to make sure it's unique (e.g. `20241013095846-helloworld.org`).
Refiling to such files is really inconvenient as I need to circle through all filenames with the TAB key.
Searching for it is really cumbersome as I'd have to type in the exact timestamp of this orgfile.

My suggestion is to remove the anchor (`^`) from the file pattern matching, and allow the search string to be anywhere in the filename.
This way I can just type `hello` or even `world` to quickly select the refile target.

### DIFF
--- a/lua/orgmode/capture/init.lua
+++ b/lua/orgmode/capture/init.lua
@@ -408,19 +408,17 @@ end
 ---@param arg_lead string
 ---@return string[]
 function Capture:autocomplete_refile(arg_lead)
-  local valid_filenames = self:_get_autocompletion_files(true)
+  local valid_files = self:_get_autocompletion_files(true)
 
-  if not arg_lead then
-    return vim.tbl_keys(valid_filenames)
+  if not arg_lead or #arg_lead == 0 then
+    return vim.tbl_keys(valid_files)
   end
-  local parts = vim.split(arg_lead, '/', { plain = true })
 
-  local selected_file = valid_filenames[parts[1] .. '/']
+  local filename = vim.split(arg_lead, '/', { plain = true })[1]
+  local selected_file = valid_files[filename .. '/']
 
   if not selected_file then
-    return vim.tbl_filter(function(file)
-      return file:match('^' .. vim.pesc(parts[1]))
-    end, vim.tbl_keys(valid_filenames))
+    return vim.fn.matchfuzzy(vim.tbl_keys(valid_files), filename)
   end
 
   local headlines = selected_file:get_opened_unfinished_headlines()


kristijanhusak:
@seflue tree-sitter was updated on master and it broke some of the functionality. I created an issue to track that https://github.com/nvim-orgmode/orgmode/issues/813. If other versions are passing we are ok.

lua/orgmode/capture/init.lua
```diff
@@ -419,7 +419,7 @@ function Capture:autocomplete_refile(arg_lead)
 
   if not selected_file then
     return vim.tbl_filter(function(file)
-      return file:match('^' .. vim.pesc(parts[1]))
+      return file:match(vim.pesc(parts[1]))
```
seflue, line 422:
```suggestion
      return string.lower(file):match(string.lower(vim.pesc(parts[1])))
```
Fuzzy finding is most convenient, if it's case-insensitive.

seflue (changes requested):
If you can make the matching case insensitive (see suggestions) I would very appreciate to see this merged, because it improves the usability a lot.

lua/orgmode/capture/init.lua
BreitbandModem: yes agree, that's even better 👍

kristijanhusak: What about [matchfuzzy](https://neovim.io/doc/user/builtin.html#matchfuzzy()) ?

seflue: @kristijanhusak Implementing matchfuzzy is different and would take much more of an effort while this simple line of code brings immediately a massive improvement. I would prefer to do experimenting with matchfuzzy in a separate PR. It is also not clear, that it would improve usability, because fuzzy matching can return an unexpectedly large result set and how intuitively it feels depends a lot on the underlying algorithm and implementation.

@kristijanhusak I would prefer to experiment with matchfuzzy in a separate PR. Implementing matchfuzzy is something different and would require a lot more effort, while this simple line of code will immediately bring a massive improvement in usability. I would also like to point out that fuzzy matching does not necessarily improve usability. It can produce an unexpectedly large result set, and how intuitive it feels depends a lot on the underlying algorithm and implementation. Which again raises the effort to implement it correctly, because it needs more experimentation.

BreitbandModem: ... I just pushed a version using the `matchfuzzy` 😅 
Doesn't seem to be much effort to implement it. It actually simplified the code imo.
In terms of usability - yes it might yield premature search results. But I don't think that's too bad for usability.

Btw, I found that the first safeguard condition that check the search string is not working.
The `arg_lead` is not nil, it's just an empty string..

seflue: I take whatever improves the current situation :wink: I was just speaking from my experience with telescope-orgmode, and that it can be hard to fine-tune the result set to something that feels intuitive. And I wasn't sure how motivated you @BreitbandModem would be to change your implementation, because although you modestly state that it wasn't much effort, it's actually a bigger change than just removing a few characters.

Many thanks for the contribution. :+1:

kristijanhusak: Let's go with the fuzzy matching for now, and we can change it later. I'm not using this autocomplete so often,p so if you run into some issues with matching feel free to open up a separate issue.

lua/orgmode/capture/init.lua
```diff
@@ -408,19 +408,17 @@ end
...
 
-  local selected_file = valid_filenames[parts[1] .. '/']
+  local filename = vim.split(arg_lead, '/', { plain = true })[1]
+  local selected_file = valid_files[filename .. '/']
 
   if not selected_file then
-    return vim.tbl_filter(function(file)
-      return file:match('^' .. vim.pesc(parts[1]))
-    end, vim.tbl_keys(valid_filenames))
+    return vim.fn.matchfuzzy(vim.tbl_keys(valid_files), vim.pesc(filename))
```
kristijanhusak, line 421: We don't need `vim.pesc` here, it is used for lua pattern matching, but `matchfuzzy` is vimscript function.

kristijanhusak (changes requested):
Looks good, but we don't need `vim.pesc` so you can remove that part.

lua/orgmode/capture/init.lua
BreitbandModem: Thanks for the hint!
I removed it.

Btw. I hope you don't mind me force pushing.
Not sure what you'd consider best practice in this project.
But for now I didn't wanna bloat the commit history...

seflue: > Btw. I hope you don't mind me force pushing.
> Not sure what you'd consider best practice in this project.
> But for now I didn't wanna bloat the commit history...

I force-push on PRs all the time and @kristijanhusak never complained. Because the PRs get squash-merged anyway I think it is no big deal. 🙂

kristijanhusak (approved):

kristijanhusak:
Thanks!

PR #817, opened 2024-10-20
feat(priority): improve priority cycling

seflue:
Make priority cycling more intuitive and closer to Emacs behavior:

- start cycle at default priority (default) or above/below default
priority, depending on config variable
- remove "none" state from the cycle

This PR succeeds #236.


kristijanhusak (approved):
Looks good, thanks!

PR #825, opened 2024-11-14
fix(virtual indent): handle invalid buffer gracefully

seflue:
Avoid errors when refiling content with headlines and enabled virtual indentation.

Cause: Indentation is also implicitly calculated during refiling. In this case the buffer could already be invalid.


kristijanhusak (approved):
Thanks!

Issue #483, opened 2023-01-05
Filtering the agenda using the API ignores categories


mdietrich16:
### Describe the bug

Using the normal agenda and TODO views, filtering can be done on tags and categories. When using the API, filtering seems to ignore categories.
When the TODO-list has been brought up before and a filter was applied, using the API to bring up the TODOs and set the filter ignores the categories and applies them on reapplying the filter.
However, when no filter was applied beforehand, using the API breaks the filtering system completely until the "normal" way of bringing up the TODOs und filtering resets the filter.
Also, the API agenda and TODOs only work after the normal agenda was pulled up before.

### Steps to reproduce

Bringing up the TODOs with `<leader>oat`, then filtering with `/` with a query of `-@tag-category` filters successfully. Quitting the TODO view and bringing up the to-dos with
```lua
require("orgmode.api.agenda").todos({ filters = "-@tag-category" })
```
filters out the `@tag`-tag, but not the category. If you bring up the filter again with `/` and simply press `<CR>`, the filter gets applied with the categories filtered.

However, after restarting Neovim, and only bringing up the TODOs with `<leader>oat`, not filtering leads to a different outcome. Quitting the view and bringing up the to-dos with
```lua
require("orgmode.api.agenda").todos({ filters = "-@tag-category" })
```
filters out the `@tag`-tag, but not the category, but upon bringing up the filtering prompt again and just reapplying with `<CR>` the filtering is reset and does not work anymore. Only bringing up the usual TODO view and applying a different filter solves this.

### Expected behavior

Tags as well as categories should be filtered when using the API to access the agenda, like when using the normal filtering with an opened agenda.

### Minimal init.lua

```lua
local status_ok, orgmode = pcall(require, "orgmode")
if not status_ok then
    return
end

orgmode.setup_ts_grammar()

orgmode.setup({
    org_agenda_files = { '~/Sync/org/gtd.org', '~/Sync/org/inbox.org', '~/Sync/org/calendar.org',
        '~/Sync/org/someday.org',
        '~/Sync/org/reference/*.org' },
    org_default_notes_file = '~/Sync/org/inbox.org',
    org_todo_keywords = { 'TODO(t)', 'WAITING(w)', 'NEXT(n)', '|', 'DONE(d)', 'CANCELLED(c)', 'DEFERRED(f)' },
    org_todo_keyword_faces = {
        NEXT = ":foreground white :background red :weight bold",
        WAITING = ":underline on",
    },
[...]

PR #831, opened 2024-11-22
fix: allow query to tags agenda api

lyz-code:
This is as far as I've gotten debugging https://github.com/nvim-orgmode/orgmode/issues/483 

As you can see my lua skills are awful, but after playing **a lot** with `options.filters` and `options.search` the only way the agenda works is if I don't touch any of those properties (I have no clue why)

The issue is that before this ugly patch if you run `require("orgmode.api.agenda").tags({ query = "+t/-DONE" })` the agenda returns no results because the query is disregarded and the user is prompted for the query as if you were to open the agenda yourself.

I know that this code is not ready for production, but maybe someone will pity me and lend me a hand (•‿•) . Or at least users may use my fork until this is solved to get the functionality

### DIFF
--- a/lua/orgmode/agenda/views/tags.lua
+++ b/lua/orgmode/agenda/views/tags.lua
@@ -22,6 +22,7 @@ function AgendaTagsView:new(opts)
     highlights = {},
     items = {},
     search = opts.search or '',
+    query = opts.query or '',
     todo_only = opts.todo_only or false,
     filters = opts.filters or AgendaFilter:new(),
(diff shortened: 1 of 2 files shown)


kristijanhusak:
I think https://github.com/nvim-orgmode/orgmode/pull/848 addresses most of these issues. Now you can provide a `match_query` to agenda tags (or tags_todo that was added), and that will apply the filter in the same way as you would provide it with `Match:` prompt.

PR #840, opened 2024-12-17
fix(help): ignore disabled mappings

olexsmir:
Hey :wave:

I have disabled a few keymaps in my config following [docs](https://github.com/nvim-orgmode/orgmode/blob/master/DOCS.md#mappings), and when I opened the help menu(`g?`), it was failing to open due to disabled keymaps.


kristijanhusak (approved):
Thanks!

PR #841, opened 2024-12-18
fix(template): expandable variables in lua expr

moreka:
This fixes the order which the template is compiled. Previously, the expressions `%()` were compiled _before_ expandable variables (such as `%x`). This made the example template in [1] to produce:
```org
* [[url][x]]
```
instead of
```org
* [[url][end part of the url]]
```
By doing the expression compilation _after_ expandable ones, this issue is resolved.

[1] https://github.com/nvim-orgmode/orgmode/wiki/Getting-Started#captures


kristijanhusak (approved):
Thanks!

PR #869, opened 2025-01-22
refactor(date): Refactor date object

kristijanhusak:


kristijanhusak:
@chipsenkbeil Just letting you know about this refactor in case you notice anything strange in org-roam.
Everything should work as usual. I have just rewritten a few things to make more sense.
There are a few "breaking"changes:
1. `Date.today()` now has a timestamp with the start of today (00:00). Previously, timestamp was also considering the current time. Formatting worked ok because of the `date_only` flag, but if you would access the `timestamp` directly, it wouldn't reflect that. Same thing for `Date.tomorrow()`
2. Date comparison (`:is_before()`, etc.) now considers the `date_only` flag, and sets the timestamp to the start of day in case that flag is set on a specific date.

I took a quick look at org-roam and I think none of these things affect you, but wanted to let you know anyway.

chipsenkbeil:
> There are a few "breaking"changes:
> 1. `Date.today()` now has a timestamp with the start of today (00:00). Previously, timestamp was also considering the current time. Formatting worked ok because of the `date_only` flag, but if you would access the `timestamp` directly, it wouldn't reflect that. Same thing for `Date.tomorrow()`

Good to know! I make use of the timestamp field within the `extensions/dailies.lua` code. It's only used to replace date variables within templates, which being based on calendar dates I wouldn't expect the time component to matter. Having it start at the beginning of the day is a better behavior from my perspective.

PR #873, opened 2025-01-28
Run viewer asynchronously using `vim.system`

moreka:
Previously, when the viewer opened, the editor was blocked until the user closes the viewer. Now, the viewer opens asynchronously (using `vim.system` rather than the old `vim.fn.system`); in this way, the user can view the PDF while continuing editing the org file in the editor.

### DIFF
--- a/lua/orgmode/utils/init.lua
+++ b/lua/orgmode/utils/init.lua
@@ -93,15 +93,18 @@ end
 
 function utils.open(target)
   if vim.fn.executable('xdg-open') == 1 then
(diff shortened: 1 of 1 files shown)


moreka:
I understand also that the next step might be to add support for detecting PDFs that are already open in a PDF viewer and not opening them after export (and just updating them). But I guess this is going to be difficult to maintain and is not the main functionality of this plugin as one might expect.

kristijanhusak (approved):
Thanks!

PR #875, opened 2025-01-30
ci, doc: add semver workflow and PR template

### LINKED ISSUE #828: Automated workflow for semantic releases

N/A

### Orgmode link

_No response_

### Feature value

The last release is a few months old. It would me nice if new releases with fixes could be published more frequently.
Since you already use conventional commits, perhaps an automated workflow with [the release-please-action](https://github.com/googleapis/release-please-action) would make it easier to publish releases more frequently?

### Additional context

I'd be happy to contribute a workflow if you're open to it :smile:

ficcdaf:
## Summary

This PR adds a GitHub actions workflow to automate the creation of semantically versioned releases with `release-please` and a PR template to encourage compliance with conventional commits.

## Related Issues

Closes #828 

## Changes

- Added `release-please` workflow
- Added PR template
- Updated Contributing documentation to mention conventional commits

## Checklist

I confirm that I have:

- [x] **Followed the
      `make test`.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,19 @@
+on:
+  push:
+    branches:
+      - master
+
+permissions:
+  contents: write
+  pull-requests: write
+
+name: release-please
+
+jobs:
+  release-please:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: googleapis/release-please-action@v4
+        with:
+          token: ${{ secrets.GH_TOKEN }}
+          release-type: simple
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,34 @@
+## Summary
+
+<!-- Give a brief description of what your PR does. -->
+
+This PR adds..
+
+## Related Issues
+
+<!-- Link issues that are related to this PR. You may link issues you think should be closed by this PR. -->
+
+Related #
+
+Closes #
+
+## Changes
+
+<!-- List the major changes made in this PR. -->
+
+- List changes here
+
+## Checklist
+
+I confirm that I have:
+
+- [ ] **Followed the
+      [Conventional Commits](https://www.conventionalcommits.org/)
+      specification** (e.g., `feat: add new feature`, `fix: correct bug`,
+      `docs: update documentation`).
+- [ ] **My PR title also follows the conventional commits specification.**
+- [ ] **Updated relevant documentation,** if necessary.
+- [ ] **Thoroughly tested my changes.**
+- [ ] **Added tests** (if applicable) and verified existing tests pass with
+      `make test`.
+- [ ] **Checked for breaking changes** and documented them, if any.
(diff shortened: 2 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,22 @@
+on:
+  push:
+    branches:
+      - main
```
mrcjkb, line 4: issue: the main branch of this repo is `master`

```suggestion
      - master
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,22 @@
...
+
+name: release-please
+
+jobs:
+  release-please:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: googleapis/release-please-action@v4
+        with:
+          # Default token:
```
mrcjkb, line 18: note: a personal access token is necessary for the luarocks workflow to be triggered by tags created by this workflow.

See https://github.com/nvim-neorocks/sample-luarocks-plugin?tab=readme-ov-file#generating-a-pat-personal-access-token

mrcjkb (changes requested):

.github/workflows/release.yml
ficcdaf: Sorry I'm a bit new to this -- does marking as resolved apply your suggested change automatically or will I need to update my branch?

mrcjkb:
I'm not sure, but this might need [a manifest](https://github.com/googleapis/release-please/blob/main/docs/manifest-releaser.md) to tell release-please about the initial version and which commit to bootstrap at.

.github/workflows/release.yml
ficcdaf: Is there a PAT already configured in this repo that I should reference in this file? Or is this something organization members would need to add? I'm not used to working with workflows in repos that aren't mine 😅

ficcdaf:
> I'm not sure, but this might need [a manifest](https://github.com/googleapis/release-please/blob/main/docs/manifest-releaser.md) to tell release-please about the initial version and which commit to bootstrap at.

Hmm, I believe release-please just starts with the latest release version. Ie as soon as the workflow is merged and it runs, it'll automatically process all the commits pushed since the last release and add them to the release PR; and it'll start keeping track of semver bumps from the last release onwards. So it won't try to calculate the version number from the start of the repo history, If that's what you're worried about. 

Is this what you mean by telling it the initial version? (Thanks for your patience btw)

.github/workflows/release.yml
kristijanhusak: @ficcdaf just push changes to the same branch you used to create this PR. Resolving the comment does not do anything except hide the comment.

.github/workflows/release.yml
kristijanhusak: You can use existing one `secrets.GH_TOKEN`. The same is used to push the docs https://github.com/nvim-orgmode/orgmode/blob/master/.github/workflows/docgen.yml#L17-L17

kristijanhusak (changes requested):
Generally looks good. Lets just fix these few comments and then we'll give it a try.

ficcdaf:
Ok, I added the correct github token and changed the branch name to `master`

kristijanhusak (approved):
Ok, looks good. Let see how it goes.
Thanks!

ficcdaf:
Thank you! I'm very excited to have contributed. I hope the workflow and the PR template helps to make maintaining the repo a bit easier. Thanks for your hard work on this plugin!

PR #880, opened 2025-01-30
feat(properties): add option `org_use_property_inheritance`

troiganto:
## Summary

This PR adds the option `org_use_property_inheritance` and allows using [property inheritance](https://orgmode.org/manual/Property-Inheritance.html) in searches and via the external API.

## Related Issues

Extracted from #878 

## Changes

The code to search for a property recursively was already there, it's just that the default behavior couldn't be configured.

I've changed the behavior of `get_property()`'s parameter `search_parents` from two-state (truthy=yes, falsy=no) to tri-state (`true`=yes, `false`=no, `nil`=use config).

For the config itself, I've added the option itself and also a method `OrgConfig:use_property_inheritance()`, which evaluates it for a given property name.

I've adjusted the special cases where we want to always ignore (or always use) inheritance by grepping the code base for `get_property` and evaluating them one by one. I'm reasonably sure I got all uses, but it doesn't hurt to double check.

## Checklist

I confirm that I have:

### DIFF
--- a/lua/orgmode/files/headline.lua
+++ b/lua/orgmode/files/headline.lua
@@ -422,7 +422,7 @@ end
 function Headline:set_property(name, value)
   local bufnr = self.file:get_valid_bufnr()
   if not value then
-    local existing_property, property_node = self:get_property(name)
+    local existing_property, property_node = self:get_property(name, false)
     if existing_property and property_node then
       vim.fn.deletebufline(bufnr, property_node:start() + 1)
     end
@@ -443,7 +443,7 @@ function Headline:set_property(name, value)
   end
 
   local property = (':%s: %s'):format(name, value)
-  local existing_property, property_node = self:get_property(name)
+  local existing_property, property_node = self:get_property(name, false)
   if existing_property then
     return self:_set_node_text(property_node, property)
   end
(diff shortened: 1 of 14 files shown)


kristijanhusak:
> The code to search for a property recursively was already there, it's just that the default behavior couldn't be configured.

Yes, but I forgot to handle situations where we need all properties with `get_properties()`. There are 2 places:

1. [API](https://github.com/nvim-orgmode/orgmode/blob/master/lua/orgmode/api/headline.lua#L67-L67) - We expose all properties for a headline
2. [Search](https://github.com/nvim-orgmode/orgmode/blob/master/lua/orgmode/files/file.lua#L281-L281) - This is used for matching in the tags agenda view

Since we already have inheritance with tags, there are two methods on headline class, `get_tags` and `get_own_tags`. Latter gets tags only from the headline itself.

Let's do the same thing, and update accordingly:

1. rename current `get_properties` to `get_own_properties` in the `OrgHeadline` class
2. Add new `get_properties` method that will fetch own priorities and all parents, and merge accordingly.
3. On the 2 places I mentioned above use `get_properties`
4. On every other place, rename it to `get_own_properties`

That should cover all the cases.

lua/orgmode/files/headline.lua
```diff
@@ -543,6 +577,7 @@ memoize('get_tags')
 function Headline:get_tags()
   local tags, own_tags_node = self:get_own_tags()
   if not config.org_use_tag_inheritance then
+    -- TODO: Why exclude the headline's own tags here?
```
kristijanhusak, line 580: Good catch! I think this is a bug. You can remove it.

kristijanhusak (commented):
Looks good. Let's just address that one TODO that you put, and I think that answers the question for your second TODO on the get_properties method.

kristijanhusak (approved):
Looks good, thanks!

PR #893, opened 2025-02-05
feat(utils): add optional flag to `utils.writefile()` for exclusive writes

troiganto:
## Summary

This PR adds an optional flag to `utils.writefile()` to enable exclusive writes. Such a write fails if a file with the given name already exists.

The org-attach feature requires functionality like this because it is really adamant about not overwriting attachments that the user might not be aware of. [Example here](https://github.com/emacs-mirror/emacs/blob/4da38c632161867e914b3a13dc760f8019255f94/lisp/org/org-attach.el#L559-L560).

## Related Issues

Extracted from #878 

Slowly getting there …

## Changes

- Add new optional parameter `opts` to `utils.writefile()`. Right now, the only option is a boolean flag `excl`.
- Add tests for `utils.writefile()`
- Add tests for `utils.readfile()`
- Also add a tiny test for folds; I was debugging an issue with folding earlier (turns out lazy loading was the culprit) and since there are no tests of folding at all rn, I didn't want to throw it away :slightly_smiling_face: 

## Checklist

I confirm that I have:

### DIFF
--- a/lua/orgmode/utils/init.lua
+++ b/lua/orgmode/utils/init.lua
@@ -55,10 +55,12 @@ end
 
 ---@param file string
 ---@param data string|string[]
+---@param opts? {excl?: boolean}
 ---@return OrgPromise<integer> bytes
-function utils.writefile(file, data)
+function utils.writefile(file, data, opts)
   return Promise.new(function(resolve, reject)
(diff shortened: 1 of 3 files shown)


kristijanhusak (approved):
Thanks!

PR #924, opened 2025-03-08
feat: expose store_link to agenda .

thuyen:
## Summary

As title, this PR exposes `store_link` to the Agenda API, so that one can create link to headlines while in the agenda view.

## Changes

- List changes here

## Checklist

I confirm that I have:

- [ ] **Followed the
- [ ] **My PR title also follows the conventional commits specification.**
- [ ] **Thoroughly tested my changes.**
      `make test`.

### DIFF
--- a/lua/orgmode/agenda/init.lua
+++ b/lua/orgmode/agenda/init.lua
@@ -454,6 +454,13 @@ function Agenda:set_tags()
   })
 end
 
+function Agenda:store_link()
+  return self:_remote_edit({
+    action = 'org_mappings.org_store_link',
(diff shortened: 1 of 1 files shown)


thuyen:
But without `remote_edit` we cannot add link to an un-opened file from the agenda right? `store_link_to_headline` would hit error with `get_valid_bufnr` since there's no buffer for the file. 

Probably we need to make `store_link_to_headline` works even when the file is not opened.

kristijanhusak:
Yes, I forgot about ID part. Pushed a fix for that, should be working fine now.

PR #925, opened 2025-03-08
feat: support updating a headline's cookie from multiple lists

thuyen:
## Summary

On master if a headline has multiple lists, the cookie is only updated for the current list. In emacs org, the cookie reflects information about all lists under the headline. This PR just makes it so. 

## Changes

Screenshot from 2025-03-08 06-55-34

On master the headline would show either 2/3 or 1/1 depending where the last update happens. 

## Checklist

I confirm that I have:

- [x] **Followed the
      `make test`.

### DIFF
--- a/lua/orgmode/files/elements/listitem.lua
+++ b/lua/orgmode/files/elements/listitem.lua
@@ -71,7 +71,7 @@ function Listitem:update_checkbox(action)
   else
     local parent_headline = self.file:get_closest_headline_or_nil()
     if parent_headline then
-      parent_headline:update_cookie(parent_list)
+      parent_headline:update_cookie()
     end
   end
 end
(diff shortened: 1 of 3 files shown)


seflue:
While you're on it, have you considered to also implement #305? I am really missing that feature and already thought about implementing it myself, but never found the time to wrap my head around the cookie code. Looking at your code, it seems, that it has gotten simpler over the last two years due to some refactorings, so it might be worth a shot. :slightly_smiling_face:

thuyen:
> While you're on it, have you considered to also implement #305? I am really missing that feature and already thought about implementing it myself, but never found the time to wrap my head around the cookie code. Looking at your code, it seems, that it has gotten simpler over the last two years due to some refactorings, so it might be worth a shot. 🙂

Let's see if  https://github.com/nvim-orgmode/orgmode/pull/926 works.

kristijanhusak (approved):
Thanks!

Issue #927, opened 2025-03-10
Support mkdirp tags in source block


SjB:
### Does this feature exist in Emacs orgmode core?

Yes

### Orgmode link

https://orgmode.org/manual/Extracting-Source-Code.html

### Feature value

The ‘mkdirp’ header argument creates parent directories for tangled files if the directory does not exist. A ‘yes’ value enables directory creation whereas ‘no’ inhibits it.

PR #928, opened 2025-03-10
feat(babel): add directory creation support with :mkdirp header argument

SjB:
When tangling code blocks to files, automatically create parent directories when the :mkdirp header argument is set to 'yes'.

## Summary

The ‘mkdirp’ header argument creates parent directories for tangled files if the directory does not exist. A ‘yes’ value enables directory creation whereas ‘no’ inhibits it.

https://github.com/nvim-orgmode/orgmode/issues/927


kristijanhusak (approved):
Thanks!

PR #931, opened 2025-03-14
fix(template): Use env variable USERNAME in Windows

Nekketsu:
## Summary

When creating capturing templates we can un the `%n` variable to insert the current user. It is done by using the environment variable `USER`. However, in Windows the variable is called `USERNAME`. This PR fixes this issue.

This PR adds..

## Related Issues

Related #

Closes #

## Changes

- Check whether it is being run in Windows.
- If is running in Windows it uses the environment variable `USERNAME` to query the current user name, otherwise it uses the `USER` variable.

## Checklist

I confirm that I have:

- [x] **Followed the
      `make test`.

### DIFF
--- a/lua/orgmode/capture/template/init.lua
+++ b/lua/orgmode/capture/template/init.lua
@@ -6,8 +6,6 @@ local Calendar = require('orgmode.objects.calendar')
 local Promise = require('orgmode.utils.promise')
 local Input = require('orgmode.ui.input')
 
-local is_windows = vim.loop.os_uname().sysname == 'Windows_NT'
-
 local expansions = {
   ['%%f'] = function()
     return vim.fn.expand('%')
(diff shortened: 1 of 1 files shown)


lua/orgmode/capture/template/init.lua
```diff
@@ -14,7 +16,7 @@ local expansions = {
     return vim.fn.expand('%:p')
   end,
   ['%%n'] = function()
-    return os.getenv('USER')
+    return is_windows and os.getenv('USERNAME') or os.getenv('USER')
```
kristijanhusak, line 19: Lets just do everything from inside this function like this and remove the `is_windows` variable:
```lua
if vim.fn.has('win32') == 1 then
  return os.getenv('USERNAME')
end
return os.getenv('USER')
```

kristijanhusak (changes requested):

kristijanhusak (approved):
Thanks!

PR #939, opened 2025-03-19
Support sheband and tangle mode header tag

SjB:
## Summary

Add shebang and tangle-mode source block header tags

## Related Issues

New features to enable features in org-mode

Related #

https://orgmode.org/manual/Extracting-Source-Code.html

Closes #

## Changes

Changes the structure of the tangle_info table.

### DIFF
--- a/lua/orgmode/babel/tangle.lua
+++ b/lua/orgmode/babel/tangle.lua
@@ -83,7 +83,6 @@ function Tangle:tangle()
     local shebang = info.header_args[':shebang']
     if shebang then
       shebang = shebang:gsub('[\'"]', '')
-      utils.echo_info(('shebang: %s'):format(shebang))
       table.insert(parsed_content, 1, shebang)
     end
 
--- a/lua/orgmode/babel/tangle.lua
+++ b/lua/orgmode/babel/tangle.lua
@@ -75,6 +75,11 @@ function Tangle:tangle()
       filemode = tangle_mode:gsub('[\'"]', '')
     end
 
+    if info.header_args[':mkdirp'] == 'yes' then
+      local path = vim.fn.fnamemodify(info.filename, ':h')
+      vim.fn.mkdir(path, "p")
+    end
+
     if info.name then
       block_content_by_name[info.name] = parsed_content
     end
--- a/lua/orgmode/babel/tangle.lua
+++ b/lua/orgmode/babel/tangle.lua
@@ -35,11 +35,12 @@ function Tangle:tangle()
 
   for _, info in ipairs(valid_blocks) do
     if tangle_info[info.filename] then
-      table.insert(tangle_info[info.filename], '')
+      table.insert(tangle_info[info.filename]['content'], '')
     else
-      tangle_info[info.filename] = {}
+      tangle_info[info.filename] = {content = {}}
     end
 
+    local filemode = tangle_info[info.filename]['mode']
     local do_noweb = info.header_args[':noweb'] == 'yes' or info.header_args[':noweb'] == 'tangle'
     local parsed_content = info.content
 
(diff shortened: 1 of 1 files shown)


lua/orgmode/babel/tangle.lua
```diff
@@ -60,15 +61,51 @@ function Tangle:tangle()
...
+    if info.header_args[':mkdirp'] == 'yes' then
+      local path = vim.fn.fnamemodify(info.filename, ':h')
+      vim.fn.mkdir(path, 'p')
+    end
+
+    local shebang = info.header_args[':shebang']
+    if shebang then
+      shebang = shebang:gsub('[\'"]', '')
+      table.insert(parsed_content, 1, shebang)
+    end
```
kristijanhusak, line 87: This logic is already done above on line 64. Is there a reason why it's needed twice?

lua/orgmode/babel/tangle.lua
```diff
@@ -60,15 +61,51 @@ function Tangle:tangle()
...
+
+    local tangle_mode = info.header_args[':tangle-mode']
+    if tangle_mode then
+      filemode = tangle_mode:gsub('[\'"]', '')
+    end
+
+    if info.header_args[':mkdirp'] == 'yes' then
+      local path = vim.fn.fnamemodify(info.filename, ':h')
+      vim.fn.mkdir(path, 'p')
```
kristijanhusak, line 80: Does `tangle-mode` also apply to directories? I can't find anything in the documentation.

lua/orgmode/babel/tangle.lua
```diff
@@ -60,15 +61,51 @@ function Tangle:tangle()
...
+    table.insert(
+      promises,
+      utils.writefile(filename, table.concat(self:_remove_obsolete_indent(block['content']), '\n'))
+    )
+
+    local mode_str = block['mode']
+    if mode_str and mode_str:sub(1, 1) == 'o' then
+      mode_str = mode_str:sub(2)
+      local mode_num = tonumber(mode_str, 8)
+      vim.loop.fs_chmod(filename, mode_num)
```
kristijanhusak, line 107: Instead of doing `chmod` after the file is created, lets pass down the mode as an argument to `utils.writefile` and pass it there when creating the file.

kristijanhusak (commented):
Thanks for the PR! Generally looks good, but I have some questions and suggestions.

lua/orgmode/babel/tangle.lua
SjB: sound good, I'm make the changes

lua/orgmode/babel/tangle.lua
SjB: Let me look in the orgmode source and see what they do.

SjB: I have looked into the orgmode lisp code [ob-tangle.el:315](https://github.com/bzg/org-mode/blob/b6dbf8881076191e1351d7cd15e26547a2531fea/lisp/ob-tangle.el#L315) and the make-directory does not use the tangle-mode value.

kristijanhusak: Ok, we can resolve this. Thanks for looking into it!

SjB:
The orgmode docs describe 3 styles of tangle-mode values: **ls-style**, **chmod style**, and **octal**. All three style are now supported, however, I'm still using the `vim.loop.fs_chmod` function, as I can't seem to support all the requirements with the `utils.writefile` function

kristijanhusak:
> The orgmode docs describe 3 styles of tangle-mode values: ls-style, chmod style, and octal. All three style are now supported

Nice! 👍🏻 

> however, I'm still using the vim.loop.fs_chmod function, as I can't seem to support all the requirements with the utils.writefile function

That's fine, I don't think it will cause any issues.

Could you add few tests that test these new tangle options? I'm not sure if CI will cause some issues testing the permissions, but you can add 1 test that modifies the perrmission and push to see how CI behaves.

It's somewhat sensitive stuff so I want to make sure we cover it with tests as much as possible.

PR #943, opened 2025-03-21
fix: meta_return on a link item

thuyen:
## Summary

Meta return doesn't work when the cursor is on a link because currently we don't handle link in the logic to find a headline or listitem node (the added test would be failing on master).

This PR factor out that logic into a reusable `find_item_or_headline` function and make it a bit more general. 

## Checklist

I confirm that I have:

- [x] **Followed the
      `make test`.


kristijanhusak (approved):
Thanks!

Issue #250, opened 2022-04-09
Support multiple todo keyword sequence definitions


Happy-Dude:
### Does this feature exist in Emacs orgmode core?

Yes

### Orgmode link

https://orgmode.org/guide/Multi_002dstate-Workflow.html

### Feature value

From the manual:
> Sometimes you may want to use different sets of TODO keywords in parallel. For example, you may want to have the basic ‘TODO=/=DONE’, but also a workflow for bug fixing. Your setup would then look like this:

```lisp
(setq org-todo-keywords
      '((sequence "TODO(t)" "|" "DONE(d)")
        (sequence "REPORT(r)" "BUG(b)" "KNOWNCAUSE(k)" "|" "FIXED(f)")))
```

In my specific case, I have different sequences defined for
* GTD
* Blockers
* Shopping

### Additional context

Example of how I defined my sequences in emacs:

```lisp
(setq org-todo-keywords
      '(
        (sequence "TODO" "PROCESS" "ORGANIZE" "REVIEW" "DO" "|" "DONE")                 ; GTD
        (sequence "IDEA" "DESIGN" "DEVELOP" "QA" "BACKLOG" "|" "COMPLETE")              ; Agile
        (sequence "|" "NOTDOING" "DELEGATED" "WAITING" "SOMEDAY")                       ; Blockers
        (sequence "WISHLIST" "CART" "SHIPPING" "ATTN" "|" "DONE" "RETURNED")            ; Shopping
        ))
```

Example of how I expected to define it in nvim-orgmode:
```lua
  org_todo_keywords = {
        {'TODO', 'PROCESS', 'ORGANIZE', 'REVIEW', 'DO', '|', 'DONE'},               -- GTD
        {'IDEA', 'DESIGN', 'DEVELOP', 'QA', 'BACKLOG', '|', 'COMPLETE'},            -- Agile
        {'|', 'NOTDOING', 'DELEGATED', 'WAITING', 'SOMEDAY'},                       -- Blockers
        {'WISHLIST', 'CART', 'SHIPPING', 'ATTN', '|', 'DONE', 'RETURNED'}           -- Shopping
      },
```

Happy-Dude:
If I were to attempt to implement this and submit a PR, what would the recommended workflow be / where would I reference in the code? Is there conscious effort made to re-implement the same algorithm/style as the mainline org-mode repo?

kristijanhusak:
[...]

Issue #185, opened 2021-12-14
Custom TODO keywords per file


aareman:
### Does this feature exist in Emacs orgmode core?

Yes

### Orgmode link

https://orgmode.org/manual/Per_002dfile-keywords.html

### Feature value

Could be used for a jira.org file that would behave one way, while a hiring.org file would have a different workflow

jira.org: `TODO PROGRESS QA | DONE`
hiring.org `LEAD INTERVIEW OFFER-PENDING | ACCEPTED REJECTED`

### Additional context

Note, it allows multiple workflows per file. (see the example in the ormode link) so "bug" could have one set of todo statuses and "features" could have another

jgollenz:
@aareman do you know how custom TODO keywords behave when displaying the list of all TODO entries? Are they present or do they get ignored? Same question goes for the agende when you schedule a custom TODO

aareman:
I believe it goes based on where the TODO is defined. so if its in a file with custom keywords, then it will use its custom words when switching types. and those types will show in the agenda as well.

As per my example above if a todo was in hiring.org and had a status of LEAD then in the agenda it would say LEAD. If in the agenda i went and toggled the state it would show me the choices from the file hiring.org. 

That's what i surmise from the examples i've seen, but I do not know what the internal org spec is on this is. 

An additional feature, is the ability to have multiple sets of keywords per file, and then each todo can only use one of the "status flows" (that probably means you can't have overlapping statuses in each flow)

jgollenz:
Exactly, it's individual sets of TODO-cycles and each heading can be part of one, while a file can have multiple sets. 

> that probably means you can't have overlapping statuses in each flow

The examples on the org homepage would agree with that, if the presence of a single set will override the default set of [TODO, DONE].

As for the agenda, it must then be known for each TODO item whether it belongs to a custom set or not.

jgollenz:
@aareman fyi I started working on this but haven't had time for some weeks now and don't expect to be able to seriously work on it for the next three months or so :weary: Happy to share my progress, if anybody wants to grab this

lyz-code:
@jgollenz probably the code is deprecated now, but could you share your work in progress?

seflue:
@kristijanhusak I was just looking into this.

For someone motivated to implement this, would you like to discuss some entry points?

Let me just brainstorm what we have and what we need. We have:

1. **Global Configuration**: In `config/defaults.lua`, default TODO keywords are defined as:
   ```lua
[...]

PR #956, opened 2025-04-03
feat(todos): support file-specific todo definitions

seflue:
While emacs orgmode actually supports multiple todo keyword sequences, this implementation explicitly focuses on one sequence, because the plugin currently also supports only one sequence to be defined in the configuration.

To support the full emacs functionality this needs to be extended later.

## Summary

Support to overwrite the globally defined todo keywords with a file specific one.

## Related Issues

Related #

#250 

Closes #

#185 

## Changes

- parses todo keywords from `#+todo:` when defined
- falls back to todo keywords defined in config
- when changing todo keyword state, retrieve it from the current org-file (and rely on the fallback) instead of using the config directly
- add tests to validate the new behavior

## Checklist

I confirm that I have:

- [x] **Followed the
      `make test`.

### DIFF
--- a/lua/orgmode/files/file.lua
+++ b/lua/orgmode/files/file.lua
@@ -269,6 +269,7 @@ function OrgFile:find_headline_by_title(title)
   end)
 end
 
+memoize('get_todo_keywords')
 function OrgFile:get_todo_keywords()
   local todo_directive = self:_get_directive('todo')
   if not todo_directive then
--- a/lua/orgmode/files/file.lua
+++ b/lua/orgmode/files/file.lua
@@ -276,7 +276,7 @@ function OrgFile:get_todo_keywords()
     return config:get_todo_keywords()
   end
 
-  local keywords = vim.split(todo_directive, '%s+')
+  local keywords = vim.split(vim.trim(todo_directive), '%s+')
   local todo_keywords = require('orgmode.objects.todo_keywords'):new({
     org_todo_keywords = keywords,
     org_todo_keyword_faces = config.org_todo_keyword_faces,
--- a/lua/orgmode/files/file.lua
+++ b/lua/orgmode/files/file.lua
@@ -269,6 +269,21 @@ function OrgFile:find_headline_by_title(title)
   end)
 end
 
+function OrgFile:get_todo_keywords()
+  local todo_directive = self:_get_directive('todo')
+  if not todo_directive then
+    return config:get_todo_keywords()
+  end
+
+  local keywords = vim.split(todo_directive, '%s+')
+  local todo_keywords = require('orgmode.objects.todo_keywords'):new({
+    org_todo_keywords = keywords,
+    org_todo_keyword_faces = config.org_todo_keyword_faces,
+  })
+
+  return todo_keywords
+end
+
 ---@return OrgHeadline[]
 function OrgFile:get_unfinished_todo_entries()
   if self:is_archive_file() then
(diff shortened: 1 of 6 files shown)


lua/orgmode/files/file.lua
```diff
@@ -269,6 +269,21 @@ function OrgFile:find_headline_by_title(title)
   end)
 end
 
+function OrgFile:get_todo_keywords()
```
kristijanhusak, line 272: Let's memoize this so it's not constantly recalculated:

```suggestion
memoize('get_todo_keywords')
function OrgFile:get_todo_keywords()
```

lua/orgmode/files/file.lua
```diff
@@ -269,6 +269,21 @@ function OrgFile:find_headline_by_title(title)
...
+  local todo_directive = self:_get_directive('todo')
+  if not todo_directive then
+    return config:get_todo_keywords()
+  end
+
+  local keywords = vim.split(todo_directive, '%s+')
```
kristijanhusak, line 278: Do we need to trim `todo_directive` to remove excess spaces?
```suggestion
  local keywords = vim.split(vim.trim(todo_directive), '%s+')
```

tests/plenary/ui/mappings/todo_spec.lua
```diff
@@ -451,4 +451,92 @@ describe('Todo mappings', function()
...
+
+    vim.cmd([[norm cit]])
+    assert.are.same({
+      '#+TODO: OPEN DOING | FINISHED ABORTED',
+      '* OPEN Test with file-local todo keywords',
+      '** DOING Subtask',
+    }, vim.api.nvim_buf_get_lines(0, 0, 3, false))
+  end)
+  local todos_with_shortcuts = '#+TODO: OPEN(o) DOING(d) | FINISHED(f) ABORTED(a)'
+  it('should respect file-local todo keywords with shortcut keys', function()
```
kristijanhusak, line 499: Even though this test confirms that file specific keywords are used, can we add a test that will shift around keywords and have it confirm it works?
For example, for the sake of test, have a config level settings set to `{ 'OPEN', 'DOING', '|', 'DONE'}`, and in the test file have `#+TODO: DOING DONE | OPEN`. Then test that it transitions to `OPEN` as done state and adds closed date.

kristijanhusak (changes requested):
Looks good as a starting point into this feature. There are few more usages of `config:get_todo_keywords()` across `mappings.lua` file where we can use the file level ones with `self.files:get_current_file():get_todo_keywords()`.

seflue:
@kristijanhusak I a trying to tackle a more fundamental issue, where I might need your help: The custom keywords are currently not correctly highlighted. I try to understand, how the highlighting of keywords works in general and why the custom keywords are currently ignored.

From what I learned form the code, I have the gut feeling, that the highlight definitions are loaded with the start of the plugin and to highlight the custom words correctly those highlighting informations need to be updated, when a file is loaded. Is my hypothesis correct? Do you have a suggestion how to fix the issue?

seflue:
@kristijanhusak I just found the code you mentioned and also saw, that some refactoring would be needed. Because it would be nice, if the highlighting updates on changing the `#+todo:` definition it might be a bit more complex - so I agree, it is easier to defer fixing the highlighting to a future PR.

tests/plenary/ui/mappings/todo_spec.lua
seflue: I added the test.

seflue:
> Looks good as a starting point into this feature. There are few more usages of `config:get_todo_keywords()` across `mappings.lua` file where we can use the file level ones with `self.files:get_current_file():get_todo_keywords()`.

I also integrated the feature to all mappings using `config:get_todo_keywords()` (hope I didn't miss any).

kristijanhusak (approved):
Looks good, thanks!

Issue #157, opened 2021-11-19
backslash in text creates error condition


hakling:
### Describe the bug

Text which includes a backslashed punctuation character creates an error conditions. e.g. \. \[, etc.

### Steps to reproduce

1. enter any line with text such as "this [word]" - no problem
2. put  a backslash in front of [ or .
3. results in "Error on text "This \[word]"  

### Expected behavior

Just display the backslashed character. 

### Emacs functionality

Backslashed text is displayed properly, No error is raised.

### Minimal init.lua

No changes made to default initl.lua

### OS / Distro

Ubuntu 21.04    5.11.0-40-generic #44-Ubuntu SMP

### Neovim version/commit

NVIM v0.6.0-dev+575-g2ef9d2a66

### Additional context

This may represent a known limitation, but I've not found it documented anywhere. Issue arose when taking notes on coding.

kristijanhusak:
This is coming from treesitter parser. I'll need to check what's happening there.

hakling:
Seems the issue may be a bit broader. The following line results in the same error:
'A random line with $ in it'
The issue seems to be the $ surrounded by blanks. If the $ is followed by text it is okay. e.g. $word is treated properly

hakling:
And, if the last line in a fold ends in a period, it seems not be included in the fold.

kristijanhusak:
> And, if the last line in a fold ends in a period, it seems not be included in the fold.

Can you give an example here? I can't reproduce this. Other things are reproduced.

hakling:
I'll try to recreate and let you know, but right now, I can't duplicate it.
I'm trying to convert files that I've been using on emacs and am having
regular weird issues with folding and "Error on text" without it being
clear what the issue is. I've attached a png of what I'm trying to sort out
now.

kristijanhusak:
I'm not seing an image, probably because you wrote from in-mail. Make sure to recalculate folds with `zx` before reporting any potential issues, since Vim/Neovim folding can be a bit confused at first.

hakling:
One problem I have that is reproducible in the inclusion of a standalone $
in text causes an error to be raised. Some of my org files include text
copied from a terminal session which tend to include such a  naked $. Don't
know if this is a bug, but it is a nuisance. If you like, I'll make that a
bug report.
Most of my other issues have been related to sloppy formatting which
emacs' orgmode seems to handle better.
Thanks.
Hal

kristijanhusak:
[...]

PR #974, opened 2025-05-04
feat: allow to define multiple todo keyword sequences

### LINKED ISSUE #250: Support multiple todo keyword sequence definitions

seflue:
They can be defined in the config or within an org file.

## Summary

This PR adds the ability to define multiple todo keyword sets as described in [Orgmode manual](https://orgmode.org/manual/Multiple-sets-in-one-file.html).

## Related Issues

Relates to #250, #157, PR #956

Closes #250

## Changes

- additionally to a single keyword sequence `org_todo_keywords` allows to define a table of keyword sets
- an org file can have multiple `#+TODO:` directives
- if multiple keyword sets are defined, either in the config or in the current org file, `org_todo` keymap triggers fast access mode to select a keyword
- `org_todo_prev` behaves like `S-RIGHT` in Emacs Orgmode

Falling back to fast access mode when multiple sets are defined is a bit of a shortcut to get a first version of this feature out of the door. Emacs Orgmode [defines](https://orgmode.org/manual/Multiple-sets-in-one-file.html) some additional keybindings to switch between keyword sets. This is a bit more elaborated and could be implemented in a further PR.

## Checklist

I confirm that I have:

- [x] **Followed the
      `make test`.


seflue:
@kristijanhusak It seems, that the indentation test is a bit flaky (those are the both failing test). It also failed occasionally locally on my machine, but that is unrelated to my changes.

lua/orgmode/objects/todo_keywords/init.lua
```diff
@@ -79,29 +100,105 @@ end
...
+    index = index,
+    sequence_index = seq_idx,
+  })
+
+  -- Track used shortcuts to avoid conflicts
+  if todo_keyword.has_fast_access then
+    used_shortcuts[todo_keyword.shortcut] = true
+  elseif not used_shortcuts[todo_keyword.shortcut] then
+    -- Mark it as a fast access key when we have multiple sequences
+    if type(self.org_todo_keywords[1]) == 'table' and #self.org_todo_keywords > 1 then
```
kristijanhusak, line 138: Having multiple sequences doesn't necessarily mean we use fast access. When there is no fast access defined in the config of the todo keywords, Emacs just cycles through the sequence it figured out that is being used.

There's a catch though. Emacs somehow keeps in memory which was the last sequence used. For example, if you have these keywords:

```lisp
(setq org-todo-keywords
      '((sequence "TODO" "|" "DONE")
        (sequence "REPORT" "BUG" "TESTING" "|" "FIXED")))
```

And you have a headline with `* REPORT foo`, cycling without the fast access will go `REPORT -> BUG -> TESTING -> FIXED -> {EMPTY} -> REPORT -> etc.`. So it keeps in memory that the last sequence is 2nd one. Now, if you switch to an empty headline without todo keyword, restart emacs, and try switching the todo state, it will use the first sequence. So it's only in memory while the emacs is open.

I'm not sure what's the best way to access that issue, but I wanted to bring it up in case you have some ideas.

kristijanhusak (changes requested):
I left few comments around the code, some around code styling changes (mostly around avoiding `else` when there's early return), and some comments around the logic.

I think we should approach this in a slightly different , and potentially simpler way:

1. Always store the todo keywords as sequences. You already did that here more-less, but there's no need have a separate logic like `parse_single_sequence` or `parse_multiple_sequence`. When `TodoKeywords` get's `org_todo_keywords` value, just check if it's a `string[]` or `string[][]`. If it's former, just convert it to `string[][]` and handle it accordingly. 
2. Instead of keeping the sequence index on the `TodoKeyword`, and `sequences` on the `TodoKeywords`, we can just search things on the fly. These things are used only when mutating the document, so it's not that necessary to do these optimizations. Lua is fast enough to handle this, and I doubt users have a lot of todo sequences. 

Another approach could be to keep the `TodoKeywords` as they are, which is basically as single `sequence`, and have a layer above that will know to figure out which of the sequence we need at the given point, and just return it and use it where necessary. 
This might complicate some other things so it might not be the best idea, but I wanted to point it out.

Let me know what you think.

seflue:
@kristijanhusak Thanks for the feedback! I've implemented most of your suggestions:

**Simplified parsing and code style:**
- Always store todo keywords as sequences (string[][]) - normalized in constructor 
- Removed separate parsing methods, single unified _parse() path
- Fixed unnecessary else blocks after early returns
- Cleaned up TodoState constructor

**Regarding the sequence index approach:**
I'd prefer to keep the current approach here. The sequence_index isn't just for performance - it's needed for proper cycling behavior. According to the Emacs orgmode manual, C-c C-t should cycle only within the current sequence. So a TODO in:

```
["TODO", "|", "DONE"]
```

cycles `TODO → DONE → empty → TODO`, while a REPORT in:

```  
["REPORT", "BUG", "|", "FIXED"]
```

stays within that sequence.

Without sequence_index, we can't determine which sequence a keyword belongs to, especially when keywords are duplicated across sequences (like having "TODO" in both bug and feature workflows).

The next step would be implementing proper cycling behavior instead of jumping to fast access mode in a separate PR, which is where this becomes important.

What do you think?

seflue:
@kristijanhusak Sorry for the long delay, I just didn't found the time to get back to this PR in the last month.

I reverted the refactoring as you requested, hopefully it is now ready to merge.

kristijanhusak (approved):
I think this is good to go now. I'll do some minor changes around building the TodoKeywords, but nothing major.
Next up we should remove the fast access if not declared, and document this.

Thanks for the PR and the patience!

Issue #977, opened 2025-05-10
blink.cmp removes #+prefix when completing org directives


seflue:
## Bug Description

When using autocompletion for org directives with blink.cmp, the `#+` prefix is incorrectly removed from completed directives.

## Steps to Reproduce

1. Type `#+fi` in an org file
2. Trigger completion with blink.cmp
3. Select `#+filetype` from the completion menu

## Expected Behavior

The completion should insert `#+filetype`

## Actual Behavior

The completion inserts just `filetype` (the `#+` prefix is removed)

## Additional Info

- This only happens when both `#` and `+` are typed
- Typing just `#` works correctly
- didn't occur with nvim-cmp
- nvim-orgmode version: latest
- blink.cmp version: 1.2.0
- Neovim version: 0.11.1

PR #978, opened 2025-05-10
fix: preserve prefix in blink.cmp org directive completions

seflue:
## Summary

This PR fixes a bug where blink.cmp autocompletion incorrectly removes the `#+` prefix when completing org directives. When typing `#+fi` and selecting `#+filetype` from the completion menu, the completion would incorrectly insert just `filetype` instead of `#+filetype`.

## Related Issues

This PR addresses the bug described in my bug report #977 .

## Changes

- Fixed the `get_completions` function in `lua/orgmode/org/autocompletion/blink.lua` to preserve the `#+` prefix for org directives
- Added a test case in `tests/plenary/org/autocompletion_spec.lua` to verify that `#+` prefixes are preserved when completing org directives
- The fix ensures that other completion types (tags, headlines, etc.) continue to work as before while only affecting org directive completions

## Checklist

I confirm that I have:

- [x] **Followed the
      `make test`.

### DIFF
--- a/lua/orgmode/org/autocompletion/blink.lua
+++ b/lua/orgmode/org/autocompletion/blink.lua
@@ -41,6 +41,9 @@ function Source:get_completions(ctx, callback)
   local triggers = { '#', '+', ':', '*', '/' }
 
   local getInsertTextOffset = function(word)
+    if #word > 1 and word:sub(1, 2) == '#+' then
+      return 0
+    end
     local word_length = #word + 1
     while word_length > 0 do
       local char = word:sub(word_length - 1, word_length - 1)
--- a/tests/plenary/org/autocompletion_spec.lua
+++ b/tests/plenary/org/autocompletion_spec.lua
@@ -382,3 +382,64 @@ describe('Autocompletion', function()
     end)
   end)
 end)
+
+describe('Blink completion', function()
+  after_each(function()
+    vim.cmd([[silent! %bw!]])
+  end)
+
+  it('should preserve `#+` prefix when completing directives', function()
+    helpers.create_file({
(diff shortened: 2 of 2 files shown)


lua/orgmode/org/autocompletion/blink.lua
```diff
@@ -60,7 +60,8 @@ function Source:get_completions(ctx, callback)
   for _, item in ipairs(results) do
     table.insert(items, {
       label = item.word,
-      insertText = insertTextOffset > 0 and item.word:sub(insertTextOffset) or item.word,
+      insertText = (item.word:match('^#%+') and item.word .. ':')
```
kristijanhusak, line 63: Instead of doing it inline here, lets do an early return in `getInsertTextOffset` to return 0 if it's a directive. Something like this:

```lua
  local getInsertTextOffset = function(word)
    -- Offset should not apply to directives
    if #word > 1 and word:sub(1, 2) == '#+' then
      return 0
    end
    local word_length = #word + 1
    while word_length > 0 do
      local char = word:sub(word_length - 1, word_length - 1)
      if vim.tbl_contains(triggers, char) or char:match('%s') then
        return word_length
      end
      word_length = word_length - 1
    end
    return 0
  end

```

kristijanhusak (changes requested):
Works great, but lets move logic around a bit.

lua/orgmode/org/autocompletion/blink.lua
seflue: Agreed. I doubled the logic unnecessarily.

kristijanhusak (approved):
Thanks!

PR #1002, opened 2025-06-27
feat(edit_src): Add `org_edit_src_filetype_map` for custom src block filetypes.

milanglacier:
We have implemented a hard-coded filetype map for common language aliases used in Org mode source blocks.

I propose making this map a user-configurable option. This would empower users to extend it with their own aliases.

A use case involves Emacs users who employ `emacs-jupyter`. This package uses the language alias `jupyter-python` for Python source blocks, which differs from the standard `python` alias. User-configurable aliases would seamlessly support such workflows.

### DIFF
--- a/lua/orgmode/utils/init.lua
+++ b/lua/orgmode/utils/init.lua
@@ -564,6 +564,7 @@ end
 ---@param skip_ftmatch? boolean
 ---@return string
 function utils.detect_filetype(name, skip_ftmatch)
+  local config = require('orgmode.config')
   local map = {
     ['emacs-lisp'] = 'lisp',
     elisp = 'lisp',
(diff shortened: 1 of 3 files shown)


lua/orgmode/utils/init.lua
```diff
@@ -576,6 +577,7 @@ function utils.detect_filetype(name, skip_ftmatch)
     shell = 'bash',
     uxn = 'uxntal',
   }
+  map = vim.tbl_deep_extend('force', map, config.org_edit_src_filetype_map)
```
kristijanhusak, line 580: Let's move this near the end of the function before the last `return`. Also, please switch up a logic like this:

```lua
  if config.org_edit_src_filetype_map[name] then
    return config.org_edit_src_filetype_map[name]
  end

  return name:lower()
```

This function is called a lot through a custom directive when rendering blocks, so i'd rather keep it as optimized as possible.

kristijanhusak (changes requested):
I'm ok with the feature, but lets switch up logic a bit.

milanglacier:
Ok, I have updated the patch as you described. Please let me know if you have any other questions.

kristijanhusak (approved):
Thanks!

PR #1010, opened 2025-07-06
fix: fuzzy autocompletion with blink

seflue:
## Summary

This PR fixes fuzzy matching for blink.cmp in orgmode completion. Previously blink.cmp only did prefix matching like nvim-cmp, now it
actually uses its fuzzy matcher.

## Related Issues

Related #977 

## Changes

- Let blink.cmp get all completion results instead of pre-filtering them
- Added `filterText` field so blink.cmp knows what to fuzzy match against
- Fixed base extraction to not break completion context

## Checklist

I confirm that I have:

- [x] **Followed the
      `make test`.


kristijanhusak:
Unfortunately it's not that simple. The substring matching goes all the way down to the sources. It needs to be refactored so the context can be passed down, and internal sources can rely on the context value. I'll see if I can figure out something out at least for testing.

Issue #899, opened 2025-02-07
Folding is not respected when using `org_move_subtree_up` or `org_move_subtree_down`


lyz-code:
### Describe the bug

When moving headlines with those bindings the foldings are not respected

### Steps to reproduce

Using the next file as a base:

```
* heading 1
** subheading 2
** subheading 3
*** subsubheading 1
    text
**** subsubsubheading1
    text
```

- If you unfold the first level and use `org_move_subtree_up` on the `subheading 3`, it will move the headline but it will fold everything so that you only see `heading 1`
- If you show the folds below `subheading 3` (but not enough to see `subsubsubheading1`) and use `org_move_subtree_down`, then all folds are open:

![Image](https://github.com/user-attachments/assets/d47e8562-e07c-4f91-b864-ff8c70cb4b8d)

### Expected behavior

Folds are respected when moving headlines around

### Minimal init.lua

```lua
local tmp_dir = vim.env.TMPDIR or vim.env.TMP or vim.env.TEMP or "/tmp"
local nvim_root = tmp_dir .. "/nvim_orgmode"
local lazy_root = nvim_root .. "/lazy"
local lazypath = lazy_root .. "/lazy.nvim"

for _, name in ipairs({ "config", "data", "state", "cache" }) do
	vim.env[("XDG_%s_HOME"):format(name:upper())] = nvim_root .. "/" .. name
end

-- Install lazy.nvim if not already installed
if not vim.loop.fs_stat(lazypath) then
	vim.fn.system({
		"git",
		"clone",
		"--filter=blob:none",
		"https://github.com/folke/lazy.nvim.git",
		"--branch=stable", -- latest stable release
		lazypath,
	})
end
vim.opt.rtp:prepend(lazypath)

require("lazy").setup({
	{
		"nvim-orgmode/orgmode",
		event = "VeryLazy",
		ft = { "org" },
		config = function()
			require("orgmode").setup({
				org_agenda_files = {
					"*.org",
				},
			})
		end,
	},
}, {
	root = lazy_root,
	lockfile = nvim_root .. "/lazy.json",
	install = {
		missing = false,
	},
})

require("lazy").sync({
	wait = true,
	show = false,
})

```

### OS / Distro

Linux

### Neovim version/commit

0.10.4

kristijanhusak:
Does it set up folds correctly after you recalculate them after with `zx` ?
This is a common Neovim issue. When Things are moved around, recalculating folds is often necessary.

[...]

PR #1013, opened 2025-07-12
fix: avoid E490 when using `move_subtree_up` and `move_subtree_down`

MarkSverdlov:
## Summary
This PR contains two commits - the first fix inconsequentaial probable type regarding the cursor positioning after calling `move_subtree_up` and `move_subtree_down`. The second add additional condition to the end of these functions to ensure E490 Fold not found is not raised.

## Related Issues

Related #899 - I believe this issue is caused by the same underlying reason, of nvim failing to understand the updated fold structure after the rearranging of lines.

### commit MarkSverdlov/orgmode@5bfa7f3
Patch a bug where E490: No fold found is raised when executing `move_subtree_up` or `move_subtree_down` in some scenarios. A minimal example is:
```
* a
* b
** c
```
In that case, if `foldlevel=0` (i.e. `** c` isn't visible) and the cursor is on `* b`, then using `<leader>oK` may raise E490. This doesn't happen on the first execution of this function after loading the buffer, but does happen consistently on the next executions of the function.

The adds additional condition for the last part of these functions (the parts that conditionally calls `vim.fn.cmd([[norm!zc]])`), that checks whether the fold exists. This guarantees the problematic instruction won't be called in these scenarios.

It's important to note that this patch doesn't solve the underlying bug - the conditioned suite shouldn't be executed to begin with unless `foldclosed > -1`, which suggests that before the rearranging of lines, the predecessor of the offending line had a positive `foldlevel` (and the offending line should have too), but since nvim doesn't agree with it, this patch at least circumvent the red error message.

### Commit MarkSverdlov/orgmode@fdcfdca
This changes the column coordinate of the placement of the cursor after the rearranging of lines. Before, the column was `pos[2]`, which is the *line* number of the previous location. Now, the column coordinate is set to be `pos[3]` which is the *column* number of the previous position. This change is quite inconsequential, but it's an easy fix.

## Checklist

I confirm that I have:

- [x] **Followed the
      `make test`.

### DIFF
--- a/README.org
+++ b/README.org
@@ -198,5 +198,5 @@ issue.
 
 - [[https://github.com/dhruvasagar][@dhruvasagar]] and his [[https://github.com/dhruvasagar/vim-dotoo][vim-dotoo]] plugin
   that got me started using orgmode. Without him this plugin would not happen.
-- [[https://github.com/milisims][@milisims]] for writing a treesitter parser for org
+- [[https://github.com/emiasims][@emiasims]] for writing a treesitter parser for org
 - [[https://github.com/jceb/vim-orgmode][vim-orgmode]] for some parts of the code (mostly syntax)
--- a/docs/contributing.org
+++ b/docs/contributing.org
@@ -98,7 +98,7 @@ can be installed, and exit with a success message.
 :PROPERTIES:
 :CUSTOM_ID: parser
 :END:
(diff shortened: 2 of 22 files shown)


kristijanhusak (approved):
Thanks!

PR #1017, opened 2025-07-31
Feat: Making virtual indentation indent wrapped lines

freddrH:
## Summary

Personally dependent on emacs orgmode style where lines represent paragraphs of written text. Built this on existing virtual indent func in nvim orgmode. 

This only adds support for wrapping lines on space (" ")

## Related Issues

https://github.com/nvim-orgmode/orgmode/issues/422

## Changes

- Changed the way indentations is created by looping over the actual strings and adding virtual spaces to end lines and add indentation.
- Removed vim.schedule from attach() on_lines, since it caused extreme jitter on text below cursor when extmarks are added and removed. 
 
## Checklist

I confirm that I have:
 - Tried my best. First lua code ever and not a programmer by profession :). Happy for feedback!

### DIFF
--- a/lua/orgmode/ui/virtual_indent.lua
+++ b/lua/orgmode/ui/virtual_indent.lua
@@ -91,7 +91,7 @@ function VirtualIndent:_get_indent_size(line, tree_has_errors)
   return 0
 end
 
-function VirtualIndent:_set_wrappoints_of_luastring(line, line_str, indent, wrap_col)
+function VirtualIndent:_set_wrappoints_of_luastring(line_nr, line_str, indent, wrap_col)
   local function update_exmarks(wrap_arr)
     local function set_extmarks(curr_line, pos, nr_spaces)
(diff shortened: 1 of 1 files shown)


kristijanhusak:
Can you clarify with an example what exactly you attempted to do here? I don't understand what you mean by "wrapping lines on space".

freddrH:
Meant to push to my fork where i'm working on it. But there's working code. 
Vim linebreak wraps words to the next line on a lot of characters other than than space. The code I've written only uses 'space' to separate words (for simplicity and fewer iterations). 

In the current virtual indentation I've only been able to get the first row of text to indent and long lines wrap and continue in col 1. 

Been trying to mimic this kind of indentation in emacs: 
linebreak with indent

Got it working now but needs additions to allow configuration to choose between both methods. Going use it at work (instead of emacs) for a while to see if i find bugs.

nvim org virt indent

PR #1024, opened 2025-08-12
feat(autocomplete): Fuzzy match autocompletion

kristijanhusak:
## Summary

Add support for fuzzy matching in autocompletion.

## Related Issues

Related https://github.com/nvim-orgmode/orgmode/pull/1010

## Changes

- Both blink.cmp and nvim-cmp now do fuzzy matching by default
- Ominfunc does fuzzy matching only if `completeopt` has `fuzzy`

## Checklist

I confirm that I have:

- [ ] **Followed the
- [ ] **My PR title also follows the conventional commits specification.**
- [ ] **Thoroughly tested my changes.**
      `make test`.


seflue:
Thanks @kristijanhusak for the response. I am giving your PR a try and get back to you as soon as I have some feedback.

I just skimmed through the code but didn't fully understand, in which way #1010 broke things and how your code avoided it. What I think I understood is, that you now provide fuzzy matching using vim.fn.matchfuzzy, circumventing the fuzzy completion of blink itself.

kristijanhusak:
@seflue the simplified base that is provided is not enough context for some of the autocompletion, most specifically the "target" part of the link completion. For example, if you type this with a valid path to org file:

```
[[file:/path/to/my/org/todos.org::*
```

You will get list of all headlines in that file as autocompletion. With your solution, the file context is never there for the autocompletion to provide that list. I solved it by providing the whole base value down, but the matching in the lower layers use the fuzzy matching. Previously it was doing substring matching everywhere.

> What I think I understood is, that you now provide fuzzy matching using vim.fn.matchfuzzy, circumventing the fuzzy completion of blink itself

That is correct. It's not ideal solution, but this way we provide fuzzy completion for everything, and implementation is not tied to blink.cmp. It would need a lot of changes to adapt to each framework, and I'd like to avoid that.

PR #1041, opened 2025-10-03
fix(files): remove string tree-sitter parser

kristijanhusak:
## Summary

With the change on nightly https://github.com/neovim/neovim/pull/35988, string parser creates a scratch buffer under the hood, and uses that as a source. This drastically decreases the performance of the current code.

This commit refactors the code to remove the string parser, and adds own loading of the org buffers for the tree-sitter usage. Performance should be same as before, with the downside that all of the org files that are in org_agenda_files are now loaded as unlisted buffers.

## Changes

- Remove usage of [get_string_parser](https://neovim.io/doc/user/treesitter.html#vim.treesitter.get_string_parser())
- Manually add and load all org agenda files as unlisted buffers, so they can be parsed with tree-sitter
- Remove some of the code and tests that are no longer relevant

## Checklist

I confirm that I have:

- [X] **Followed the
      `make test`.

### DIFF
--- a/lua/orgmode/colors/highlighter/markup/init.lua
+++ b/lua/orgmode/colors/highlighter/markup/init.lua
@@ -169,7 +169,7 @@ end
 ---@return OrgMarkupPreparedHighlight[]
 function OrgMarkup:get_prepared_headline_highlights(headline)
   local highlights =
-    self:get_node_highlights(headline:node(), headline.file:get_source(), select(1, headline:node():range()))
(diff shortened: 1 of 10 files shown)


kristijanhusak:
@chipsenkbeil just a heads-up about this change. I don't know if it affects org-roam.

ribru17:
I think there may be a simpler way to prevent this perf regression; i think this was less about using a scratch buffer and more about parser:source() always returning a number now. I'm about to be afk but i will do some investigating

Issue #911, opened 2025-02-27
`org_refile` can't refile headings


cezarmathe:
### Describe the bug

I'm trying to move a heading from `my-org-file.org` to `refile.org`. Both files are located in `${HOME}/org`. When I do `<Leader>or` the autocomplete omits a few characters from the start of each file, e.g. `refile.org` is shown as `efile.org`. Upon selecting any file or typing the correct file myself I get the following error: `"nil" is not a file specified in the "org_agenda_files" setting. Refiling canceled.`.

### Steps to reproduce

1. Create a new file: `refile.org`.
2. Create a new file: `2025-02-27.org`.
3. In `2025-02-27.org`: `<Leader>oit` and add `Hello, world!`.
4. In `2025-02-27.org`: `<Leader>or` on the header you've added previously and type `refile.org`.

### Expected behavior

I expect that `<Leader>or` moves the heading to the desired destination.

### Minimal init.lua

```lua
-- You can add your own plugins here or in other files in this directory!
--  I promise not to create any merge conflicts in this directory :)
--
-- See the kickstart.nvim README for more information
return {
  'nvim-orgmode/orgmode',
  event = 'VeryLazy',
  ft = { 'org' },
  config = function()
    -- Setup orgmode
    require('orgmode').setup {
      org_agenda_files = '~/org/**/*',
      org_default_notes_file = '~/org/refile.org',
    }

    -- NOTE: If you are using nvim-treesitter with ~ensure_installed = "all"~ option
    -- add ~org~ to ignore_install
    -- require('nvim-treesitter.configs').setup({
    --   ensure_installed = 'all',
    --   ignore_install = { 'org' },
    -- })
  end,
}
```

### Screenshots and recordings

https://github.com/user-attachments/assets/8599d067-ed1e-48fe-b9b6-37603360a067

### OS / Distro

MacOS 15.3.1

### Neovim version/commit

NVIM v0.10.4 / Build type: Release / LuaJIT 2.1.1736781742

kristijanhusak:
The error is due to the missing `/`, I'll address that, but the autocompletion is strange. I'm not able to reproduce it, so I'll need some assistance from your side to figure out what's the issue. I think I know where it is, but I cannot reproduce.

Are you able to apply this patch to your local orgmode installation and give me the message output? These are the steps:
1. Apply this patch:
```diff
diff --git a/lua/orgmode/utils/fs.lua b/lua/orgmode/utils/fs.lua
[...]

PR #1042, opened 2025-10-03
fix(windows): handle Windows file paths correctly in refile

allhearingeye:
## Summary

There was a bug on Windows where the first character of the file name was dropped during refile. This happened because an extra leading / was added to the beginning of the path, resulting in something like: `/C:/org/file.org` which was then displayed as: `ile.org`.

Some issue was reported in https://github.com/nvim-orgmode/orgmode/issues/911, but not fixed on Windows.

## Changes

Added OS check and fixed path construction

## Checklist

I confirm that I have:

- [x] **Followed the
      `make test`.

### DIFF
--- a/lua/orgmode/utils/fs.lua
+++ b/lua/orgmode/utils/fs.lua
@@ -111,7 +111,12 @@ function M.trim_common_root(paths)
     end, paths)
   end
 
-  local root = '/' .. table.concat(common_root, '/') .. '/'
+  local root
+  if vim.fn.has('win32') == 1 then
+    root = table.concat(common_root, '/') .. '/'
+  else
+    root = '/' .. table.concat(common_root, '/') .. '/'
+  end
   local result = {}
   for _, path in ipairs(paths) do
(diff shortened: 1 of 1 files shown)


lua/orgmode/utils/fs.lua
```diff
@@ -111,7 +111,12 @@ function M.trim_common_root(paths)
     end, paths)
   end
 
-  local root = '/' .. table.concat(common_root, '/') .. '/'
+  local root
+  if vim.fn.has('win32') == 1 then
+    root = table.concat(common_root, '/') .. '/'
+  else
+    root = '/' .. table.concat(common_root, '/') .. '/'
+  end
```
kristijanhusak, line 119: We can simplify it a bit. Only difference is that on non-Windows we need `/` prefix.
```suggestion
  local root = table.concat(common_root, '/') .. '/'
  if vim.fn.has('win32') == 0 then
    root = '/' .. root
  end
```

kristijanhusak (approved):
Thanks!

PR #1076, opened 2025-12-29
fix(files): handle stale treesitter nodes gracefully

seflue:
## Summary

This PR fixes intermittent "Index out of bounds" errors when treesitter nodes become stale after buffer modifications.

The error occurs when:
1. Buffer is parsed, nodes are cached
2. Buffer content changes (user edits)
3. Cached nodes are accessed via async callback (e.g., `vim.schedule`)
4. Node positions now point to invalid ranges → crash

Example error:

```
Error executing vim.schedule lua callback: .../treesitter.lua:195: Index out of bounds
stack traceback:
        [C]: in function 'nvim_buf_get_text'
        .../treesitter.lua:195: in function 'get_node_text'
        .../orgmode/files/headline.lua:545: in function 'get_property'
```

## Changes

- Wrap `get_node_text()` in `pcall` to gracefully handle stale nodes
- Return empty string on failure instead of crashing
- Add `_parse_tick` field to track buffer state at parse time
- Add `OrgFile:is_tree_stale()` for callers that need explicit staleness checks

## Checklist

I confirm that I have:

- [ ] **Added tests** (if applicable) and verified existing tests pass with `make test`.


lua/orgmode/files/file.lua
```diff
@@ -193,9 +195,30 @@ function OrgFile:parse(skip_if_not_modified)
...
+    self._parse_tick = vim.api.nvim_buf_get_changedtick(bufnr)
+  end
+
   return self.root
 end
 
+---Check if the parsed tree is stale (buffer changed since last parse)
+---Useful for async callers to detect when node positions may be invalid
+---@return boolean true if tree is stale and should be re-parsed
+function OrgFile:is_tree_stale()
```
kristijanhusak: Where is this used?

lua/orgmode/files/file.lua
```diff
@@ -30,6 +30,8 @@ local is_nightly = vim.fn.has('nvim-0.12') > 0
 ---@field metadata OrgFileMetadata
 ---@field parser vim.treesitter.LanguageTree
 ---@field root TSNode
+---@field _load_timing? { since_last_ms: number, since_start_ms: number } Progressive loading timing data
```
kristijanhusak: This seems like a leftover from something, I don't see it being defined or used.

kristijanhusak (commented):
Thanks for the PR! It seems unfinished though.

lua/orgmode/files/file.lua
seflue: This was actually refactord and inlined and then I forget to remove the original function. Fixed.

lua/orgmode/files/file.lua
seflue: Rebasing artifact, should now be gone.

lua/orgmode/files/file.lua
```diff
@@ -193,6 +194,13 @@ function OrgFile:parse(skip_if_not_modified)
   self.parser = self:_get_parser()
   local trees = self.parser:parse()
   self.root = trees[1]:root()
+
+  -- Track changedtick for staleness detection
+  local bufnr = self:bufnr()
+  if bufnr > -1 then
+    self._parse_tick = vim.api.nvim_buf_get_changedtick(bufnr)
```
kristijanhusak: What is this used for?

lua/orgmode/files/file.lua
```diff
@@ -32,9 +32,17 @@ local clean_empty_line = vim.fn.has('nvim-0.13') == 1 or vim.fn.has('nvim-0.12.3
...
+---@field _parse_tick? number Buffer changedtick at last parse
 local OrgFile = {}
 
+-- The cache is keyed by the root node id, so a stale root would hand back
+-- nodes for a buffer that no longer matches them. A buffer edited since the
+-- last parse gets parsed first, which moves the key to the fresh tree.
 local memoize = Memoize:new(OrgFile, function(self)
+  local bufnr = self:bufnr()
+  if bufnr > -1 and self._parse_tick ~= vim.api.nvim_buf_get_changedtick(bufnr) then
+    self:parse()
```
kristijanhusak: No need to parse here, the method that is being called should do the parsing when it's really called. We just need to make sure that the cache is invalidated. We can append the changedtick to the id, and that does the trick. This works for me:

```lua
local memoize = Memoize:new(OrgFile, function(self)
  local tick = 0
  local bufnr = self:bufnr()
  if bufnr and bufnr > -1 then
    tick = vim.api.nvim_buf_get_changedtick(bufnr)
  end
  return {
    file = self,
    id = table.concat({ 'file', self.root and self.root:id() or '', tick }, '_'),
  }
end)
```

We also don't need to track the `_parse_tick` any more.

kristijanhusak (changes requested):
Thanks for the PR, and sorry for the delay. 
I left a comment on the logic, we can simplify it a bit.

lua/orgmode/files/file.lua
seflue: Done.

kristijanhusak (approved):
Thanks!

PR #1077, opened 2025-12-29
perf(foldtext): cache fold highlights by state

seflue:
## Summary

Navigation over folded sections in larger org files felt sluggish due to repeated treesitter queries on every cursor move.

This PR optimizes foldtext highlighting by checking fold state before using cached values. Cache hit when fold state unchanged, full recompute only when folds toggle.

**Before:** `foldtext.on_line` was the top profiler hotspot
**After:** No longer appears in profiler results

Navigation feels now much snappier.

## Changes

- Check `foldclosed()` before using cached highlights (very cheap)
- Cache hit when fold state unchanged, skip expensive `treesitter.get_captures_at_pos()`
- Full recompute only when fold actually toggles
- Remove unused `utils` import

## Checklist

I confirm that I have:

- [ ] **Added tests** (if applicable) and verified existing tests pass with `make test`.

### DIFF
--- a/tests/plenary/colors/foldtext_spec.lua
+++ b/tests/plenary/colors/foldtext_spec.lua
@@ -1,0 +1,398 @@
+local helpers = require('tests.plenary.helpers')
+local config = require('orgmode.config')
+local org = require('orgmode')
+local api = vim.api
+
+describe('foldtext highlighter', function()
+  local ns_id = api.nvim_create_namespace('org_custom_highlighter')
+
+  local function setup_file_with_folds(content)
+    -- Enable colored folds so foldtext extmarks are used
+    config.ui.folds.colored = true
+    helpers.create_file(content)
+    -- Disable ephemeral mode so extmarks persist and are queryable
+    -- Must be set after create_file initializes the highlighter
+    ---@diagnostic disable-next-line: invisible
+    org.highlighter._ephemeral = false
+  end
+
+  ---Get the column position of the foldtext ellipsis extmark on a line.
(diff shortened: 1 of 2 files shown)


lua/orgmode/colors/highlighter/init.lua
```diff
@@ -81,25 +81,27 @@ function OrgHighlighter:_on_win(_, win, bufnr, topline, botline)
...
 ---@param bufnr number
 ---@param line number
 ---@param use_cache? boolean
-function OrgHighlighter:_on_line_impl(bufnr, line, use_cache)
+function OrgHighlighter:_on_line_impl(bufnr, line, use_cache, winid)
   if self.buffers[bufnr].tree then
     self.markup:on_line(bufnr, line, self.buffers[bufnr].tree, use_cache)
     self.stars:on_line(bufnr, line)
-    self.foldtext:on_line(bufnr, line)
+    if require('orgmode.config').ui.folds.colored then
```
kristijanhusak: Let's remove the check here and return early inside the `foldtext:on_line` since `OrgFoldText` already has `config` imported.

kristijanhusak (changes requested):
Thanks for the PR! Looks great! Let's just move the check as suggested and it's good to go.

lua/orgmode/colors/highlighter/init.lua
seflue: Done.

kristijanhusak (approved):
Thanks!

PR #1078, opened 2025-12-30
fix(folds): update fold boundaries after headline edits

seflue:
## Summary

Fold boundaries go stale when editing folded headlines (TODO state, priority, tags, archive). Neovim's treesitter foldexpr only re-evaluates fold levels for the changed line, not the full section, so the fold end boundary is never updated.

Two commits:

1. **Foldtext cache invalidation**: Track `changedtick` per buffer and clear the foldtext highlight cache when content changes. Fixes stale ellipsis positions.

2. **Targeted fold update**: Capture the section range before the edit (TS nodes become stale after buffer changes), then schedule `vim._foldupdate()` for the full section after treesitter reparses. Covers TODO state changes, priority, tag and archive tag edits.

## What changed since the initial version

Dropped the `zx` approach entirely. It recomputed all folds in the buffer, resetting manually opened folds (as @kristijanhusak pointed out). Now using `vim._foldupdate(win, start, end)` to update only the affected section.

## Checklist

- [x] **Followed the [Conventional Commits](https://www.conventionalcommits.org/) specification**

### DIFF
--- a/lua/orgmode/org/mappings.lua
+++ b/lua/orgmode/org/mappings.lua
@@ -17,6 +17,25 @@ local Promise = require('orgmode.utils.promise')
 local Input = require('orgmode.ui.input')
 local Footnote = require('orgmode.objects.footnote')
 
+---Schedule a fold update for the given range. Call after buffer edits.
+---OrgRange is 1-indexed; vim._foldupdate expects 0-indexed lines.
+---@param range OrgRange
+local function schedule_fold_update(range)
+  local bufnr = vim.api.nvim_get_current_buf()
+  vim.schedule(function()
+    if not vim.api.nvim_buf_is_valid(bufnr) then
+      return
+    end
+    local start_line = range.start_line - 1
+    local end_line = math.min(range.end_line, vim.api.nvim_buf_line_count(bufnr))
+    for _, win in ipairs(vim.fn.win_findbuf(bufnr)) do
+      if vim.wo[win].foldmethod == 'expr' then
+        vim._foldupdate(win, start_line, end_line)
+      end
+    end
+  end)
+end
+
 ---@class OrgMappings
 ---@field capture OrgCapture
 ---@field agenda OrgAgenda
@@ -50,6 +69,8 @@ function OrgMappings:set_tags(tags)
   local headline = self.files:get_closest_headline()
   local headline_tags = headline:get_own_tags()
   local current_tags = utils.tags_to_string(headline_tags)
+  -- Capture range before promise chain — TS nodes become stale after edits
+  local range = headline:get_range()
 
   return Promise.resolve()
     :next(function()
@@ -69,14 +90,17 @@ function OrgMappings:set_tags(tags)
         return
       end
 
-      return headline:set_tags(new_tags)
+      headline:set_tags(new_tags)
+      schedule_fold_update(range)
     end)
 end
 
 ---@return nil
 function OrgMappings:toggle_archive_tag()
   local headline = self.files:get_closest_headline()
+  local range = headline:get_range()
   headline:toggle_tag('ARCHIVE')
+  schedule_fold_update(range)
 end
 
 function OrgMappings:cycle()
@@ -344,7 +368,9 @@ function OrgMappings:set_priority(direction)
     end
   end
 
+  local range = headline:get_range()
   headline:set_priority(new_priority)
+  schedule_fold_update(range)
 end
 
 function OrgMappings:todo_next_state()
@@ -419,6 +445,9 @@ function OrgMappings:_todo_change_state(direction)
   local headline = self.files:get_closest_headline()
   local old_state = headline:get_todo()
   local was_done = headline:is_done()
+
+  local range = headline:get_range()
+
   local changed = self:_change_todo_state(direction, true)
 
   if not changed then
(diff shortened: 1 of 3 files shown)


kristijanhusak:
Thanks for the PR! Even though this fixes that specific issue, I think it might introduce some other issues.
`zx` affects the whole buffer, and if someone opened few folds around the file, and does a todo change on one thing, everything gets recomputed to the current foldlevel, which might not be desired result.

I'm not really a user of folds, but can you clarify when the issue happens more often, when doing todo change on the closed fold or on the opened fold?
You could hook up into `TodoChanged` event and do this for your use case, only issue is that the event is fired before the log notes and everything, but we might be able to move that around if it helps.

seflue:
Hey @kristijanhusak, I took another stab at this. You were right about `zx` being too aggressive.

Turns out the root cause is that treesitter's `on_changedtree` only reports the changed headline line to the fold system, but fold boundaries extend to the section end. So Neovim just never re-evaluates the fold end.

The fix now uses `vim._foldupdate(win, section_start, section_end)` to poke Neovim into updating just the affected section. I capture the section range before the edit (TS nodes go stale), then schedule the update via `vim.schedule` so the tree is reparsed first. No more `zx`, no more resetting other folds.

Also added foldtext cache invalidation via `changedtick` in a separate commit - that was causing stale ellipsis positions after edits.

Covers TODO, priority, archive and tag changes. Promote/demote doesn't need it since the structural change already gives treesitter a big enough range.

lua/orgmode/org/mappings.lua
```diff
@@ -50,6 +75,8 @@ function OrgMappings:set_tags(tags)
   local headline = self.files:get_closest_headline()
   local headline_tags = headline:get_own_tags()
   local current_tags = utils.tags_to_string(headline_tags)
+  -- Capture TS range before promise chain — node may be stale by the time input resolves
```
kristijanhusak: Setting tags should not shift the vertical position (start and end line) of the headline/section, as well as for the priority changes. Is there some edge case?

lua/orgmode/org/mappings.lua
```diff
@@ -17,6 +17,31 @@ local Promise = require('orgmode.utils.promise')
 local Input = require('orgmode.ui.input')
 local Footnote = require('orgmode.objects.footnote')
 
+---Capture section range for fold update. Must be called before buffer edits
+---because TS nodes become stale after changes.
+---@param headline OrgHeadline
+---@return fun() trigger Call after the buffer edit to schedule the fold update
+local function schedule_fold_update(headline)
```
kristijanhusak: I'd structure this function a bit differently:

1. Let it accept only `OrgRange`
2. Do not return function but instead do a `vim.schedule` directly
3. Update the implementation to consider that `OrgRange` is 1-indexed
4. When you need to call it on the stale (old) range, store the headline range in variable with `headline:get_range()` and then call the function when necessary.

kristijanhusak (changes requested):
Thanks for the changes, it looks much better now!

Regarding the stale nodes, I think in most of these cases we don't need to worry about the positions since changes do not modify start and end line of a headline/section.

Ideally, I'd have these updates be handled by the single event handler like we have for align tags. This would require adding few more events like `TagsChanged` and `PriorityChanged`, but it's useful to have those anyway.

Let me know what you think.

seflue:
Good suggestions, makes sense. Will push an update soon.

> Setting tags should not shift the vertical position (start and end line) of the headline/section, as well as for the priority changes. Is there some edge case?

The issue isn't about line count changes — it's about how Neovim's fold engine responds to treesitter changes.

`folds.scm` defines `(section)` as `@fold`, so folds span the entire section (headline + body). When tags or priority change, `on_changedtree` only reports the headline line. Neovim re-evaluates `foldexpr` for that line alone — never for the section end where the fold boundary lives. On a folded heading this breaks the fold.

Easy to reproduce: fold a heading with body content, set tags, fold is broken. Same with priority, and with TODO changes that don't produce a CLOSED line (e.g. TODO to PROGRESS).

lua/orgmode/org/mappings.lua
seflue: See [comment](https://github.com/nvim-orgmode/orgmode/pull/1078#issuecomment-4069300060) below.

lua/orgmode/org/mappings.lua
seflue: Done.

lua/orgmode/org/mappings.lua
```diff
@@ -17,6 +17,25 @@ local Promise = require('orgmode.utils.promise')
...
+---OrgRange is 1-indexed; vim._foldupdate expects 0-indexed lines.
+---@param range OrgRange
+local function schedule_fold_update(range)
+  local bufnr = vim.api.nvim_get_current_buf()
+  vim.schedule(function()
+    if not vim.api.nvim_buf_is_valid(bufnr) then
+      return
+    end
+    local start_line = range.start_line - 1
+    local end_line = math.max(range.end_line - 1, vim.api.nvim_buf_line_count(bufnr))
```
kristijanhusak: `nvim_buf_line_count` also returns 1-indexed value. We should do math.max on the values as they are and then subtract 1 because of `_foldupdate`
```suggestion
    local end_line = math.max(range.end_line, vim.api.nvim_buf_line_count(bufnr))
    end_line = end_line - 1
```

Edit: Also, shouldn't this be `math.min` ?

kristijanhusak (changes requested):
Thanks for the changes!

Just fix the end line thing and it's good to go.

lua/orgmode/org/mappings.lua
seflue: Changed to `math.min`, good catch.

The `- 1` is not needed because `erow` is 0-indexed exclusive (`treesitter/_fold.lua:212`), so 1-indexed inclusive values can be passed directly.

kristijanhusak (approved):
Looks good, thanks!

PR #1101, opened 2026-02-04
fix(edit_file): Hide tmp buffer window

celsobenedetti:
## Summary

This is a small tweak to the [tmp_edit_window](https://github.com/nvim-orgmode/orgmode/blob/tweak/edit_file-hide_tmp_edit_buffer/lua/orgmode/utils/init.lua#L523) in `utils.edit_file` to make it hidden and not focusable.
I don't think this is strictly needed, but I think it makes sense, and makes the agenda a bit cleaner.

I came across this due to a really obscure issue where the tmp edit window would not close after [Agenda._remote_edit](https://github.com/celsobenedetti/orgmode/blob/tweak/edit_file-hide_tmp_edit_buffer/lua/orgmode/agenda/init.lua#L563) actions.
Then, when I would try to `goto_item` (`<tab>`), the tmp window would be focused (hence, `focusable=false`).

I think the changes here make the UX a bit cleaner, since we don't even see the tmp edit window, and won't risk accidentally focusing it for some reason.

## Related Issues

N/A

## Changes

- Added `hide=true` and `focusable=false` to `vim.api.nvim_open_win` args
- Removed existing TODO comment

## Checklist

I confirm that I have:

- [x] **Followed the
      `make test`.

### DIFF
--- a/lua/orgmode/utils/init.lua
+++ b/lua/orgmode/utils/init.lua
@@ -524,12 +524,13 @@ function utils.edit_file(filename)
       vim.api.nvim_open_win(bufnr, true, {
         relative = 'editor',
         width = 1,
-        -- TODO: Revert to 1 once the https://github.com/neovim/neovim/issues/19464 is fixed
-        height = 2,
+        height = 1,
         row = 99999,
         col = 99999,
         zindex = 1,
         style = 'minimal',
+        focusable = false,
+        hide = true,
       })
       vim.api.nvim_set_option_value('swapfile', false, { buf = bufnr })
     end,


celsobenedetti:
I pinpointed the issue down to the `dashboard` module of  `snacks.nvim`.
I have not looked much deeper into this, but the issue may not be in `nvim-orgmode`, but some particular interaction with snacks.
May not be worth troubleshooting it.

I can reproduce it with this minimal config:

```diff
diff --git i/scripts/minimal_init.lua w/scripts/minimal_init.lua
index 8aaef87..0580f7f 100644
--- i/scripts/minimal_init.lua
+++ w/scripts/minimal_init.lua
@@ -29,6 +29,18 @@ require('lazy').setup({
       require('orgmode').setup()
     end,
   },
+  -- lazy.nvim
+  {
+    "folke/snacks.nvim",
+    ---@type snacks.Config
+    opts = {
+      dashboard = {
+        -- your dashboard configuration comes here
+        -- or leave it empty to use the default settings
+        -- refer to the configuration section below
+      }
+    }
+  }
 }, {
   root = lazy_root,
   lockfile = nvim_root .. '/lazy.json',
```

1. open nvim without a buffer (shows dashboard)
2. open agenda
3. do some `_remote_edit` action like `ona`.
4. tmp buffer won't close correctly
5. `tab` in an agenda heading will focus floating tmp buffer

Anyway, I think we can go ahead with the changes in this PR. Its an improvement overall, and fixes this particular bug described here.

kristijanhusak (approved):
Ok, thanks for looking into it!

PR #1113, opened 2026-03-10
feat(citations): add Org 9.5 citation support with BibTeX source, completions, and go-to-source

### LINKED ISSUE #718: Add support for citations

Yes

### Orgmode link

https://orgmode.org/manual/Citations.html

### Feature value

Introduced in orgmode 9.5 (so fairly recent), this is most commonly used in academia with a bibliography file.

### Additional context

There's a companion function, `org-cite-insert, where one can then insert and edit citations.

We discussed this in the org-roam plugin discussion. This isn't a requirement for the initial version of org-roam.nvim, but I wanted to document it here as a request because it is part of core orgmode.

NickHu:
## Summary

This PR adds support for [Citations](https://orgmode.org/manual/Citations.html), introduced by orgmode 9.5.

## Related Issues

Depends on https://github.com/nvim-orgmode/tree-sitter-org/pull/7 (this is why the tests fail)

Closes #718 

## Changes

- **`lua/orgmode/org/citations/init.lua`** — `OrgCitations` class: registers sources, aggregates `get_items()`, dispatches `follow(key)`, and resolves the citation key at cursor via tree-sitter `citation_reference` node
- **`lua/orgmode/org/citations/bibtex.lua`** — built-in `OrgCitationBibtex` source; parses `.bib` files (mtime-cached), resolves paths from `citations.org_cite_global_bibliography` config and file-local `#+bibliography:` directives, implements `follow(key)` to open the file at the entry line
- **`lua/orgmode/org/autocompletion/sources/citations.lua`** — `omnifunc` completion source; matches `[cite:@` / `[cite/style:@` prefix and returns keys from all registered sources
- **`lua/orgmode/config/defaults.lua`** — adds `citations.org_cite_global_bibliography` and `citations.sources` defaults
- **`lua/orgmode/colors/highlighter/markup/citation.lua`** + `queries/org/markup.scm` — syntax highlighting for citation nodes
- **`lua/orgmode/org/mappings.lua`** — `org_open_at_point` now delegates to `citations:follow()` when on a citation key
- **`docs/configuration.org`** — new `Citations` section covering syntax, bibliography config, custom source API, and Zotero Local API example

## Checklist

I confirm that I have:

- [X] **Followed the
      `make test`.

## Warning

The bulk of this PR is AI-generated, although I have combed over it extensively and cleaned it up by hand and fixed things.
It is basically exactly the changes you would expect to see from the PR title, except for the following design considerations:
1. citation sources are extensible; this is to facilitate populating them from outside a bibtex file (e.g. Zotero library which is given as an example in the documentation);
2. `lua/orgmode/org/citations/_meta.lua` adds two fields `label` and `description` to `OrgCitationItem`, which currently aren't used anywhere; the hope is that these items can be utilised in the completion popup, but it seems like support for ancillary information in completion items is not fleshed out in nvim-orgmode yet as a whole. The intended use-case is so that the completion call/popup can show or react to something like `Doe, John 2026 Some Article` (`label`) while still completing the key `@doe2026`, and LSP hover can display extra information (like an article abstract) from the `description` field.

I have tested it extensively, and I think it is ready for upstream review, but I am marking it as draft for the above reasons.

### DIFF
--- a/lua/orgmode/config/defaults.lua
+++ b/lua/orgmode/config/defaults.lua
@@ -96,6 +96,10 @@ local DefaultConfig = {
   hyperlinks = {
     sources = {},
   },
+  citations = {
+    sources = {},
+    org_cite_global_bibliography = {},
+  },
   mappings = {
     disable_all = false,
     org_return_uses_meta_return = false,
--- a/lua/orgmode/org/citations/bibtex.lua
+++ b/lua/orgmode/org/citations/bibtex.lua
@@ -1,0 +1,155 @@
+local config = require('orgmode.config')
+
+---@type table<string, { mtime_sec: number, items: OrgCitationItem[] }>
+local _cache = {}
+
+---@param content string
+---@return OrgCitationItem[]
+local function parse_bibtex(content)
+  local items = {}
+  for entry_type, key in content:gmatch('@(%a%w*)%s*[{(]%s*([^%s,}%)]+)') do
+    local lt = entry_type:lower()
+    if lt ~= 'string' and lt ~= 'preamble' and lt ~= 'comment' then
+      table.insert(items, { key = key })
+    end
+  end
+  return items
+end
+
+---@param path string
+---@return OrgCitationItem[]
+local function parse_file(path)
+  local stat = vim.uv.fs_stat(path)
+  if not stat then
+    return {}
+  end
+  local mtime_sec = stat.mtime.sec
+  local cached = _cache[path]
+  if cached and cached.mtime_sec == mtime_sec then
+    return cached.items
+  end
+  local lines = vim.fn.readfile(path)
+  local items = parse_bibtex(table.concat(lines, '\n'))
+  _cache[path] = { mtime_sec = mtime_sec, items = items }
+  return items
+end
+
+---@param raw string
+---@param base_dir? string
+---@return string
+local function resolve_path(raw, base_dir)
+  raw = vim.trim(raw)
+  if raw:sub(1, 1) == '~' then
+    return vim.fn.expand(raw)
+  end
+  if raw:sub(1, 1) ~= '/' then
+    local base = base_dir or vim.fn.getcwd()
+    return vim.fn.fnamemodify(base .. '/' .. raw, ':p')
+  end
+  return raw
+end
+
+---@class OrgCitationBibtex:OrgCitationSource
+---@field private files OrgFiles | nil
+local OrgCitationBibtex = {}
+OrgCitationBibtex.__index = OrgCitationBibtex
+
+---@param opts { files: OrgFiles | nil }
+function OrgCitationBibtex:new(opts)
+  return setmetatable({ files = opts and opts.files or nil }, OrgCitationBibtex)
+end
+
+---@return string
+function OrgCitationBibtex:get_name()
+  return 'bibtex'
+end
+
+---@return OrgCitationItem[]
+function OrgCitationBibtex:get_items()
+  local items = {}
+  for _, path in ipairs(self:_get_bib_paths()) do
+    vim.list_extend(items, parse_file(path))
+  end
+  return items
+end
+
+---Open the .bib file at the entry for the given key.
+---@param key string
+---@return boolean
+function OrgCitationBibtex:follow(key)
+  for _, path in ipairs(self:_get_bib_paths()) do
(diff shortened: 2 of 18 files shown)


NickHu:
@kristijanhusak rebased onto master and tests passing

lua/orgmode/org/mappings.lua
```diff
@@ -919,6 +921,13 @@ function OrgMappings:open_at_point()
   if footnote then
     return self:_jump_to_footnote(footnote)
   end
+
+  if self.citations then
```
kristijanhusak, line 925: Why do we need this check?

lua/orgmode/config/defaults.lua
```diff
@@ -96,6 +96,10 @@ local DefaultConfig = {
   hyperlinks = {
     sources = {},
   },
+  citations = {
+    sources = {},
+    org_cite_global_bibliography = {},
```
kristijanhusak, line 101: Let's move this to the top level config as other things to make it consistent. Sources configuration can stay like this.

lua/orgmode/org/citations/bibtex.lua
```diff
@@ -0,0 +1,155 @@
...
+  local lines = vim.fn.readfile(path)
+  local items = parse_bibtex(table.concat(lines, '\n'))
+  _cache[path] = { mtime_sec = mtime_sec, items = items }
+  return items
+end
+
+---@param raw string
+---@param base_dir? string
+---@return string
+local function resolve_path(raw, base_dir)
```
kristijanhusak, line 40: We can use [get_real_path](https://github.com/nvim-orgmode/orgmode/blob/master/lua/orgmode/utils/fs.lua#L55-L55) from `utils.fs.lua` instead of this function.

lua/orgmode/org/citations/bibtex.lua
```diff
@@ -0,0 +1,155 @@
+local config = require('orgmode.config')
```
kristijanhusak, line 1: Since we also support custom sources with this PR, we can create an `org/citations/sources` folder and put this there. I see that Emacs orgmode also supports json format so we can also add that as a built in.

lua/orgmode/org/citations/bibtex.lua
```diff
@@ -0,0 +1,155 @@
...
+    end
+  end
+
+  if self.files then
+    local current_filename = vim.fn.expand('%:p')
+    if current_filename ~= '' then
+      local file = self.files:load_file_sync(current_filename)
+      if file then
+        local file_dir = vim.fn.fnamemodify(file.filename, ':p:h')
+        local directives = file:_get_directive('bibliography', true)
```
kristijanhusak, line 123: Let's not use private methods outside of the file. Add a public method `file:get_bibliography` and do this inside.
```suggestion
        local directives = file:get_bibliography()
```

lua/orgmode/org/citations/bibtex.lua
```diff
@@ -0,0 +1,155 @@
...
+  local stat = vim.uv.fs_stat(path)
+  if not stat then
+    return {}
+  end
+  local mtime_sec = stat.mtime.sec
+  local cached = _cache[path]
+  if cached and cached.mtime_sec == mtime_sec then
+    return cached.items
+  end
+  local lines = vim.fn.readfile(path)
```
kristijanhusak, line 31: We should use async file reading whenever possible. Look into `utils.readfile`. This would require refactoring few things to be async, but shouldn't be big of a problem.

Also, what Copilot mentioned below, try using nsec when available. When it is not available, `nsec` is 0, so you need to fallback to sec. You can check how it's done in `files/file.lua`.

lua/orgmode/org/citations/bibtex.lua
```diff
@@ -0,0 +1,155 @@
...
+    if type(global) == 'string' then
+      add(global, nil)
+    else
+      for _, p in ipairs(global) do
+        add(p, nil)
+      end
+    end
+  end
+
+  if self.files then
```
kristijanhusak, line 117: We can default `self.files` to empty table in the `new()` so there's no need to do these kind of checks.

kristijanhusak (changes requested):
Sorry for the late review. Overall it seems to be working fine. I'd just like to refactor few things before we merge it in.

PR #1120, opened 2026-03-20
fix(indents): delegate indents for code blocks to treesitter instead of vim

dtvillafana:
## Summary

delegate indents for code blocks to treesitter instead of vim.
This fixes the issue where code blocks are incorrectly formatted for languages that vim does not have indents set up for.

The issue I was having was that the code blocks were not getting formatted according to the injections.scm that is already in the repo.

## Related Issues

Related #

Closes #

## Changes

delegate indents for code blocks to treesitter instead of vim

## Checklist

I confirm that I have:

- [x] **Followed the
      `make test`.

### DIFF
--- a/lua/orgmode/org/indent.lua
+++ b/lua/orgmode/org/indent.lua
@@ -279,21 +279,46 @@ local function indentexpr(linenr, bufnr)
         local block_ft = vim.treesitter.get_node_text(block_parameters[1], bufnr)
 
         if block_ft and block_ft ~= vim.bo.filetype then
-          local curr_indentexpr = vim.filetype.get_option(block_ft, 'indentexpr') --[[@as string]]
+          local block_header_indent = vim.fn.indent(match.line_nr)
+          local got_indent = false
 
-          if curr_indentexpr and curr_indentexpr ~= '' then
-            curr_indentexpr = curr_indentexpr:gsub('%(%)$', '')
-
-            local buf_shiftwidth = vim.bo.shiftwidth
-            vim.bo.shiftwidth = vim.filetype.get_option(block_ft, 'shiftwidth')
-            local ok, block_ft_indent = pcall(function()
-              return vim.fn[curr_indentexpr]()
-            end)
-            if ok then
-              new_indent = math.max(block_ft_indent, vim.fn.indent(match.line_nr))
+          -- Try treesitter indent via the injected parser first
+          local ts_ok, ts_indent_mod = pcall(require, 'nvim-treesitter.indent')
+          if ts_ok then
+            local parser_ok, parser = pcall(vim.treesitter.get_parser, bufnr)
+            if parser_ok and parser then
+              local lang = parser:language_for_range({ linenr - 1, 0, linenr - 1, 0 })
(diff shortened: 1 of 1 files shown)


kristijanhusak:
If you use nvim-treesitter indentexpr for a filetype that is being used in the code block, it should pull that up automatically.

dtvillafana:
Ah, so this is really an error on my end...
Even so, should it explicitly delegate it?
If it does, it would improve the "out of the box" experience for those who haven't fully configured their indent expressions.

kristijanhusak:
If your `indentexpr` is correctly set, then yeah. First line in the if statement picks up the indentexpr.
How are you setting your indentexpr?

dtvillafana:
in my config (generated by nixvim):
```lua
vim.bo.indentexpr = "v:lua.require'nvim-treesitter'.indentexpr()"
```
then later on:
```lua
vim.api.nvim_create_autocmd('FileType', {
   pattern = 'org',
   group = vim.api.nvim_create_augroup('org_indent_fix', { clear = true }),
   callback = function()
     local ok, orgmode = pcall(require, 'orgmode')
     if ok then
         vim.bo.indentexpr = "v:lua.require('orgmode.org.indent').indentexpr()"
     end
   end,
})
```

maybe whatever version of treesitter/org grammar I was using was bad...

relevant files in my config if you wanted to look: 
- https://github.com/dtvillafana/nixvim/blob/master/config/org.nix
- https://github.com/dtvillafana/nixvim/blob/master/config/treesitter.nix

kristijanhusak:
You shouldn't need to do this for org, it should automatically happen from [here](https://github.com/nvim-orgmode/orgmode/blob/master/indent/org.lua).

What I meant is how do you set indentexpr in your other filetypes, like the ones that you use in your code blocks. For example, if you use lua in your code blocks, how do you set `indentexpr` for lua?

dtvillafana:
Currently, it's all from treesitter (unless overriden by LSP for some filetypes).

in my init.lua it's set this way:
```lua
    -- Enable features via autocommands for modern nvim-treesitter
    vim.api.nvim_create_autocmd("FileType", {
        group = augroup,
        pattern = "*",
        callback = function()
            pcall(vim.treesitter.start)
            vim.bo.indentexpr = "v:lua.require'nvim-treesitter'.indentexpr()"
        end,
    })

```

which is then later overrode by this as shown previously:
```lua
vim.api.nvim_create_autocmd("FileType", {
    pattern = "org",
    group = vim.api.nvim_create_augroup("org_indent_fix", { clear = true }),
    callback = function()
        -- OrgIndent handles visual indentation; for = formatting, use orgmode's indent
        local ok, orgmode = pcall(require, "orgmode")
        if ok then
            vim.bo.indentexpr = "v:lua.require('orgmode.org.indent').indentexpr()"
        end
    end,
})
```

But for whatever reason, the indentation for code blocks in org files still does not pick up an injected indentexpr (without the patch I've made), I am not smart enough to figure out why.

I'm beginning to think this may be a lazy-loading issue, or an ordering issue, but again I'm not sure. 🤷‍♂️

PR #1131, opened 2026-04-20
feat(promises): Switch to coroutine based Promise

kristijanhusak:
## Summary

Update Promise class to use coroutines. This change will allow writing more readable async code by providing async wrapper where necessary (new `Promise.async()` method), removing the need for multiple `next()` chains.
Changes should be fully backward compatible, with removal of the few unused methods, like `Promise.any`, `Promise.race` and `Promise.all_settled`.

## Checklist

I confirm that I have:

- [X] **Followed the
      `make test`.

### DIFF
--- a/lua/orgmode/utils/promise.lua
+++ b/lua/orgmode/utils/promise.lua
@@ -1,471 +1,572 @@
----@diagnostic disable: undefined-field
--- Taken from https://github.com/notomo/promise.nvim
+---@alias OrgPromiseState 'pending' | 'fulfilled' | 'rejected'
 
-local vim = vim
+---@class OrgPromisePackedValues
+---@field n integer
+---@field [integer] any
 
-local PackedValue = {}
-PackedValue.__index = PackedValue
-
-function PackedValue.new(...)
-  local values = vim.F.pack_len(...)
-  local tbl = { _values = values }
-  return setmetatable(tbl, PackedValue)
-end
-
-function PackedValue.pcall(self, f)
-  local ok_and_value = function(ok, ...)
-    return ok, PackedValue.new(...)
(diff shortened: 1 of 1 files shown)


kristijanhusak:
@seflue @chipsenkbeil just a heads up about this change. It should not affect org-roam, but I'd appreaciate if you can confirm that nothing is broken.

chipsenkbeil:
@kristijanhusak I'll try to take a look this weekend. Finishing up some work on neovim core.

Btw, just got some early portion of `vim.ui.img` landed, and working on buffer-local images next. Once they land, you might be interested in an addition to nvim-orgmode to support [orgmode images](https://orgmode.org/manual/Images.html) in some fashion as part of the neovim 0.13 release whenever that happens.

kristijanhusak:
Sounds good. I'm following the progress on that :)

Currently we use Snacks for images, but I'm definitely looking forward to using the built in api once it's ready.

kristijanhusak:
Hey @chipsenkbeil , did you have a chance to look at this?

chipsenkbeil:
@kristijanhusak there's one bug that this PR causes for me, but I think it's based on something I'm doing.

It looks like `Promise:wait()` doesn't pump the event loop when the promise is already fulfilled, while the old implementation always did because :next was deferred through vim.schedule.

I've got some code relying on that side effect, which I'll need to fix. Go ahead and merge and I'll add a fix myself. (Went ahead and [did it here](https://github.com/chipsenkbeil/org-roam.nvim/commit/0240378c92eb0c9d6bc4ccbc8a6046e3c3d02ece) because I think this works with the old and new way that Promise:wait works)

PR #1132, opened 2026-04-21
feat(capture): Support for `%^g` and `%^G` expansions in capture templates.

alecStewart1:
## Summary

Add capture expansions for:

`%^g`:

 - Prompt for tags, with completion on tags in target file.
 
`%^G`:

 -  Prompt for tags, with completion all tags in all agenda files.

## Changes

Added functions: 

- `uniq_tags`
- `get_target_tags` (for `%^g`)
- `get_all_tags` (for `%^G`)

For getting tags for `%^g` / `%^G` expansions.

Added `prompt_tags` to prompt on `%^g` / `%^G` expansions.

The above are used in `expansions` in `lua/orgmode/capture/template/init.lua`.

Addressed LuaLS warning in `lua/orgmode/capture/templates.lua` regarding `templates or config.org_capture_templates`.

Added plenary tests, update test helper `create_agenda_files` to return files for template tests.

## Checklist

I confirm that I have:

- [X] **Followed the
      `make test`.


lua/orgmode/capture/template/init.lua
```diff
@@ -5,6 +5,89 @@ local Calendar = require('orgmode.objects.calendar')
...
+    end
+  end
+
+  return uniq_tags(tags)
+end
+
+---@description For `%^G` expansion in capture templates: gets all tags in all agenda files.
+---@return string[]
+local function get_all_tags()
+  local org = require('orgmode')
```
kristijanhusak: We should not require the global orgmode. We need to pass down `files` from `capture/init.lua` through the constructor, because capture can have different files from the `require('orgmode').files` (for example in org-roam.nvim).
In the `capture:open_template`, we can set the files on the template with `template.files = self.files`, and then we can provide the `files` to these template expansions.
It requires some refactoring, but it's necessary to make this work as expected.

kristijanhusak (changes requested):

lua/orgmode/capture/template/init.lua
alecStewart1: Refactored. Sorry about that!

alecStewart1: Oh should probably see if I need to update tests, too...oops.

lua/orgmode/capture/template/init.lua
```diff
@@ -5,6 +5,90 @@ local Calendar = require('orgmode.objects.calendar')
...
+---@param single boolean
+---@param tags_source string[]
+---@return OrgPromise<string>
+local function prompt_tags(single, tags_source)
+  local completion = function(arg_lead)
+    return vim.tbl_filter(function(tag)
+      return tag:match('^' .. vim.pesc(arg_lead))
+    end, tags_source)
+  end
+  local prompt = single and 'Tag: ' or 'Tags: '
```
kristijanhusak: From what I see in the [Orgmode docs](https://orgmode.org/manual/Template-expansion.html#:~:text=%E2%80%98%25%5Eg%E2%80%99,all%20agenda%20files.), there's not single/multiple type of prompt. It's either tags in single file, or tags in all agenda files.

lua/orgmode/capture/template/init.lua
```diff
@@ -5,6 +5,90 @@ local Calendar = require('orgmode.objects.calendar')
...
+  local ok, file = pcall(function()
+    return files:get(template:get_target())
+  end)
+
+  if not ok or not file then
+    return {}
+  end
+
+  local tags = {}
+  for _, tag in ipairs(file:get_filetags()) do
```
kristijanhusak: We already have this functionality in `get_tags` in `orgmode/files/init.lua`. Let's refactor it a bit to allow reusing it:
1. Create a `get_tags()` method in `orgmode/files/file.lua`
2. Copy over [this logic](https://github.com/nvim-orgmode/orgmode/blob/master/lua/orgmode/files/init.lua#L98-L113) to that method
3. Update the `get_tags()` in `files/init.lua` to call `orgfile:get_tags()`
4. Use the same method here with 'file:get_tags()`

lua/orgmode/capture/template/init.lua
```diff
@@ -5,6 +5,90 @@ local Calendar = require('orgmode.objects.calendar')
...
+---@param tags_source string[]
+---@return OrgPromise<string>
+local function prompt_tags(single, tags_source)
+  local completion = function(arg_lead)
+    return vim.tbl_filter(function(tag)
+      return tag:match('^' .. vim.pesc(arg_lead))
+    end, tags_source)
+  end
+  local prompt = single and 'Tag: ' or 'Tags: '
+  return Input.open(prompt, '', completion):next(function(input)
```
kristijanhusak: We can use the existing function for completion like it's done [here](https://github.com/nvim-orgmode/orgmode/blob/master/lua/orgmode/agenda/types/tags.lua#L111-L113)

lua/orgmode/capture/template/init.lua
```diff
@@ -357,14 +448,30 @@ end
 ---@return OrgPromise<string>
 function Template:_compile_prompts(content)
   local prepared_inputs = {}
-  for exp in content:gmatch('%%%^%b{}') do
+  -- Match %^{...} with optional g/G suffix for tag prompts
+  for exp in content:gmatch('%%%^%b{}[gG]?') do
```
kristijanhusak, line 418: Why do we need this change?

tests/plenary/helpers.lua
```diff
@@ -85,7 +85,7 @@ function M.create_agenda_files(fixtures, config)
   }, config or {})
   local org = orgmode.setup(cfg)
   org:init()
-  return files
+  return files, org.files
```
kristijanhusak, line 89: You can `require('orgmode')` in files when you need it.

kristijanhusak (changes requested):
Thanks for the changes! It is much better, but there are still few things to address before merging.

lua/orgmode/capture/template/init.lua
alecStewart1: I'll double check, but I was running into some issue with `g`/`G` still being with the resulting tag(s).

alecStewart1: To clarify: without this, tests were failing _and_ the outputted tags when I tested would include `g`/`G`.

alecStewart1:
The CI failures on macos/ubuntu are 4 refile tests in `api_spec.lua`. I don't _think_ I touched anything related to those.

alecStewart1:
Terribly sorry for not getting back on this sooner! Seems to work fine.

PR #1148, opened 2026-05-16
perf(highlighter): Cache decoration-provider hot paths

seflue:
Typing in folded org-files is slow because `nvim_set_decoration_provider`'s `on_line` does per-line work on every redraw.

Two commits add caching to foldtext: move `vim.fn.foldclosed` and `vim.fn.col` from `on_line` to `on_win`, and replace the `changedtick` wholesale-drop with a per-entry comparison of the cached line content, the same check markup already uses. Unedited closed folds keep their `hl_group` across edits and skip the treesitter lookup.

A third commit fixes stale virtual indent after archiving or refiling a subtree, which I hit while testing this. It is independent and can be split out if you prefer.

### DIFF
--- a/lua/orgmode/colors/highlighter/foldtext.lua
+++ b/lua/orgmode/colors/highlighter/foldtext.lua
@@ -11,8 +11,8 @@ local config = require('orgmode.config')
 ---@field highlighter OrgHighlighter
 ---@field namespace number
 ---@field cache table<number, table<number, OrgFoldtextLineValue>>
----@field cache_tick table<number, number>
 ---@field win_state table<number, OrgFoldtextWinState>
+---@field attached table<number, boolean>
 local OrgFoldtext = {}
 
 ---@param opts { highlighter: OrgHighlighter }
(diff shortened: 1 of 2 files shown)


kristijanhusak:
I just pushed a change to master to hide leading stars via tree-sitter queries (5ea62517521c), so we can rebase and simplify this a bit more.

seflue:
I removed the commit regarding the stars from my PR and force-pushed. You said, you have some simplifications in mind. Is it regarding the window map, holding a cache for each window? I was a bit undecided ... it's covering an edge case, when the user has the buffer open in two windows in parallel. You could argue, that this is in indeed a bit "edgy". On the other hand - especially when working with large files it is not completely uncommon to have the same buffer at different positions open in parallel.

lua/orgmode/colors/highlighter/foldtext.lua
```diff
@@ -1,102 +1,174 @@
...
+  if self.attached[bufnr] then
     return
   end
-
-  vim.api.nvim_buf_set_extmark(bufnr, self.highlighter.namespace, line, value.col, {
-    hl_mode = 'combine',
-    virt_text = { { config.org_ellipsis, value.hl_group } },
-    virt_text_pos = 'overlay',
-    ephemeral = self.highlighter._ephemeral,
+  local ok = vim.api.nvim_buf_attach(bufnr, false, {
```
kristijanhusak: The highlighters already run inside the decoration provider, so doing a buf attach inside a decoration provider handler is not really ideal.
If decoration provider is not working for foldtext, then we should remove it from decoration provider and attach that separately in the highlighter.

Alternatively, nvim 0.12+ now has `on_range` in the decoration provider, which can handle multiple lines. 
Could you try using that instead of `on_line` and see if that would fix the issue?
If not, I'd go with my above suggestion to just use nvim_buf_attach only.

kristijanhusak (changes requested):
Thanks for the PR! I left one comment, would like to see if any of the suggested options work before we proceed.

lua/orgmode/colors/highlighter/foldtext.lua
seflue: The attach was not a workaround for the decoration provider; drawing goes through `on_win`/`on_line` as before. It only fed cache invalidation: `on_bytes` dropped the entries in the edited rows and shifted the rest, so unedited closed folds kept their cached `hl_group` instead of the whole cache being dropped on every changedtick.

`on_range` does not help with that. It changes how the provider hands out redraw ranges but carries no information about which rows were edited, so it cannot replace `on_bytes`.

I dropped the attach and validate cache entries by line content instead, [the same way `OrgMarkup` does](https://github.com/nvim-orgmode/orgmode/blob/ae52811b8b3561b0ad867bdea69472585016686c/lua/orgmode/colors/highlighter/markup/init.lua#L52) with `line_content`. `on_win` already fetches the visible lines, so the check is a string compare per closed fold and there is no listener or row shifting anymore. Added a test where a same-length edit changes the capture at the line end, which a length-only check would miss.

kristijanhusak (approved):
Thanks!

PR #1164, opened 2026-06-25
fix(priority): don't insert an empty [# ] cookie when clearing priority

liamrlawrence:
## Summary
- `Headline:set_priority` only treated a blank priority as "remove the cookie"
  when a cookie already existed. The other branches ran when the headline had no
  cookie yet and always built a new one, even for a blank priority. As a result,
  pressing `<Space>` at the priority prompt on a headline with no existing cookie
  returned `" "` from `PriorityState:prompt_user`, fell through to an insertion
  branch, and produced a literal `[# ]` cookie instead of clearing the priority.

## Changes
- Move the empty-priority check to the top of `set_priority` so it applies to all
  paths: if the trimmed priority is empty, remove the existing cookie if present,
  otherwise do nothing. The insertion branches now only run with a real priority
  and can no longer fabricate an empty `[# ]` cookie.
- Add a test covering `<Space>` on a headline without an existing priority.

## Checklist
I confirm that I have:
- [x] **Followed the
      `make test`.

### DIFF
--- a/lua/orgmode/files/headline.lua
+++ b/lua/orgmode/files/headline.lua
@@ -347,9 +347,17 @@ end
 ---@param priority string
 function Headline:set_priority(priority)
   local _, priority_node = self:get_priority()
+  priority = vim.trim(priority)
+
+  if priority == '' then
+    if priority_node then
+      return self:_set_node_text(priority_node, '')
+    end
+    return
+  end
+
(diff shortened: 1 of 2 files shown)


kristijanhusak (approved):
Thanks!

PR #1167, opened 2026-06-29
feat(todos): route fast-access through ui.menu.handler

seflue:
The fast-access prompt writes its keys straight to the message area with `nvim_echo`, and there is no hook a UI plugin can use to take it over. This routes it through `ui.menu.handler` when one is configured, the same path the agenda menu already uses, so a plugin like [org-modern.nvim](https://github.com/danilshvalov/org-modern.nvim) can render it. Each keyword becomes a menu option keyed by its shortcut, plus a space option to clear it. With no handler set the prompt keeps the existing `nvim_echo` output, so the default look does not change.

The menu data keeps the structure so a handler can use it: keywords are grouped by sequence (`groups`, a list of option lists), each option carries its highlight group (`hl`), and the menu has a stable `kind` (`todo_fast_access`) to branch on. `items` stays as the flat view for handlers that ignore grouping, and the built-in echo menu is unchanged.

A matching org-modern change renders these groups as a colored matrix ([commit on my fork](https://github.com/seflue/org-modern.nvim/commit/6a857479f3038db06c9fc2d1ecdd570fae773ee7)).

A handler must return the result of the selected action, since fast-access uses it to get the chosen keyword back. Menus whose actions only have side effects never needed this, so a handler that does not return will silently no-op here until it adds the `return`.

### DIFF
--- a/lua/orgmode/objects/todo_state.lua
+++ b/lua/orgmode/objects/todo_state.lua
@@ -25,6 +25,58 @@ end
 
 ---@return OrgTodoKeyword | nil
 function TodoState:open_fast_access()
+  if config.ui.menu.handler then
+    return self:_open_fast_access_menu()
+  end
+  return self:_open_fast_access_echo()
+end
+
+--- Delegates to the configurable menu handler so UI plugins can render the
+--- prompt in a window instead of the message area.
+---@private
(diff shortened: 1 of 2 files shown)


kristijanhusak (approved):
Thanks!

PR #1179, opened 2026-09-15
fix(memoize): release cache with file, reset on re-parse

seflue:
In long-running Neovim sessions with org files open, memory use keeps growing as I edit them and never comes back down. I have seen this for a while; one instance was at over 10 GB.

The memory is tree-sitter trees, pinned through the memoize cache, which never lets go of them. It is meant to shrink on its own. It is a module-level table declared `__mode = 'k'`, which reads as "entries go away when their key is collected". They do not, because the keys are filenames and string keys are never weak in Lua ([`iscleared()` in Lua 5.1's `lgc.c`](https://www.lua.org/source/5.1/lgc.c.html), [5.4 manual §2.5.4](https://www.lua.org/manual/5.4/manual.html#2.5.4)). The obvious repair, keying by the file object, does not work either. A cached value references the file it was built from, and a weak-key table still marks its values ([`gc_traverse_tab`](https://github.com/LuaJIT/LuaJIT/blob/v2.1/src/lj_gc.c) skips marking only when keys *and* values are weak), so every entry keeps its own key alive. Lua 5.2 added ephemeron tables for exactly this case; LuaJIT has none. This is why the leak was easy to miss. The code looks like it handles the lifetime, and the reasons it does not are two GC details that hold only for strings and only for LuaJIT.

On top of that, an entry grows while its file is edited. Results are bucketed by node id, lookups after a re-parse use the ids of the new tree, and the entry is only reset when the mtime changes, so a file collects one set of dead buckets, and with it one tree, per re-parse between saves.

Two commits:

- The cache moves from the module-level `Memoize.cache` to a field on the `OrgFile` instance, `file.memoize_cache`, so it lives exactly as long as the instance, with no invalidation call to remember.
- The entry carries `__generation`, the root node id, next to `__version`. When either changes the entry is replaced as a whole, so a re-parse drops the old buckets and with them the tree they pinned.

Measured with a headless harness: load a set, render the week agenda, then rounds of edit a buffer, reload, redraw. Lua heap from `collectgarbage('count')`, C heap from `mallinfo2()`, RSS from `/proc`.

About 1500 files, 4.5 MB of org text, agenda open: 52 MB Lua heap, 209 MB C heap, 360 MB RSS on both versions. After unloading all files:

| | Lua heap | C heap | RSS |
|---|---|---|---|
| `master` | 49 MB | 196 MB | 358 MB |
| this branch | 17 MB | 9 MB | 164 MB |

Ten rounds of edit, reload, redraw on 350 files with 100 of them open in buffers, live Lua + C heap:

| | this branch | `master` |
|---|---|---|
| after the first render | 32 MB | 32 MB |
| after 10 rounds | 35 MB | 52 MB |

`master` gains about 2 MB per round, one more tree of each edited file.


kristijanhusak (approved):
Great find! Thanks!

PR #1180, opened 2026-09-16
perf(files): cache path discovery, filter before resolve

seflue:
## Summary

`OrgFiles:_files()` runs on every call that walks the file set, and each run globs, resolves and stats every match. With a recursive glob like `~/org/**/*` that costs 70-100 ms per call on my ~1000 file tree. org-roam.nvim produces exactly that glob for every configured directory and hands it to its own `OrgFiles` instance, which is where I measured it. Opening one org file triggers three of them.

Two commits.

1. Filter before resolve. The glob matches every file in the tree, org or not, and discovery resolved all of them before dropping the non-org ones. On my tree that is well over a thousand wasted `vim.fn.resolve` calls per run. Cold discovery roughly halves, from about 140 ms to about 75 ms.

2. Keep the result until the next `load()` or `unload()`. `load()` always rediscovers, so a forced reload still sees new files. `load_file` with `persist` also rediscovers, because it asks whether a file that did not exist when the cache was built belongs to the configured paths. That runs once per created file. A warm `all()` goes from about 100 ms to nothing measurable.

## Changes

- `_files(refresh?)` replaces `_files(skip_resolve?)`. The unresolved variant only existed to avoid a second resolve during load, and `load_file` resolves the name anyway, so both callers now share one cached result. The call in `load()` still reads `_files(true)`, but `true` now means refresh.
- `_path_cache` on `OrgFiles`, cleared in `unload()`.

The trade-off is what happens when a file disappears from disk outside the editor. Before, the next `all()` dropped it because the stat failed during rediscovery. Now it stays in the list until the next load, so the agenda keeps showing its entries until you press `r`. I think that is acceptable: the parsed content in `self.files` was stale either way, and refreshing the agenda is what you do after changing files outside the editor anyway. But it is a behaviour change.

## Checklist

I confirm that I have:

- [x] **Followed the
      `make test`.


kristijanhusak (approved):
Thanks!

PR #1181, opened 2026-09-17
perf(buffers): cache filename to buffer lookups

seflue:
Walking every headline of a large agenda set through the API, which is what telescope-orgmode does on every open, takes several seconds. Most of that time is in `Buffers.get_buffer_by_filename`, which a cold traversal of a few hundred files calls tens of thousands of times.

`_bufs` only caches hits. For a file with no buffer, the normal case for an agenda set, every call repeats `vim.fs.normalize(vim.fn.resolve(filename))` and a `vim.fn.bufnr()` scan. At this call count that is most of the wall time. Since #1076 the memoize key getter on `OrgFile` calls `self:bufnr()` too, so every memoized method call now pays for a miss as well.

Caching misses needs invalidation, and there is none today. `add` and `remove` are not called from outside `state/buffers.lua`, `_bufs` is seeded in `init()` and afterwards only grows on hits, so an entry survives a rename or a wipeout and keeps resolving to a buffer that no longer holds the file.

One commit, all in `state/buffers.lua`:

- `_resolved` memoizes `_resolve_filename`.
- `_misses` caches negative lookups, checked after `_bufs`.
- `_setup_autocmds`, called from `init()`. `BufAdd`, `BufNew`, `BufFilePost` and `FileType` re-register the buffer and drop the miss cache; `BufDelete` and `BufWipeout` drop the buffer.

`FileType` is in the event list because a file without an org extension only counts as an org buffer once its filetype is set, and `add()` stores nothing for it before that, so nothing else would ever clear its miss.

Three tests in `tests/plenary/state/buffers_spec.lua`.


kristijanhusak (approved):
Thanks!

Issue #1048, opened 2025-10-28
63ace3ca makes Neovim startup slow and brittle


seflue:
### Describe the bug

It is a bit hard to describe, but with the change done in 63ace3ca I got two problems:

First without mitigation starting more then one Neovim instances get's an adventure with surprising effects. Because I am an org-roam user I have hundreds of little org-files. For all of these orgfiles now I get a swap file created on startup and when I start a second instance, it complaints (in Org-Roam) about those very swap files. But Roam is only the messenger here, the issue is Orgmode and one of it's API functions which got changed by this commit. This effect I found a one-line solution within orgmode, which basically disables swap file creation when loading org files over the API.
The second one is that startup get's really really slow (multiple seconds) and I got strange glitches and multiple notifications (Hardwarp enabled). For this I didn't find a solution.

I pinned down the causes of these effects to this very commit and reverted my local instance to the commit before to be able to use Neovim and Orgmode productively.

Because I am very busy right now I wasn't able to provide a deeper and more detailed analysis of the problem. But this commit breaks my daily workflow, so for me it is quite severe.

### checkhealth

orgmode:                                                                    ✅

Orgmode ~
- ✅ OK Treesitter grammar installed (version 2.0.0)
- ✅ OK Setup called
- ✅ OK `org_agenda_files` configured
- ✅ OK `org_default_notes_file` configured

### Steps to reproduce

Open neovim.
Open neovim somewhere else.

### Expected behavior

Same startup behavior as before commit 63ace3ca.

### Minimal init.lua

```lua

-- Enter your minimal_init.lua here

```

### nvim-orgmode version

All commits since 63ace3ca

### OS / Distro

Arch Linux

### Neovim version/commit

v0.11.4

PR #1182, opened 2026-09-17
perf(files): background load, parse-free clock scan

seflue:
ref #1048, replaces #1073

Opening a single org file waits until every agenda file has been read. The startup clock scan walks the headlines of every file, and that parses each one with tree-sitter. With a few hundred org-roam files that is seconds of blocked editor.

The parse is the expensive part, not the read; deferring the reads alone does not help. Two independent commits:

**The clock scan parses only candidate files.** A running clock is a CLOCK line with a single timestamp and nothing after it. A raw line scan picks the files that have one; only those get parsed, and the parser still decides. Archive files are skipped.

**Loading moves to the background.** `Org:init` no longer waits for `files:load()`; `Clock:init` chains on the load promise, and a second `load()` call during loading gets that promise instead of blocking. Callers that need all files still wait at `ensure_loaded`. The cost moves to the first agenda open.


kristijanhusak:
`load_promise` looks good, but I'm not sure about the clock line parsing. Do you know how time is saved by parsing the line with regex? With the `load_promise` change it should be more performant anyway.

seflue:
`load()` only reads the files. The parse happens in `get_clocked_headline`. With `load_promise` alone the scan still parses all of them once the reads finish; the freeze moves from `:edit` to a moment later. The pre-filter is what removes it.

Measured on 500 generated org-roam style files (1.8 MB, one running clock in the last file), median of 3 runs:

| variant | open (ms) | done (ms) | scan (ms) | rescan (ms) | files parsed |
|---|---:|---:|---:|---:|---:|
| master cd431cb1 | 583 | 583 | 470 | 110 | 500 |
| load_promise dc52ca2d | 44 | 595 | 470 | 123 | 500 |
| load_promise + pre-filter 697e8a58 | 46 | 143 | 11 | 2 | 2 |

- open: how long `:edit` of the first org file blocks
- done: time from `:edit` until the startup clock scan has finished
- scan: duration of that scan, the editor is blocked meanwhile
- rescan: a second scan, every clock in/out/cancel runs one
- parsed: files the startup scan parsed with tree-sitter

Script: [orgbench.lua](https://gist.github.com/seflue/17692b704e78edaef2078bcfc1f1e1d2). `nvim -l orgbench.lua` writes the fixture, checks out the three commits as worktrees next to the script and prints the table, `--files 2000` scales it. To feel it: `nvim -u orgbench/init-master.lua orgbench/fixture/note-0001.org`, same with `init-prefilter.lua`. Vibe-coded with Claude's help.

kristijanhusak:
Ok, thanks for the benchmark. We can go with it for now, but I'm hoping to find a better solution in the future for tracking clocked in headline.
Thanks for the PR!

kristijanhusak (approved):

PR #1186, opened 2026-09-19
fix(links): Skip links without a url

seflue:
Typing `[[]]` into an org file, for example while starting a link, breaks every caller of `OrgFile:get_links()` until the brackets are filled:

```
vim/treesitter.lua:179: attempt to index local 'node' (a nil value)
  in function 'get_node_text'
  lua/orgmode/org/links/hyperlink.lua:53: in function 'from_node'
  lua/orgmode/files/file.lua: in function 'get_links'
```

The grammar still produces a `link` node for the empty brackets, but without a `url` field. Same for `[[][desc]]`.

`from_node` now returns nil for a link node without a url. `get_links` skips those nodes; the `at_cursor` callers already handle nil.


kristijanhusak:
There should be no empty `(link)` node in the TS tree (link without description) when it's empty. It should be just `(expr)` in that case.
For the one with description, I'd rather do it on the query level like this:

```diff
diff --git a/lua/orgmode/files/file.lua b/lua/orgmode/files/file.lua
index 632b5b2..17d9d95 100644
--- a/lua/orgmode/files/file.lua
+++ b/lua/orgmode/files/file.lua
@@ -857,7 +857,7 @@ function OrgFile:get_links()
   local links = {}
   local matches = self:get_ts_captures([[
     (link) @link
-    (link_desc) @link
+    (link_desc url: (expr)+) @link
   ]])
 
   local source = self:get_source()

```

seflue:
Checked the tree with 2.0.4: `[[]]` already comes out as `(expr)` nodes, the scanner rejects it. The description was wrong about that. What crashed `get_links` is `[[][desc]]`, a `link_desc` without `url`.

Switched to your query filter and dropped the nil check in `from_node`. Description updated.

`OrgHyperlink.at_cursor()` still throws on `[[][desc]]` under the cursor, so `org_open_at_point` on such a link shows the `get_node_text` error instead of doing nothing. That's a follow-up PR.

kristijanhusak (approved):

